PR #39, opened 2017-05-19
fmtread: match uppercase INFINITY

phomes:
The check for y was done lower case twice instead of both
lower and upper. Change this to both y and Y to match "INFINITY"


phomes:
What is needed to land this?

mleair:
The fix has been added (see previous message).

PR #40, opened 2017-05-19
Fix -Wlogical-not-parentheses compile errors

pointhi:
Those two lines of code are ambiguous, and can result in compile warnings.


sscalpone:
This issue was addressed in commit b5061109a5e841ce785d4c54993b64fe0e24feb0

PR #57, opened 2017-05-23
Use `uname -m` instead of `-i` in CMakeLists.txt

tdaff:
On My Gentoo machine `-i` reports `GenuineIntel` so the build cannot determine the architecture. Documentation also says that it is non-portable, so should use the `-m` for the machine instead.

```
  -m, --machine            print the machine hardware name
  -i, --hardware-platform  print the hardware platform (non-portable)
```


sscalpone:
Fixed in a different commit.

PR #69, opened 2017-05-29
Add presentation PDF and disambiguation note (fixes #38)

### LINKED ISSUE #38: Relationship with llvm-flang
Hi, sorry if this is not the right place to ask, but I can't seem to figure out what's the relationship of this project with https://github.com/llvm-flang/flang. Are they simply two distinct projects with the same name, or do they have common ancestry?

waldyrious:


waldyrious:
I don't have the sources for that document, unfortunately. I was pointed to it by @ChrisTX in #38 -- maybe he can help?

kiranchandramohan:
@waldyrious Do you have an update for this PR? If not can you consider closing it?

waldyrious:
Hey @kiranchandramohan — I'd be happy to make the change suggested by @pawosm-arm if someone could point me to the sources of the PDF file. Otherwise, if you are OK with adding the PDF as-is, I can certainly rebase the PR to fix the merge conflicts, so that the PR is ready to merge.

pawosm-arm:
@waldyrious I looked though this PDF and even some if its sources would be binary JPEG image files. Hence the last option is to use `pstopdf` or `ps2pdf` or `a2ps -1` to generate textual .ps file. My test showed, `pstopdf` creates 15MB .ps textual file, while `ps2pdf` and `a2ps -1` both create 64MB .ps textual file. Certainly, we'd like the smaller file (I'm attaching it here: [ARM-HPC-UG-SC16-PGI.ps.txt]() filename with added `.txt` suffix (remove it before using this file) as github does not allow attachments of unknown file types, even if they are plaintext only). I've tried various viewers to verify the quality and as usual, GhostView was terrible, GIMP import from postscript was almost acceptable, the best was Okular with its 'import PostScript' option.

waldyrious:
Hey @pawosm-arm — thanks for the additional comments! Are you sure that `.ps` file is preferable, though? IMO it is not that much of an improvement over a binary file, since the bulk of the (fairly large) plaintext source consists of opaque image blobs, such as the one starting at line 20344:

```ps
︙
<<
  /ImageType 1
  /Width 3000
  /Height 1687
  /ImageMatrix [3000 0 0 -1687 0 1687]
  /BitsPerComponent 8
  /Decode [0 1 0 1 0 1]
  /DataSource currentfile
    /ASCII85Decode filter
    /LZWDecode filter
>>
image
J1MOKW"]jL"b]2#:hXAd)[hEh@M(SI<[AN=7H774jC17kO2kcD3>@c/T32kbKi'f2
>q_R=AlC^_80eI:.#@23ktZLfT$QV")C5HE6_i%C'Y7:&P>A&T.Ff?5:9@Eq<GOI<
eTSK1UFb"C"CaWB*"IS_QFg\K&e,!XN?r]$.8ldN.k_^>c0kOY^gAcsr#Q(LLVmS?
\@b9e'og:EnlY7U7k9eHAoej&&j^c5-^fk-+<V$GVmZk"@H8W.ntg#s2B+6.cY^)F
︙
B,6VF(=@%:&T\acG8HBg(X[^L&VCotLDZ/3(t"B^&X+)0QPkpT):>&p&Yg7AV](\u
)UY`-&[NER[i:IA)puD?&]5Sc`uL5b*7<(Q&^qatf,^".*RWac&`Xp0k8ocO*msEu
&b@)ApE,Op+4:*2&d'7S#9]0;+O.\E+9~>
```

...so in a way, one can say it's still mostly binary content, just encoded in a different (and substantially more verbose) format.

The PDF file, besides being smaller in size (1.36 MB), can be rendered inline by GitHub, as shown [here](https://github.com/waldyrious/flang-1/blob/disambig/docs/ARM-HPC-UG-SC16-PGI.pdf).

kiranchandramohan:
Can you upload into google doc/drive with a publicly available link and use that here?

Any other suggestions?

SouraVX:
I may be mistaken :) but the title of the PR does not align with the changes.

waldyrious:
> Can you upload into google doc/drive with a publicly available link and use that here?

I could, but then that file would be attached to my own Google account, which doesn't seem like a reliable solution for a public project... for instance, if a few years from now I forget about it and delete the file, or for some other reason it stops being accessible, there's no easy way to recover it (it is already inaccessible in the link from #38, by the way).

> Any other suggestions?

I honestly don't think a 1.3M file is that much of an issue for a git repo. I agree that a plaintext source is preferable to binary blobs, but that's not available in this case, and the alternatives explored so far seem to have more cons than pros (IMHO). So I'd ask you to reconsider.

kiranchandramohan:
> for instance, if a few years from now I forget about it and delete the file, or for some other reason it stops being accessible, there's no easy way to recover it

This project will likely not be active after a few years.

Overall I think discussing this more is not productive. And i agree that 1.3M is not going to cause much of a problem. So please rebase and make it fit for submission. If one more person in the review list agrees, I will merge.

waldyrious:
@kiranchandramohan done. 👍

kiranchandramohan (approved):
LGTM.

bryanpkc (approved):
I have no objection to this change. I will merge it.

SouraVX (commented):
Thanks! for doing this :)
Since we have converged on including bin files, I was wondering if we could attach @pawosm-arm  presentation also :). It's already open-source :)
Additionally we can make an entry in Wiki something like  `introduction to Flang`
What do you guys think ?

kiranchandramohan:
In the case of @pawosm-arm's presentation, I would prefer a link since it is already publicly available on the llvm-hpc workshop website. 

I have created a page in the wiki which points to this presentation and other docs. https://github.com/flang-compiler/flang/wiki/Introduction-to-Flang

PR #76, opened 2017-05-31
Fixed building bug on Arch & Gentoo Linux

Ben0mega:
I was unable to build on a Gentoo computer with an error about the "GenuineIntel" architecture.  Changing the uname flag to -m fixed the build here (and on Arch Linux).  This change is not necessary for Ubuntu, but would be for Mac OS X.  uname -m is both standardized (IEEE) and well supported.


Ben0mega:
#57 Covers this patch.

junghans:
flang is on its way into Gentoo: gentoo/gentoo#4867.

PR #83, opened 2017-06-05
cmake: fix a typo in target architecture detection

junghans:
"uname -i" returns 'GenuineIntel' or 'Marvell Kirkwood' on most
Linux systems, however, what flang is looking for seems to be
"uname -m", which return 'x86_64' or 'armv5tel'.

Also "uname -i" is non-portable, while "uname -m" is. See "man
uname" for details.

Signed-off-by: Christoph Junghans <junghans@lanl.gov>


junghans:
Duplicate of #57.

PR #135, opened 2017-07-13
towards architectural neutrality

pawosm-arm:
Let's try something lightweight for the first pull request.
Proposed change ensures equal treatment of 64-bit architectures by not limiting availability of 64-bit specific parts of the code to x86_64 targets.


pawosm-arm:
Seems like after latest changes this pull request has less and less of the content.

bryanpkc:
It seems that `TARGET_64` is not used anywhere, but there is a precedent for `TARGET_64BIT`. Can we use `TARGET_64BIT` and `TARGET_32BIT` instead?

pawosm-arm:
> It seems that `TARGET_64` is not used anywhere, but there is a precedent for `TARGET_64BIT`. Can we use `TARGET_64BIT` and `TARGET_32BIT` instead?

I'm seeing both used. `TARGET_64` is used in CMake files (next to other `...64` suffixed stuff so I guess there is some consistency here), while `TARGET_64BIT` is only defined in `pgifeat.h` and only checked for in `iliutil.cpp` file.

pawosm-arm:
rebased, refreshed.

tools/flang2/utils/ilitp/CMakeLists.txt
```diff
@@ -13,8 +13,7 @@ target_compile_definitions(ilitp
   PRIVATE
   ILITP_UTIL
   PGF90=1
-  TARGET_X86
-  TARGET_X8664
+  TARGET_64
```
bryanpkc, line 16: Can we just delete `TARGET_X86` and `TARGET_X8664`? They aren't used under this directory at all.

bryanpkc:
> I'm seeing both used. 

grep shows:
```
$ grep -r TARGET_64
runtime/libpgmath/lib/CMakeLists.txt:    TARGET_LINUX_GENERIC TARGET_64 PG_PIC)
runtime/libpgmath/lib/CMakeLists.txt:    TARGET_LINUX_GENERIC TARGET_64 PG_PIC)
tools/shared/pgifeat.h:#define TARGET_64BIT 1
tools/flang2/flang2exe/iliutil.cpp:#if DEBUG && defined(TARGET_64BIT)
```
Even though `TARGET_64` is defined, it isn't used anywhere. I think we should unify the macros into just one named `TARGET_64BIT`.

tools/flang2/utils/ilitp/CMakeLists.txt
pawosm-arm: I've noticed that downstream we removed all of those TARGET_64 and TARGET_32 and no harm was done, so I'm reworking this old patch again.

pawosm-arm:
@bryanpkc is it more acceptable now?

runtime/include/komp.h
```diff
@@ -174,7 +174,7 @@ typedef void ident_t;
 
 typedef int kmp_int32;
 typedef long long kmp_int64;
-#if defined(TARGET_X8632) || defined(TARGET_LINUX_ARM32)
+#if __WORDSIZE == 32
```
bryanpkc, line 177: `__WORDSIZE` is not defined by default in MSVC (https://godbolt.org/z/M8rfsPhYn). A brief Google search shows that it is recommended to do this in portable software:
```
#ifdef _MSC_VER
#ifdef _WIN64
#define __WORDSIZE 64
#else
#define __WORDSIZE 32
#endif
#endif
```

bryanpkc: Comment @xoviat ?

pawosm-arm: @bryanpkc I've added the snippet that you'd suggested. Is it portable enough now?

runtime/include/komp.h
```diff
@@ -172,9 +172,21 @@ class __cplus_omp_init_lock_t
...
 
+#ifndef __WORDSIZE
+#ifdef _MSC_VER
+#ifdef _WIN64
+#define __WORDSIZE 64
+#else
+#define __WORDSIZE 32
+#endif
+#else
+#error "Unknown word size!"
```
mmuetzel, line 183: Does this still compile correctly for MinGW?
Afaict, `__WORDSIZE` isn't defined for that platform either. IIUC, it is only defined on platforms with glibc.

pawosm-arm: Sadly, I don't have any MinGW machine available at present. __WORDSIZE isn't only defined on platforms with glibc (e.g. XCode defines it in stdint.h, which is covered by the next pull request). If you can verify that the following code fails to compile on MinGW, please let us know:
```
#include <stdint.h>

#ifndef __WORDSIZE
#error "__WORDSIZE not defined"
#endif
```

mmuetzel: With MINGW64 (from MSYS2):
```
$ cat wordsize.c
#include <stdint.h>

#ifndef __WORDSIZE
#error "__WORDSIZE not defined"
#endif

Markus@Markus-PC MINGW64 ~
$ gcc wordsize.c
wordsize.c:4:2: error: #error "__WORDSIZE not defined"
    4 | #error "__WORDSIZE not defined"
      |  ^~~~~
```

mmuetzel: See also #1271.

PR #150, opened 2017-07-19
make use of FLANG_VENDOR

pawosm-arm:
FLANG_VENDOR is defined in CMakeLists.txt and when set, adds -DFLANG_VENDOR=\"${FLANG_VENDOR} \" to compiler's command line (note space always added after vendor string!). As FLANG_VENDOR is nowhere used in the code, this commit adds the use in relevant places.


pawosm-arm:
Unlikely to be accepted. Closing it.

tools/flang1/flang1exe/main.c
```diff
@@ -127,7 +127,11 @@ static char *accel = NULL;
 
 /** Product name in debug output
  */
+#ifdef FLANG_VENDOR
+#define DNAME FLANG_VENDOR "F90"
```
bryanpkc, line 131: Instead of hardcoding "F90" here, can we use `version.lang` instead?

tools/shared/ccffinfo.c
```diff
@@ -54,7 +54,11 @@ static int prevchildnest = -1;
 static int prevlineno = 0;
 static bool anymessages;
 
+#ifdef FLANG_VENDOR
+#define BUILD_VENDOR FLANG_VENDOR
```
bryanpkc, line 58: IMHO this belongs in the `version` structure as well.

tools/flang2/flang2exe/main.cpp
```diff
@@ -56,6 +56,14 @@
 
 static bool process_input(char *argv0, bool *need_cuda_constructor);
 
+/** Product name in debug output
+ */
+#ifdef FLANG_VENDOR
+#define DNAME FLANG_VENDOR "F90"
```
bryanpkc, line 62: Same here; can we use `version.lang` instead?

tools/flang1/flang1exe/listing.c
```diff
@@ -116,8 +116,13 @@ list_ln(char *beg, char *txt)
     if (!lf)
       return; /* in case error msg written before file
                * opened */
+#ifdef FLANG_VENDOR
+    fprintf(lf, "\n\n\n%s%s (Version %8s)          %s      page %d\n\n",
+            FLANG_VENDOR, version.lang, version.vsn, gbl.datetime, pgno);
```
bryanpkc, line 121: Actually, instead of using the macro everywhere, perhaps the `version` struct should have a `vendor` field. Or we could repurpose the `product` field. What do you think?

tools/flang1/flang1exe/main.c
pawosm-arm: `"F90"` is used here only because `DNAME` is otherwise defined as `"F90"` (see few lines below)

tools/flang1/flang1exe/listing.c
pawosm-arm: The main reason this macro was introduced is due to our LLVM guys demand to have flang's counterpart to their `CLANG_VENDOR`. I have no opinion on that, I've only materialized their will.

pawosm-arm (commented):
x

tools/flang1/flang1exe/main.c
bryanpkc: Yeah, I understand that. What I meant was that we should just replace `DNAME` with `version.lang`.

tools/shared/ccffinfo.c
pawosm-arm: I agree, but this isn't a refactor patch.

tools/flang1/flang1exe/main.c
pawosm-arm: The way DNAME is used rules out the use of version.lang here. What I could do was to make sure that both version.lang and DNAME are referring the same name.

pawosm-arm:
@reviewers: gentle reminder

PR #174, opened 2017-08-03
flang1, flang2: larger buffers, to hold longer filenames

pawosm-arm:
It's not uncommon for Linux systems to have complex directory trees with long paths (e.g. when things are kept on NFS). In such a case, 258-chars MAX_FNAME_LEN is too small: it is used when combining paths as given by -I or -J flags with imported module names. We could easily cause compiler to malfunction with buffer that small.


pawosm-arm:
Solves problems present only in our development environment. Unlikely to be accepted.

PR #193, opened 2017-08-16
Add LLVM_LIBDIR_SUFFIX for /usr/lib64

xiangzhai:
Hi flang developers,

Just like [llvm](https://github.com/llvm-mirror/llvm/blob/master/CMakeLists.txt#L281), there is `LLVM_LIBDIR_SUFFIX` for `/usr/lib64` library path, so I add it to CMakeLists.txt, please review it, thanks a lot!

Regards,
Leslie Zhai


kiranchandramohan:
Hello @xiangzhai,
Is this patch still valid? If so, would you be able to rebase and push. The CI will run and then we can see whether this patch can be merged.

PR #270, opened 2017-10-13
Pull 2017-10-13T07-26 Recent NVIDIA Changes

tskeith:
Replace duplicated files with shared ones in tools/shared and runtime/shared.


tools/shared/direct.c
```diff
@@ -366,6 +410,14 @@ direct_rou_load(int func)
   }
 #endif
 
+#if (defined(TARGET_X86) || defined(TARGET_LLVM)) && !defined(FE90)
```
sscalpone: TARGET_X86 || TARGET_LLVM is always true.

PR #318, opened 2017-11-28
runtime: fix parallel build of omp_lib

pawosm-arm:
This is to prevent following error when running 'make' with -j argument:

```
F90-F-0004-Corrupt or Old Module file ../../include/omp_lib_kinds.mod (runtime/flang/omp_lib.F95: 593)
F90/aarch64 Linux Flang - 1.5 2017-05-01: compilation aborted

```

### DIFF
--- a/runtime/flang/CMakeLists.txt
+++ b/runtime/flang/CMakeLists.txt
@@ -146,6 +146,7 @@ SET(FTN_INTRINSICS_DESC_INDEP
   merge.c
   mvbits3f.c
   nargs3f.c
+  omp_lib_kinds.F95
   omp_lib.F95
   outstr3f.c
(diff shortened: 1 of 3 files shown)


xoviat:
@pawosm-arm is on a roll!

xoviat:
Note: I don't think that this is the reason for this error. I think the reason is that the module directory is the same for both static and shared libraries, resulting in flang overwriting module files built with the other library.

sscalpone:
I agree with 'ghost' -- need to decide either that we'll install the generated mod files after they are built or else we'll redirect the mod files for one of static/shared to the bitbucket.

pawosm-arm:
Needs different approach.

PR #350, opened 2018-01-02
Pull 2018-01-02T13-01 Recent NVIDIA Changes

tskeith:


tools/flang2/flang2exe/cgmain.c
```diff
@@ -2398,6 +2398,8 @@ write_instructions(LL_Module *module)
       if ((!flg.ieee) || XBIT(216, 1))
         switch (i_name) {
         case I_FADD:
+          if (XBIT(216, 2))
```
sscalpone: Added to prevent accuracy problem when llvm vectorizes spec's fontonik3d; still investigating root cause.

PR #369, opened 2018-01-25
Pull 2018-01-24T16-58 Recent NVIDIA Changes

tskeith:


nncarlson:
Thanks for addressing the `bind(C, name='')` issue so quickly!

sscalpone:
Thanks Neil for helping us get it right.

PR #393, opened 2018-02-08
Pull 2018-02-08T12-20 Recent NVIDIA Changes

gklimowicz:
Apply nodepchk to a loop/routine as a pragma nodepchk for simd
clause or simd construct. Since the host function has
information about nodepcheck, need to propagate the info to
outlined function when it was created because we lose sptr of
host function when we expand outlined function.

### DIFF
--- a/tools/flang2/flang2exe/main.c
+++ b/tools/flang2/flang2exe/main.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997-2017, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 1997-2018, NVIDIA CORPORATION.  All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


tools/flang2/flang2exe/main.c
```diff
@@ -319,7 +319,13 @@ process_input(char *argv0, LOGICAL *need_cuda_constructor)
     dmp_dtype();
   if (gbl.rutype != RU_BDATA) {
     cuda_emu_end();
-    direct_rou_end();
+      /* TDB: make it look better!*/
```
sscalpone: TBD?

tools/shared/pragma.c
```diff
@@ -1488,6 +1488,37 @@ rouprg_enter(void)
   load_dirset(&direct.rou_begin);
 }
 
+/*
```
sscalpone: doxygen style

PR #398, opened 2018-02-12
Pull 2018-02-12T06-45 Recent NVIDIA Changes

gklimowicz:


sscalpone:
5aeaa73 mentions FS#s.

sscalpone:
5fb27e2 I think the unhiding has to do with pgi driver.  Also, not related to this commit but generally: Do we know if clang driver flang is hooked up for knl?

sscalpone:
3e52291 does flang have llvect?

PR #436, opened 2018-04-04
Pull 2018-04-04T05-53 Recent NVIDIA Changes

gklimowicz:


runtime/libpgmath/lib/common/dispatch.c
```diff
@@ -1269,13 +1269,7 @@ __math_dispatch_init()
       fputs("waiting for __math_dispatch\n", stderr);
     }
     while (false == __math_dispatch_is_init) {
-#if     defined(TARGET_X8664)
       __asm__("pause");
-#elif   defined(TARGET_LINUX_POWER)
```
sscalpone: Why aren't we preserving TARGET_LINUX_POWER?

runtime/libpgmath/lib/common/sincosf.c
```diff
@@ -15,9 +15,7 @@
  *
  */
 
-#ifdef  TARGET_X8664
```
sscalpone: Do we need to preserve this #ifdef for other targets?

PR #442, opened 2018-04-09
AArch64 specific tweaks

pawosm-arm:


pawosm-arm:
Separate updated pull requests for those commits will be issued soon.

Issue #468, opened 2018-05-13
[OpenMP]: ICE: wrong interplay between modules, threadprivate and copyin


pawosm-arm:
Following program causes flang1 to segfault when compiled with -fopenmp (can be compiled by gfortran!):
```
module somemod
  integer, allocatable :: sth(:)
  !$omp threadprivate(sth)
end module somemod

program some
  use somemod
  !$omp parallel copyin(sth)
  !$omp end parallel
end program some
```
```
error: unable to execute command: Segmentation fault (core dumped)
```
When modified as follows, it compiles properly (note that it needs 'save' to make sense):
```
module somemod
end module somemod

program some
  use somemod
  integer, allocatable, save :: sth(:)
  !$omp threadprivate(sth)

  !$omp parallel copyin(sth)
  !$omp end parallel
end program some
```
The first question here is, does flang module file format carry thread private attribute properly?

pawosm-arm:
lol, adding SAVE to the module variable also makes it buildable!

pawosm-arm:
There's another workaround: add SAVE statement to the module. Which leads me to this: https://software.intel.com/en-us/forums/intel-visual-fortran-compiler-for-windows/topic/275783
> On the practical side, our implementation always saves module variables
I guess flang should do the same.

gklimowicz:
Do you have a slightly bigger example that shows the use of ```sth```?

We'd like to make sure that we have a fix that works in a larger context.

Did this come from a code we should know about?

pawosm-arm:
The code was developed by Andy Davis during STFC hackaton:
```
module external_mod

integer,allocatable :: bob(:)
!$omp threadprivate(bob)

contains

subroutine init_bob(n)
Allocate(bob (n))
bob(1:n) = 42
end subroutine init_bob

end module external_mod
 
program thread_test
use external_mod
implicit none
 
integer :: loop,i
loop = 10
 
call init_bob(5)
 
!$omp parallel copyin(bob)
!$omp do schedule(runtime)
do i = 1, loop
enddo
!$omp end do
!$omp end parallel
 
stop
 
end program thread_test
```
My first soultion was to make all modules SAVE. Unfortunately, this broke RAPS16 compilation. So I came up with something less intrusive in semand.c: all THREADPRIVADE variables (THREADP(sptr, 1) are also made SAVEP(sptr, 1) iff they are part of a module (if (sem.mod_sym) is true).

pawosm-arm:
[...]

PR #469, opened 2018-05-15
Always SAVE module variables

pawosm-arm:
Something that other compilers always do, e.g. https://software.intel.com/en-us/forums/intel-visual-fortran-compiler-for-windows/topic/275783

> On the practical side, our implementation always saves module variables

This is to address bug #468


pawosm-arm:
This is wrong and causes RAPS16 build to fail. I have better fix, will be published soon.

PR #470, opened 2018-05-15
Pull 2018-05-15T04-05 Recent NVIDIA Changes

mleair:


tools/flang2/flang2exe/mwd.c
```diff
@@ -3208,6 +3208,8 @@ smsz(int m)
   return msz;
 } /* smsz */
 
+char* scond(int);
```
sscalpone: Are we also actively removing these kinds of mid-file declarations?

PR #487, opened 2018-06-01
Introduce weak symbols in the libompstub library

pawosm-arm:
Flang compiler, when started without -fopenmp links the final
executable binary with libompstub - a small library which
contains all the symbols of OpenMP runtime library with
empty implementations. This is because certain places
in Flang runtime library make use of OpenMP runtime in order
to use its mechanisms for handling concurret code execution.
For example, all malloc() calls in Flang runtime library
are wrapped by __kmpc_critical()/__kmpc_end_critical() calls.

The problems start to appear when Fortran program built
without -fopenmp (and therefore linked against libompstub)
is also linked against some library that was dynamically
linked against real OpenMP runtime library, resulting in
unexpected behaviour.

### DIFF
--- a/runtime/ompstub/ompstubs.c
+++ b/runtime/ompstub/ompstubs.c
@@ -26,210 +26,202 @@
 #include <stdio.h>
 #include "komp.h"
 
-int
-omp_get_thread_limit(void)
+#ifndef TARGET_WIN
+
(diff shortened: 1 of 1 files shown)


pawosm-arm:
Obsolete idea, closing it.

PR #502, opened 2018-06-14
Pull 2018-06-14T05-53 Recent NVIDIA Changes

mleair:


tools/flang1/flang1exe/symacc.c
```diff
@@ -60,11 +60,25 @@ sym_init_first(void)
     assert(stb.n_base, "sym_init: no room for namtab", stb.n_size, ERR_Fatal);
     stb.n_base[0] = 0;
     STG_ALLOC(stb.dt, 400);
+    STG_ALLOC_SIDECAR(stb.dt, stb.dt_shapemap);
```
sscalpone: Ideally this all would be inside of #ifdef OPENACC.

PR #503, opened 2018-06-14
Cavium port of flang and libpgmath to AArch64.

flang-cavium:
Rebased port of libpgmath to include all changes as of 06/11/2018.

AArch64 support in flang.

Explicit Dependencies:

https://github.com/flang-compiler/flang/pull/488
https://github.com/flang-compiler/flang/pull/489

When the two pull requests above are integrated, I will remove the corresponding changes from this pull request.


CMakeLists.txt
```diff
@@ -289,7 +289,7 @@ endif()
 # Add appropriate flags for GCC
 if (LLVM_COMPILER_IS_GCC_COMPATIBLE)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fno-common -Woverloaded-virtual -Wcast-qual -fno-strict-aliasing -pedantic -Wno-long-long -Wall -W -Wno-unused-parameter -Wwrite-strings")
-  option(WITH_WERROR "Compile with '-Werror' C compiler flag" ON)
+  option(WITH_WERROR "Compile with '-Werror' C compiler flag" OFF)
```
schweitzpgi: Do these changes cause the compiler to emit warnings?

runtime/flang/CMakeLists.txt
```diff
@@ -19,6 +19,12 @@ enable_language(C ASM Fortran) # Enable assembly and Fortran
 SET(ASM_OPTIONS "-DLINUX_ELF")
 SET(CMAKE_ASM_FLAGS "${CFLAGS} ${ASM_OPTIONS}" )
 SET(CMAKE_SHARED_LINKER_FLAGS "-no-flang-libs")
+if (${TARGET_ARCHITECTURE} STREQUAL "aarch64")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-z -Wl,combreloc")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-z -Wl,stack-size=536870912")
```
schweitzpgi: This looks like a phone number. Perhaps a comment would be good here.

CMakeLists.txt
flang-cavium: In file included from /builds-1/steleman/llvm-git/20180614-flang-github-cavium-60/llvm-git/tools/flang/runtime/libpgmath/lib/aarch64/acos.c:18:0:
/builds-1/steleman/llvm-git/20180614-flang-github-cavium-60/llvm-git/tools/flang/runtime/include/mthdecls.h:552:24: warning: ISO C does not allow extra ‘;’ outside of a function [-Wpedantic]
 FLTDECL_C(__mth_i_cabs);
                        ^
In file included from /builds-1/steleman/llvm-git/20180614-flang-github-cavium-60/llvm-git/tools/flang/runtime/libpgmath/lib/aarch64/acosh.c:21:0:
/builds-1/steleman/llvm-git/20180614-flang-github-cavium-60/llvm-git/tools/flang/runtime/include/mthdecls.h:552:24: warning: ISO C does not allow extra ‘;’ outside of a function [-Wpedantic]
 FLTDECL_C(__mth_i_cabs);
                        ^
/builds-1/steleman/llvm-git/20180614-flang-github-cavium-60/llvm-git/tools/flang/runtime/include/mthdecls.h:553:27: warning: ISO C does not allow extra ‘;’ outside of a function [-Wpedantic]
 CMPLXDECL_C(__mth_i_cacos);
                           ^
/builds-1/steleman/llvm-git/20180614-flang-github-cavium-60/llvm-git/tools/flang/runtime/include/mthdecls.h:554:27: warning: ISO C does not allow extra ‘;’ outside of a function [-Wpedantic]
 CMPLXDECL_C(__mth_i_casin);

[ ... etc etc etc ... ]

runtime/flang/CMakeLists.txt
flang-cavium: 536870912 bytes == 512 MB.

For OpenMP programs, including SPEC benchmarks, VERY LARGE stack sizes are required. For example, OMP_NUM_THREADS == 64 with OMP_STACKSIZE == 128M.

-Wl,-z -Wl,stack-size=<N> does not set the stack size. It only creates a permissible upper bound, which may or may not be in effect at run-time.

sscalpone:
@flang-cavium please resolve the conflicts & I'll get to the reviewing and merging.  Thanks!

flang-cavium:
Conflicts done.

include/int128.h
```diff
@@ -29,7 +29,8 @@ extern "C" {
 #include <stdbool.h>
 #include <stdint.h>
 
-#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ > 8)
+#if ((__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ > 8)) && \
+  !defined(__aarch64__)
```
sscalpone: Is __int128 not supported or why is this disabled for aarch64?

flang-cavium: Performs better when aligned on 4 instead of 8. Also, spurious SEGV's observed when using the native __int128 on AArch64 with GCC.

runtime/libpgmath/lib/common/arm64intrin.h
```diff
@@ -0,0 +1,2761 @@
...
+  t[10] = x[10];
+  t[11] = x[11];
+  t[12] = x[12];
+  t[13] = x[13];
+  t[14] = x[14];
+  t[15] = x[15];
+}
+
+/*
+ * No corresponding Altivec intrinsic to generate a scalar mask
```
pawosm-arm: Altivec references left throughout this header.

flang-cavium: Yes, because this is an implementation of the Altivec interfaces.

sscalpone:
There are ~65 files added as part of aarch64 that are duplicates of files in the generic directory.  I looked for dups in all of the files in libpgmath and found others too -- there are 4 in common and 9 files that appear in aarch64, generic and x86_64.   I am not convinced that we should deal with dup CMakeList.txt file.

The first bit of feedback is that we should eliminate the duplication of files.  @flang-cavium, we don't have a recommendation for how to eliminate the dups -- let me know if you'd like one.  As for the files that you didn't add -- either PGI or you can clean those up.  If PGI does it, we'll do it after this pull request is finished.

flang-cavium:
Is the current implementation of libpgmath stable? Meaning: is there going to be more re-arranging of file locations and directory structures?

I'm not convinced that dealing with duplicates now is a priority, as long as the internal structure of libpgmath is not final and stable.

sscalpone:
I see a lot of files in common with "avx" in their names; I've asked the developers to clarify for me how common ought to be used.

flang-cavium:
AVX vectorization is specific to Intel. These Intel-specific vectorization optimizations cannot be used on AArch64, or on any other ISA that I know of, except Intel.

sscalpone:
Yup, that's what prompted the question.

runtime/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ endif()
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
 
 add_subdirectory(ompstub)
+add_subdirectory(libpgmath)
```
sscalpone: I don't know how to accept this change & build that with different compilers for flang & libpgmath.

flang-cavium: If building libpgmath with a different compiler than the rest of flang is a requirement, a reasonable way of doing this would be:

1. Allow (but not require) the definition of libpgmath-specific compilers in CMAKE_OPTIONS:

-DLIBPGMATH_C_COMPILER:FILEPATH=<insert-c-compiler-path-here>
-DLIBPGMATH_CXX_COMPILER:FILEPATH=<insert-c++-compiler-path-here>
-DLIBPGMATH_CFLAGS:STRING=<insert-libpgmath-c-flags-here>
-DLIBPGMATH_CXXFLAGS:STRING=<insert-libpgmath-c++-flags-here>

2. Add the appropriate logic in libpgmath's toplevel CMakeLists.txt file:

if ("${LIBPGMATH_C_COMPILER}" STREQUAL "")
else()
  set(${CMAKE_C_COMPILER} "${LIBPGMATH_C_COMPILER}")
endif()

if ("${LIBPGMATH_CXX_COMPILER}" STREQUAL "")
else()
  set(${CMAKE_CXX_COMPILER} "${LIBPGMATH_CXX_COMPILER}")
endif()

if ("${LIBPGMATH_CFLAGS}" STREQUAL "")
else()
  set(${CMAKE_C_FLAGS} "${LIBPGMATH_CFLAGS}")
endif()

if ("${LIBPGMATH_CXXFLAGS}" STREQUAL "")
else()
  set(${CMAKE_CXX_FLAGS} "${LIBPGMATH_CXXFLAGS}")
endif()

[ ... etc etc etc ... ]

3. I am skeptical that LLVM will be open to the concept of building flang with one compiler and building libpgmath with another. That's not something they have ever allowed for LLVM or clang.

flang-cavium: Also:

> %> pwd
> ${top_srcdir}/tools/flang/runtime

> %> diff -u CMakeLists.txt.orig CMakeLists.txt
> --- CMakeLists.txt.orig 2018-06-20 14:47:29.239660258 +0000
> +++ CMakeLists.txt      2018-06-21 18:24:09.446198990 +0000
> @@ -49,7 +49,6 @@
 > include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
 >
 > add_subdirectory(ompstub)
> -add_subdirectory(libpgmath)
 >
 > add_subdirectory(flangrti)
 > add_subdirectory(flang)

*Without* the explicit

> add_subdirectory(libpgmath)

line in ${top_srcdir}/tools/flang/runtime/CMakeLists.txt, the following happens:

> cd /builds-1/steleman/llvm-git/20180620-flang-github-cavium-60/build/tools/flang/runtime/flang && /usr/local/gcc73/bin/gcc  -DLINUX -DMAXCPUS=256 -DMAXCPUSL=8 -DMAXCPUSR=8 -DNATIVE_FPCVT -DPGF90 -DPGFLANG -DPGI_LITTLE_ENDIAN -DTARGET_LINUX -DTARGET_LINUX_ARM -DTARGET_LLVM -DTARGET_LLVM_ARM64 -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -Dflang_shared_EXPORTS -DDESC_I8 -DHAVE_LONG_LONG_INT -DINT32PTR64 -DTM_I8 -I/builds-1/steleman/llvm-git/20180620-flang-github-cavium-60/build/tools/flang/runtime/flang -I/builds-1/steleman/llvm-git/20180620-flang-github-cavium-60/llvm-git/tools/flang/runtime/flang -I/builds-1/steleman/llvm-git/20180620-flang-github-cavium-60/llvm-git/tools/flang/include -I/builds-1/steleman/llvm-git/20180620-flang-github-cavium-60/build/tools/flang/include -I/usr/include/libxml2 -I/builds-1/steleman/llvm-git/20180620-flang-github-cavium-60/build/include -I/builds-1/steleman/llvm-git/20180620-flang-github-cavium-60/llvm-git/include -I/builds-1/steleman/llvm-git/20180620-flang-github-cavium-60/llvm-git/tools/flang/runtime/include  -g -O2 -pthread -fno-strict-aliasing -fno-tree-slp-vectorize -fno-tree-vectorize -fno-reorder-blocks -fno-toplevel-reorder -Wall -Wcast-align -Wno-long-long -Woverflow -Wstack-protector -ffunction-sections -fdata-sections -fkeep-static-consts -fstack-protector-all -fno-omit-frame-pointer -Wl,-rpath -Wl,/builds-1/steleman/llvm-git/20180620-flang-github-cavium-60/build/lib -Wl,-rpath -Wl,/usr/local/gcc73/lib64  -fPIC -Werror=date-time -Wall -W -Wno-unused-parameter -Wwrite-strings -Wno-missing-field-initializers -pedantic -Wno-long-long -Wno-comment -ffunction-sections -fdata-sections -O2 -DNDEBUG -fPIC   -fPIC -mcpu=thunderx2t99 -o CMakeFiles/flang_shared.dir/I8_sources/type.c.o   -c /builds-1/steleman/llvm-git/20180620-flang-github-cavium-60/build/tools/flang/runtime/flang/I8_sources/type.c
> gmake[2]: *** No rule to make target 'lib/libpgmath.so', needed by 'lib/libflang.so'.  > Stop.
> gmake[2]: *** Waiting for unfinished jobs....

runtime/libpgmath/lib/common/pow/fma3/sdpow.cpp
```diff
@@ -18,6 +18,8 @@
 
 #if defined(TARGET_LINUX_POWER)
```
sscalpone: runtime/libpgmath/lib/common/pow/fma3/sdpow.cpp:19:0: error: unterminated #if
 #if defined(TARGET_LINUX_POWER)

Did this file (and it looks like a few more) build for you?

flang-cavium: Apparently not, but I didn't get any errors either.

I will fix this right now.

jjones-cavium:
Are there any comments about this from the code implementation standpoint, as opposed to infrastructure? If there is something fundamentally objectionable to what's been done, I think it would be better to know sooner, rather than later.

As much as possible, a pull request should be coherent. To that end, I suggest the following:

1. Delegate building with separate compilers to another pull request, unless there is an explicit stated reason why this is needed for AArch64 support. Maybe I missed it, but I don't think that there has been.

2. Review the substance of this pull request as much as possible without rebasing. Once it is approved, freeze-out other checkins (or at least any that may conflict), rebase it, and separate out the directory auto-detection from the libpgmath support. Submit these as two separate pull requests, with the libpgmath support pull request relying on the directory auto-detection. 

Comments on my suggestions.

1. Putting support for multiple compilers in a single build in a separate pull-request allows it to be removed more cleanly if needed. Although I don't recall any discussions on llvm-dev in the last eight years about supporting separate compilers, but I think there might be some objection, better not to muddy the waters.

2. I think auto-detection is the way to go. This mimics what <llvm-root>/tools/CMakeLists.txt does. The same argument for separation of concerns is less strong here, as one of the two mechanisms is needed to be sure libpgmath actually builds, right?

pawosm-arm:
Hi Joel,
I don't have comments to the implementation core as it's either straight forward or imported from known sources. Regarding your plan, it sounds reasonable. Regarding your comment 1, I think in current flang project we're too detached from LLVM to suggest ourselves with llvm-dev discussions, this may come with f18 project. Comment 2 I'll leave to PGI :)

CMakeLists.txt
sscalpone: Would you remove the warning from the new and changed files please?  We'll get after the existing files with warning.

runtime/flang/CMakeLists.txt
sscalpone: Thanks for the explanation.  What's the downside of not including this option in the library build?  It seems odd to hard-code in SPEC benchmark requirements.  In any case, if the line must stay please document it in the CMakeLists.txt file.

runtime/CMakeLists.txt
sscalpone: We'd like to keep the libpgmath build separate from the rest of the flang build.  The intention is to promote libpgmath to the top-level of flang-compiler.  The readme and other docs describe the separate build-step for libpgmath.

sscalpone:
Hi Joel,

Agree, let's leave building with a different compilers to a different pull request.  For now, we'd like to keep the pgmath library build as a separate step in the documented build process as we plan to promote libpgmath to a top-level flang-compiler project.

runtime/CMakeLists.txt
flang-cavium: OK, but I don't see how you can do that and build flang at the same time.

As shown in my copy-and-paste example above, without the explicit

> add_subdirectory(libpgmath)

line, flang won't build.

sscalpone:
Each reviewer has asked that the dup files be removed (see list https://gist.github.com/sscalpone/69c2b5babd8cc3a78628) and the aarch CMakeLists.txt files reference the generic files.

CMakeLists.txt
flang-cavium: There are **many** more other warnings throughout flang code, unrelated to libpgmath:

> /builds-1/steleman/llvm-git/20180625-flang-github-cavium-60/llvm-git/tools/flang/tools/flang2/flang2exe/aarch64-Linux/ll_abi.c: In function ‘update_homogeneous’:
/builds-1/steleman/llvm-git/20180625-flang-github-cavium-60/llvm-git/tools/flang/tools/flang2/flang2exe/aarch64-Linux/ll_abi.c:64:3: warning: enumeration value ‘DT_NONE’ not handled in switch [-Wswitch]
>   switch (dtype) {
>   ^~~~~~

> /builds-1/steleman/llvm-git/20180625-flang-github-cavium-60/llvm-git/tools/flang/tools/flang2/flang2exe/aarch64-Linux/ll_abi.c:64:3: warning: enumeration value ‘DT_WORD’ not handled in switch [-Wswitch]
> /builds-1/steleman/llvm-git/20180625-flang-github-cavium-60/llvm-git/tools/flang/tools/flang2/flang2exe/aarch64-Linux/ll_abi.c:64:3: warning: enumeration value ‘DT_DWORD’ not handled in switch [-Wswitch]
> /builds-1/steleman/llvm-git/20180625-flang-github-cavium-60/llvm-git/tools/flang/tools/flang2/flang2exe/aarch64-Linux/ll_abi.c:64:3: warning: enumeration value ‘DT_HOLL’ not handled in switch [-Wswitch]
> /builds-1/steleman/llvm-git/20180625-flang-github-cavium-60/llvm-git/tools/flang/tools/flang2/flang2exe/aarch64-Linux/ll_abi.c:64:3: warning: enumeration value ‘DT_BINT’ not handled in switch [-Wswitch]
> /builds-1/steleman/llvm-git/20180625-flang-github-cavium-60/llvm-git/tools/flang/tools/flang2/flang2exe/aarch64-Linux/ll_abi.c:64:3: warning: enumeration value ‘DT_SINT’ not handled in switch [-Wswitch]

[ ... etc etc etc ... ]

> /tools/flang2/flang2exe/aarch64-Linux/ll_abi.c:73:10: warning: implicit declaration of function ‘size_of’ [-Wimplicit-function-declaration]
>    size = size_of(dtype);
>           ^~~~~~~

[ ... etc etc etc ... ]

> /builds-1/steleman/llvm-git/20180625-flang-github-cavium-60/llvm-git/tools/flang/tools/flang2/flang2exe/aarch64-Linux/ll_abi.c:231:7: warning: implicit declaration of function ‘alignment’; did you mean ‘indent’? [-Wimplicit-function-declaration]
>    if (alignment(dtype) > 4 && size % 8 == 0) {
>        ^~~~~~~~~
>        indent

[ ... etc etc etc ... ]
> /builds-1/steleman/llvm-git/20180625-flang-github-cavium-60/llvm-git/tools/flang/tools/flang2/flang2exe/bihutil.c:46:22: warning: passing argument 4 of ‘stg_alloc’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
>    STG_ALLOC(bihb, 128);
>                       ^
[ ... etc etc etc ... ]

Yes, I agree, flang should compile with zero warnings. That would make -Werror possible.

There are **numerous** const-correctness warnings throughout flang.

This is GCC 7.3. Note: I am **not** compiling with -ansi. That would make GCC very angry on all the K&R C function definitions.

runtime/libpgmath/lib/common/dispatch.c
```diff
@@ -1264,7 +1281,8 @@ __math_dispatch_init()
   if (__sync_bool_compare_and_swap(&__math_dispatch_in_prog, false, true)) {
     if (__mth_i_debug == 0x100) {
       fputs("calling __math_dispatch()\n", stderr);
-      sleep(1);
+      struct timespec tsp = { 0, 250000000 };
```
sscalpone: What is the reason to use nanosleep over sleep?

runtime/flang/CMakeLists.txt
flang-cavium: The downside is that any OpenMP program that has large stack size memory requirements might fail, independently of SPEC.

Very large stack sizes for OpenMP programs is common-place, and is unrelated to SPEC.

runtime/libpgmath/lib/common/dispatch.c
```diff
@@ -1264,7 +1281,8 @@ __math_dispatch_init()
   if (__sync_bool_compare_and_swap(&__math_dispatch_in_prog, false, true)) {
```
sscalpone: Just a quick comment not about this commit but in general -- shouldn't this be a loop?

runtime/libpgmath/lib/common/dispatch.c
flang-cavium: sleep is an approximation. it may or may not sleep for the number of seconds in the argument.

nanosleep is precise.

runtime/libpgmath/lib/common/dispatch.c
flang-cavium: In theory, the builtin __sync_bool_compare_and_swap should loop, if and only if it's implemented as LL/SC and not as a lock-free atomic.

 In AArch64 for T99, we support LSE - which means it's a lock-free atomic. I'm pretty sure it's the same on Intel X86_64 and PPCle. 

If the builtin is written as LL/SC and doens't loop, that's a builtin bug.

CMakeLists.txt
flang-cavium: I have placed the full output from building flang - with warnings - on T99 AArch64 here:

https://drive.google.com/open?id=19XCxUYDO39iih3zKQPEO5CqPHC2-oHoA

Warning: file is 28MB in size.

sscalpone: I don't get those warnings w/ gcc 7.3.0.  Seems like -pedantic isn't passed through for me.

flang-cavium: What does *Seems like -pedantic isn't passed through for me* mean?

Is -pedantic passed on compile-line?

Many of the warnings triggered by -pedantic are **serious** and they deserve attention.

sscalpone: *Seems like* means "it appears to be true but not known for sure" -- in this case, I don't see pedantic on the verbose command line but I did not investigate why pedantic does not appear to appear for me.

runtime/libpgmath/lib/common/mth_intrinsics.h
```diff
@@ -14,9 +14,18 @@
  * limitations under the License.
  *
  */
+#ifdef __cplusplus
```
pawosm-arm: Sadly, to compile this code with clang++, stdint.h needs to be included instead. I'd suggest something like this:
```
#if defined(__cplusplus) && !defined(__clang__)
#include <cstdint>
#else
#include <stdint.h>
#endif
```

runtime/libpgmath/lib/common/arm64intrin.h
```diff
@@ -0,0 +1,2826 @@
...
+  return r;
+}
+
+static inline __m128d
+__attribute__((always_inline))
+_mm_min_sd(const __m128d& a, const __m128d& b)
+{
+  double aa = vec_extract(a, 0);
+  double bb = vec_extract(b, 0);
+  aa < bb ? aa : bb;
```
sscalpone: Is the assignment missing?
https://github.com/flang-compiler/flang/issues/557

runtime/libpgmath/lib/common/arm64intrin.h
```diff
@@ -0,0 +1,2826 @@
...
+
+    if (ld != 0.0)
+      r.b += ld;
+
+    return r;
+  }
+
+  inline __s128i& operator+=(unsigned int i) {
+    if (i != 0U)
+      xird = xird + i;
```
sscalpone: https://github.com/flang-compiler/flang/issues/557
Looks like the clang is complaining about the conversion from unsigned int to int.
From GCC docs:
For convenience, it is allowed to use a binary vector operation where one operand is a scalar. In that case the compiler transforms the scalar operand into a vector where each element is the scalar from the operation. The transformation happens only if the scalar could be safely converted to the vector-element type.

gklimowicz:
@flang-cavium I believe we have merged the necessary changes for aarch64 support into libpgmath via pull request #538. I am closing this as a matter of bookkeeping. If we are missing something, please do another pull request for any remaining items.

PR #506, opened 2018-06-15
Pull 2018-06-15T06-29 Recent NVIDIA Changes

mleair:


tools/flang2/utils/ilmtp/aarch64/ilmtp.n
```diff
@@ -3609,3 +3609,11 @@ Call function returning float128 complex.
 
 .so ilmtp_longdouble.n
 .so ilmtp_atomic.n
+
```
sscalpone: Do we have a way to share these from a single file?

PR #513, opened 2018-06-25
Systems without gcc (only clang) cannot build flang

ibara:
Hi --

There's a peculiar bit of the flang compile process where a pair of .F95 files are run through the C preprocessor.

When clang is the C compiler being used, clang silently passes the handling of this preprocessor step to gcc. On systems where clang is the only C compiler available (such as OpenBSD/arm64 by default), this causes the build to fail. I'm not sure why gcc doesn't mind using the C preprocessor to handle .F95 files but that's a different issue.

I'm not certain that passing .F95 files through the C preprocessor is necessarily what is intended here, but this small fix allows systems that have only clang to build flang. This lets flang build and run successfully on OpenBSD/arm64. It should be a no-op with gcc.

Thanks.

### DIFF
--- a/runtime/flang/CMakeLists.txt
+++ b/runtime/flang/CMakeLists.txt
@@ -420,7 +420,7 @@ set(I8_FILES_DIR I8_sources)
 # Fortran files with macros as module names need to be preprocessed. 
 add_custom_command(
   OUTPUT "${I8_FILES_DIR}/ieee_arithmetic.F95"
-  COMMAND "${CMAKE_Fortran_COMPILER}" -E
+  COMMAND "${CMAKE_C_COMPILER}" -E -x c 
   "${CMAKE_CURRENT_SOURCE_DIR}/ieee_arithmetic.F95" -DDESC_I8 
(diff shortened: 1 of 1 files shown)


gklimowicz:
Brian, can you do us the favor of executing an individual contributor license agreement? NVIDIA is requiring that we get these before accepting patches from new contributors (regardless of how small the patch is).

Let me know if you have any questions.

Pointers to the CLAs can be found on the Wiki page for the Flang Community: https://github.com/flang-compiler/flang/wiki/Community

ibara:
Sorry for the delay; just submitted the completed CLA.

ibara:
As noted on the Slack, we can just use `flang -E` directly for this.

ibara:
Apologies for the noise, but it turns out we can't use flang -E. If you don't already have flang1 installed, you can't use flang -E, so there's a bootstrapping problem.

Probably best at this point just to close this, and if this change is desired, pull it in with #521.

gklimowicz:
Closing per ibara.

PR #517, opened 2018-06-29
Pull 2018-06-28T19-57 Recent NVIDIA Changes

mleair:
Fix comment in semant3.c and add openmp_examples tests.


sscalpone:
check_compilation.py is not where it ought to be.  you'll see 11 failing tests in openmp_examples; these might be related to missing check_compilation.py.

PR #521, opened 2018-07-05
Add OpenBSD as a recognized OS

ibara:
Hi --

This pull request adds OpenBSD as a recognized OS in the Flang build process. This may also serve as a starting point for other POSIX operating systems to add themselves to the Flang build scaffolding.

Note that this pull request intentionally does not include changes to libpgmath; there were a lot of changes needed for that, and therefore I think it is better to submit as a follow-up PR after this one is integrated.

Because I am bad at git, this PR also includes the changes I submitted in #513.

Thanks.


ibara:
Hi --

Don't mean to be pushy about this, but it's been about 3 weeks without any comment.
What can I do to help push this along?

Thanks.

gklimowicz:
Sorry for the delays. I'm trying to set up an environment for validating that these will help with the Flang build on OpenBSD. It's not an environment we have in the PGI lab.

ibara:
Sure no problem.
The build may very well fail with this patch though, as it is incomplete--it is missing the bits for libpgmath, which is relatively invasive.

There have also been changes to the build system that you all have made since I opened this, and that has caused me to have to re-alter the OpenBSD additions.

You can see a complete listing of all the patches (not all are suitable for upstreaming to you) here for flang sans libpgmath:
https://github.com/openbsd/ports/tree/master/lang/flang/flang/patches

and libpgmath specifically:
https://github.com/openbsd/ports/tree/master/lang/flang/libpgmath/patches

I can roll all this up into one large patch if that would be best for you all.

ibara:
@iotamudelta Sure. What do you have? I'm sure we can create a combined patch that I can then submit here.

@gklimowicz Thinking about your comment again. I'm not sure what more validation that these patches help with the Flang build on OpenBSD we can get than "I am the OpenBSD developer that ported Flang to OpenBSD, I wrote these patches because Flang does not build on OpenBSD without them, and I have been using these exact patches to build the official OpenBSD package of Flang for many months now." I know of a number of Flang installations on OpenBSD (which, by virtue of the fact that they are installations of my package, use these patches), and there is general interest in making Flang a first-class Fortran compiler on OpenBSD, at least on amd64 and arm64.

sscalpone:
@ibara The PGI environment issue is more about making sure future changes don't break OpenBSD.

ibara:
That makes sense, and I guess I don't have a great answer for that at the moment either. I'll try to think of something.

CMakeLists.txt
```diff
@@ -49,6 +49,20 @@ if( ${TARGET_OS} STREQUAL "Linux" )
...
     return()
   endif()
+elseif( ${TARGET_OS} STREQUAL "OpenBSD")
+  set(OS "OPENBSD")
+  set(OSNAME "OpenBSD")
+  if ( ${TARGET_ARCHITECTURE} STREQUAL "amd64" )
+    set(ARCHNAME x86-64)
+    set(ARCH X86)
+    set(WRDSZ 64)
+    set(TARGET_ARCHITECTURE x86_64)
```
sscalpone: When I was playing around with the port, I did the same thing -- set TARGET_ARCHITECTURE to x86_64.  I'm wondering if we shouldn't factor our TARGET_ARCHITECTURE to be something like FLANG_TARGET_ARCHITECTURE.  The downside is that it might introduce confusion and bugs because people know about TARGET_ARCHITECTURE.  This isn't a call to action; just a discussion point that probably better raised in the mailing list :)

runtime/flangrti/CMakeLists.txt
```diff
@@ -113,6 +113,11 @@ find_library(
   HINTS ${CMAKE_BINARY_DIR}/lib)
 target_link_libraries(flangrti_shared ${LIBPGMATH})
 
+# BSDs need libexecinfo
+if( ${TARGET_OS} STREQUAL "OpenBSD" )
+  target_link_libraries(flangrti_shared ${FLANG_LIBEXECINFO})
```
sscalpone: Will flang-driver need to add libexecinfo to the linker command line?

sscalpone:
I should have mentioned that PGI has OpenBSD and FreeBSD systems available now.

runtime/flangrti/CMakeLists.txt
ibara: No; I do a static link of libexecinfo into flangrti.so
I also set up the flang driver to do all the linking magic behind the scenes. Users simply do something like:
`$ flang myfile.F95`
and everything just works.

See https://github.com/openbsd/ports/blob/master/lang/flang/driver/patches/patch-lib_Driver_ToolChains_OpenBSD_cpp#L54

ibara:
Hooray! That's great news :)

iotamudelta:
@ibara your changes in here look good (as in: could be duplicated for FBSD). I'm solving the missing `gregset_t` a bit differently and haven't done `libexecinfo` (thanks for the idea! :-) ). But your solutions would work for us. So bottom line really is, the changes in this PR can simply be duplicated.

I've spent some time yesterday to look at your `libpgmath` changes, the ones needed for FreeBSD are a subset of that. Similarly, FBSD does have `libaio` functionality, so we don't need those changes.

I'd be very thankful if you could add those duplications, if you don't mind terribly. :-)

iotamudelta:
@sscalpone concerning the FBSD system - I'd be happy to assist there as needed. Please note that at least for me it looks like `libpgmath` means requiring 12.0 or higher - the CPUID bits are not in the headers for 11.1 or lower.

Similarly, my port on FBSD depends on `devel/llvm60`.

ibara:
@iotamudelta Is your port the same one that freshports shows when I search for flang? It looks like that one is somewhat outdated so I'm guessing you have something newer.

I don't mind adding the FreeBSD stuff. But I don't use FreeBSD so I would just be guessing and I don't want to do that. And the port I see appears outdated so I'm not sure how useful it would be.

If you send me an email (my email should be very easy to find) with diffs, I can add them. Or however you want to coordinate is fine by me.

iotamudelta:
@ibara yeah, I have an update in the pipeline. Could you ping me on the email that is visible on freshports? I'll provide diffs tonight! Thanks!

PR #524, opened 2018-07-06
Pull 2018-07-06T16-11 Recent NVIDIA Changes

mleair:


sscalpone:
??


Fix a bug related to intrinsic and variable name overloading.
A test case with an intrinsic routine reference in a host routine
that also has contained routine that has a variable with the same
name as the intrinsic fails to compile.  The fix is to check for
an explicit INTRINSIC attribute when processing the contained
routine variable to determine if the two references refer to the
same symbol.

Also add a test to exercise the bug fix.

# Merged CL 153854
*** populateTests errors:
ERROR: Suite openmp_examples doesn't exist!

*** populateFlang errors:
populateFlang: source file not found: P4/flang/test/f90_correct/inc/fs23823.mk
populateFlang: source file not found: P4/flang/test/f90_correct/inc/fs23824.mk
populateFlang: 2 errors encountered

test for fs25895 -- intrinsic vs. variable name reference

PR #529, opened 2018-07-10
runtime: PoC: generate call to LLVM intrinsic instead of calling runtime functions when possible; handling NINT for a beginning

pawosm-arm:


sscalpone:
The overall flow of this pull request is like this: 
 
If(op == IL_NINT) {
    op = “_mth_i_nint”;
}
….
If(op == “_mth_i_nint”) {
   op == llvm.intrinsic.cvt;
}
 
My preference is to skip the intermediate conversion to _mth_i_nint.
 
If(op == IL_NINT) {
op = llvm.intrinsic.cvt
}
 
Advantages to skipping the intermediate conversion:
Availability of IL_NINT for folding
Code is less complex; does not use strcmp and magic routine name
 
Assuming we want to map as many flang ILIs directly to llvm intrinsics, it would be nice to have a good model in place.

bryanpkc:
Closing stale PR.

PR #530, opened 2018-07-10
AArch64: disable test cases invalid on aarch64

pawosm-arm:


gklimowicz:
@pawosm-arm These tests no longer seem to be failing in my builds of Arm and "run check-flang".

Can we close this pull request, or is there something more for us to do?

pawosm-arm:
Indeed.

PR #532, opened 2018-07-11
runtime: remove duplicate files

pawosm-arm:


sscalpone:
I am working with pgmath to understand if these files belong in flang or pgmath.  I think they belong in flang but the jury is still out.

pawosm-arm:
Refreshed in #695

PR #536, opened 2018-07-17
libpgmath: move includes to separate include directory, plus remove unfortunate duplication of mthdecls.h file

pawosm-arm:


sscalpone:
@pawosm-arm  I'm taking a different approach.  I am removing the references to mthdecls from runtime flang and flangrti.  Then mthdecls will only be needed in libpgmath.

In any case, mthdecls is supposed to be private to libpgmath so it is not appropriate to install it.  The fact that the same contents are shared across flang and libpgmath needs to be rewritten to separate this dependency.  (As I said, my approach is to remove the dependency, by, for example, replacing LONG_LONG_T with int64_t.)

sscalpone:
I think this is done.  The only copy of mthdecls is in libpgmath and the only references are in libpgmath.

pawosm-arm:
None of the goals of this pull request were achieved: The include directory is still not exposed for libpgmath and there are still two copies of mthdecls.h:
https://github.com/flang-compiler/flang/blob/master/runtime/libpgmath/lib/x86_64/mthdecls.h
and
https://github.com/flang-compiler/flang/blob/master/runtime/libpgmath/lib/common/mthdecls.h
...so I'm going to refresh this one as now it is haunted by merge conflict.

sscalpone:
The include directory for libpgmath is not supposed to be exposed.  Also, I expect some major refactoring of libpgmath to happen sometime before the end of the year.

pawosm-arm:
Ok, closing it as it doesn't even affect our arch.

PR #538, opened 2018-07-18
Integrated and re-organized changes to support ARM64.

simoatze:
This patch eliminates all the duplicates present in the Cavium PR and introduces support for ARM64 using the existing generic plus some architecture specifics for ARM.


runtime/libpgmath/lib/CMakeLists.txt
```diff
@@ -147,16 +147,46 @@ elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "ppc64le")
...
+  # Definitions and compiler flags for level 2 directories
+  set_property(GLOBAL APPEND PROPERTY "DEFINITIONS_ARM64_L2" ${DEFINITIONS_L2})
+  set_property(GLOBAL APPEND_STRING PROPERTY "FLAGS_ARM64_L2" ${FLAGS_L2})
+  # aarch64
 else()
   set(DEFINITIONS_L1 HOST_LINUX MAXCPUS=256 MAXCPUSL=8 MAXCPUSR=8 TARGET_LINUX
     TARGET_LINUX_64 TARGET_LLVM TARGET_LLVM_64 TARGET_LLVM_LINUX64
     TARGET_LINUX_GENERIC TARGET_64 PG_PIC)
-  set(FLAGS_L1 "-O3 -mfma -ffp-contract=fast ")
+  set(FLAGS_L1 "-O3 -ffp-contract=fast ")
```
pawosm-arm: A side note is that although this change (and one below) reduces amount of warnings, it is not related to aarch64 port

runtime/libpgmath/CMakeLists.txt
```diff
@@ -46,23 +46,50 @@ if (LIBPGMATH_STANDALONE_BUILD)
...
+#   string(REPLACE "-fno-tree-slp-vectorize" "-ftree-slp-vectorize" CMAKE_C_FLAGS ${CMAKE_C_FLAGS})
+#   string(REPLACE "-fno-tree-slp-vectorize" "-ftree-slp-vectorize" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+#   string(REPLACE "-fno-reorder-blocks" "-freorder-blocks" CMAKE_C_FLAGS ${CMAKE_C_FLAGS})
+#   string(REPLACE "-fno-reorder-blocks" "-freorder-blocks" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+#   string(REPLACE "-fno-toplevel-reorder" "-ftoplevel-reorder" CMAKE_C_FLAGS ${CMAKE_C_FLAGS})
+#   string(REPLACE "-fno-toplevel-reorder" "-ftoplevel-reorder -felide-constructors" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+#   string(REPLACE "-std=c++11" "-std=gnu++11" CMAKE_CXX_FLAGS_DEBUG ${CMAKE_CXX_FLAGS_DEBUG})
+#   string(REPLACE "-std=c++11" "-std=gnu++11" CMAKE_CXX_FLAGS_RELWITHDEBINFO ${CMAKE_CXX_FLAGS_RELWITHDEBINFO})
+# endif()
+
```
pawosm-arm: AFAIK this can be un-commented and fixed.

simoatze: Yes, I already fixed this in my internal patch.

runtime/libpgmath/lib/aarch64/cos/CMakeLists.txt
```diff
@@ -23,26 +23,28 @@ set(SRCS)
 # Set source files, compiler flags and definitions
 if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "x86_64")
```
pawosm-arm: why do we have this x86_64 part in CMakeLists.txt down below aarch64 directory?

simoatze: Good catch, I'll clean that up.

runtime/libpgmath/lib/common/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
 if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Windows")
-  if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "x86_64|ppc64le" AND NOT ${LIBPGMATH_WITH_GENERIC})
+  if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "x86_64|ppc64le|aarch64" AND NOT ${LIBPGMATH_WITH_GENERIC})
```
pawosm-arm: I assume next step will be to move those to generic?

pawosm-arm: ok, so common is a place for things that are shared among known architectures and not built for generic...

pawosm-arm: powi, sincos - shouldn't they move to generic at some point?
sincosf - shouldn't it move to x86_64 at some point?

simoatze: powi and sincos probably at some point when we cleanup the rest of the common directory (e.g. we move headers under `include/`, etc.). Sincosf yes it should be moved.

simoatze: powi and sincos are generic but they also get used for x86_64, power and arm. I need to think about it, because if we move them under generic then for example for x86_64 I have to add the generic dir when I build and add the logic to only build `pow` and `sincos` and ignore the rest of generic (since everything is actually under `x86_64`. I'll leave as they are for now and think about it, in a future commit will fix that plus create the `include` folder. I moved `sincosf` under `x86_64` though.

runtime/libpgmath/CMakeLists.txt
sscalpone: Why does this patch require, for example, that -std=gnu++11 be used instead of c++11?

runtime/libpgmath/lib/common/arm64intrin.h
```diff
@@ -0,0 +1,2823 @@
...
+    return xlrd;
+  }
+
+  inline operator int32x4_t() const {
+    return nird;
+  }
+
+  inline operator int64x2_t() const {
+    return nlrd;
+  }
```
pawosm-arm: This won't build with gcc8. I had do wrap it like this to make the build succeed:
```
#if __GNUC__ < 8
  inline operator int32x4_t() const {
    return nird;
  }

  inline operator int64x2_t() const {
    return nlrd;
  }
#endif
```

gklimowicz:
Closed per request for Paul Osmialowski

PR #542, opened 2018-07-24
Pull 2018-07-23T18-28 Recent NVIDIA Changes

gklimowicz:


sscalpone:
AVC 513?
Add pure AVX513 single and double precision implementations

PR #546, opened 2018-07-25
Loosen ties to x86

pawosm-arm:
Two commits in this pull request: in order to increase readability, renaming x86 files is done by separate commit.


sscalpone:
Looks like a renaming change except for the change to the file
tools/shared/pragma.c

Did you test this change with another platform, say Arm, for example?

pawosm-arm:
Tested on Arm, builds fine, passes check-all tests. The pragma.c file isn't the only place where check for TARGET_X86 was removed. The intent is to make mach.h fit to the architecture so #ifdefs won't be needed anymore (more maintainable code!) in all of those places of inclusion.

tools/shared/mach.c
```diff
@@ -76,7 +39,7 @@ set_mach(X86TYPE *mach, int machtype)
  *  -tp may not be appropriate for ARM; so machtype may refer to
  *  ARM attriubutes rather than a TP_ value.
  */
-#if !defined(TARGET_LLVM_ARM)  /* { */
+#if defined(TARGET_X86)  /* { */
```
schweitzpgi: Heads-up: all targets that are not ARM != x86 targets.

tools/shared/mach.c
```diff
@@ -23,49 +23,12 @@
...
-                  "skylake-64",
-                  "?"};
-
-char *version_name[] = {
-    "0",         "py",       "px",       "p5",         "athlon",
-    "p6",        "athlonxp", "piii",     "k8",         "p7",
-    "k8e",       "piv",      "gh",       "core2",      "penryn",
-    "shanghai",  "istanbul", "nehalem",  "bulldozer",  "sandybridge",
-    "ivybridge", "haswell",  "larrabee", "piledriver", "zen",
-    "knl",       "skylake"};
```
schweitzpgi: Where did these arrays end up?

pawosm-arm: Nowhere, they don't seem to be referenced by anyone.

tools/shared/mach.c
pawosm-arm: comment updated

pawosm-arm:
rebased, refreshed.

bryanpkc (approved):
LGTM

Issue #424, opened 2018-03-22
"constant expression type mismatch" for valid code


dev-zero:
given the following two files

`dbcsr_data_types.f90`:
```fortran
module dbcsr_data_types
   TYPE dbcsr_data_obj
      TYPE(dbcsr_data_area_type), POINTER           :: d => Null()
   END TYPE dbcsr_data_obj

   TYPE dbcsr_mempool_type
      TYPE(dbcsr_mempool_entry_type), POINTER :: root => Null()
      INTEGER                                 :: capacity = 1
   END TYPE dbcsr_mempool_type

   TYPE dbcsr_mempool_entry_type
      TYPE(dbcsr_data_obj)                    :: area
      TYPE(dbcsr_mempool_entry_type), POINTER :: next => Null()
   END TYPE dbcsr_mempool_entry_type

   TYPE dbcsr_memtype_type
      TYPE(dbcsr_mempool_type), POINTER :: pool => Null()
   END TYPE dbcsr_memtype_type

   TYPE(dbcsr_memtype_type), PARAMETER :: dbcsr_memtype_default = dbcsr_memtype_type(pool=Null())

   TYPE dbcsr_data_area_type
      INTEGER                                  :: ref_size = -1
      INTEGER                                  :: refcount = -1
      INTEGER                                  :: id = -1
      TYPE(dbcsr_memtype_type)                 :: memory_type = dbcsr_memtype_default
      INTEGER                                  :: data_type = -1
   END TYPE dbcsr_data_area_type

end module
```

`dbcsr_data_types_user.f90`:
```fortran
module a
   use dbcsr_data_types, only: dbcsr_memtype_type

   type foo
      type(dbcsr_memtype_type) :: val = dbcsr_memtype_type()
   end type
end module
```

trying to compile them using

```bash
flang -c dbcsr_data_types.f90
flang -c dbcsr_data_types_user.f90
```

gives

```
/tmp/dbcsr_data_types_user-985010.ll:9:47: error: constant expression type mismatch
@_a_10_ = global %struct_a_10_ < { [16 x i8]  [i8 0,i8 0,i8 0,i8 0,i8 0,i8 0,i8 0,i8 0]  } > , align 64
                                              ^
1 error generated.
```
for `dbcsr_data_types_user.f90`.

Works with gfortran-4.8 and gfortran-8.

pawosm-arm:
[...]

PR #561, opened 2018-08-05
flang2: corrected fix for #424

pawosm-arm:


tools/flang2/flang2exe/llassem.cpp
```diff
@@ -1239,16 +1239,16 @@ process_dsrt(DSRT *dsrtp, ISZ_T size, char *cptr, bool stop_at_sect,
...
         }
         ptrcnt = 0;
       } else if (!i8cnt) {
         if (!first_data && skip_size)
           fprintf(ASMFIL, ", ");
         ptr = put_next_member(ptr);
-        fprintf(ASMFIL, "[");
+        fprintf(ASMFIL, "zeroinitializer ");
+        free(cptrCopy);
+        return dsrtp;
```
schweitzpgi: I don't follow what changed in the compiler to need to move this block of code around. Do you have an example?

pawosm-arm: Ok, what changed in the compiler:
Step 1: by mistake I placed wrong commit into my pull request for bug #424 
Step 2: my pull request was accepted and landed on master branch
Step 3: it turned out, bug #424 is still not fixed
Step 4: I found out my mistake
Step 5: I opened new pull request with a commit that moves my change to the correct place
Bug #424 comes with minimal reproducer.

schweitzpgi: Thanks!

PR #573, opened 2018-08-25
libpgmath: make it clang friendly; addresses bug #557

pawosm-arm:


runtime/libpgmath/lib/common/acos/fma3/ssacos.cpp
```diff
@@ -77,7 +77,11 @@ float __fss_acos_fma3(float const a)
         _sq = _mm_setr_ps(0.0f, sq, 0.0f, 0.0f);
         p1 = _mm_fmadd_ps(p, _x2_x, F);
 
+#if defined(__clang__) && defined(TARGET_LINUX_ARM64)
+        __m128 pi_mask = (__m128)((long double)_mm_cmpgt_epi32(ZERO, (__m128i)((long double)_a)));
```
sscalpone: Intel claims the type of _mm_cmpgt_epi32 is __m128i.  You should change pi_mask 

__m128i _mm_cmpgt_epi32 (__m128i a, __m128i b)
Synopsis
__m128i _mm_cmpgt_epi32 (__m128i a, __m128i b)
#include <emmintrin.h>
Instruction: pcmpgtd xmm, xmm
CPUID Flags: SSE2
Description
Compare packed 32-bit integers in a and b for greater-than, and store the results in dst.

pawosm-arm: The only places pi_mask is used expect it to be __m128, e.g. line below it's used as first argument to `_mm_and_ps()` call, which is declared as such:
```
__m128 _mm_and_ps (__m128 a, __m128 b)
```
I see no need for deferring this conversion.

sscalpone:
This change needs documentation.  What is the problem it is solving? How does this solve the problem?  Are there alternatives? What is the long-term prognosis for this code?

runtime/libpgmath/lib/common/acos/fma3/ssacos.cpp
sscalpone: If pi_mask is declared properly, do you still need the two casts?  Is _mm_cmpgt_epi32 a pre-processor macro with its own casts?

pawosm-arm: Yes, `_mm_cmpgt_epi32` is a pre-processor macro with its own cast to `__m128i`. So in places where `__m128i` value is needed, no casting is done, otherwise, explicit cast is required. As I explained above, in this case a cast to `__m128` is required.

pawosm-arm:
Added more verbose commit message.

d-parks:
Would using the intrinsic `_mm_castsi128ps`  and `_mm_castps_si128` be a solution that is syntactically more elegant?

__m128i _mm_castps_si128 (__m128 a)
Synopsis
__m128i _mm_castps_si128 (__m128 a)
#include <emmintrin.h>
CPUID Flags: SSE2
Description
Cast vector of type __m128 to type __m128i. This intrinsic is only used for compilation and does not generate any instructions, thus it has zero latency.


__m128 _mm_castsi128_ps (__m128i a)
Synopsis
__m128 _mm_castsi128_ps (__m128i a)
#include <emmintrin.h>
CPUID Flags: SSE2
Description
Cast vector of type __m128i to type __m128. This intrinsic is only used for compilation and does not generate any instructions, thus it has zero latency.

CarolineConcatto:
Hi @d-parks so I looked in the code and I don't know if this patch is the best place to replace the cast for _mm_castsi128_ps or _mm_castps_si128. Maybe another patch could be created to do that.
This code has a goal to make some fix to arm64 machine. It adds some extra lines, based on what was already in the code.

For instance this line is already in the code:
1) _m128 pi_mask = (__m128)_mm_cmpgt_epi32(ZERO, (__m128i)_a);

the patch only adds this extra for arm64 machines:
2)#if defined(__clang__) && defined(TARGET_LINUX_ARM64)
3)__m128 pi_mask = (__m128)((long double)_mm_cmpgt_epi32(ZERO, (__m128i)((long double)_a)));
So if I change the third line, it would be strange not to change the first one. 

In my point of view, changing all the lines that the cast could be done is outside the scope of this patch, no?

OBS: As far as I understand mm_cmpgt_epi32 returns __m128i. I only brought this as an example, but there are other lines that follow the same pattern

@d-parks Is it ok we leave the patch the way it is?

d-parks:
Hi  @CarolineConcatto 

Yes, let's leave the mod as proposed.

Dave

kiranchandramohan:
@d-parks Can you approve if no mods are needed for this patch?

bryanpkc:
LGTM as well.

PR #589, opened 2018-09-13
Fix clang++ build after recent changes

pawosm-arm:


tools/flang2/flang2exe/iliutil.cpp
```diff
@@ -2093,7 +2093,7 @@ reciprocal_division(int n, INT divisor, int sgnd)
   } else {
     INT64 twol;
     int shpre = 0;
-    INT64 one_64 = {0x0, 0x1}, divisor64 = {0x0, udiv};
+    INT64 one_64 = {0x0, 0x1}, divisor64 = {0x0, (signed)udiv};
```
sscalpone: Generally we are attempting to become more type safe by removing casts.  Is there a way to achieve a fix without introducing a cast?

pawosm-arm: There is.

tools/flang2/flang2exe/iliutil.cpp
```diff
@@ -2102,7 +2102,7 @@ reciprocal_division(int n, INT divisor, int sgnd)
   } else {
     INT64 twol;
     int shpre = 0;
-    INT64 one_64 = {0x0, 0x1}, divisor64 = {0x0, udiv};
+    INT64 one_64 = {0x0, 0x1}, divisor64 = {0x0, divisor};
```
sscalpone: That's fairly subtle!  You might want to merge the three if(sgnd) statements into one for clarity.

pawosm-arm: ok

pawosm-arm:
Although I like my patch more, there's already patch on master that fixes clang++ issue: https://github.com/flang-compiler/flang/commit/ed74454a0889ca5ffd3a99d05d82fd206ae98f26
hence, closing it.

PR #622, opened 2018-11-26
Add explicit braces to avoid dangling else problem

kiranchandramohan:
The dangling else problem was found to cause some failures in
a proprietary test suite.


pawosm-arm (approved):

pawosm-arm:
@kiranchandramohan will you be able to resolve those conflicts?

PR #623, opened 2018-11-26
[CMake][pgmath] Fix for building in lib directory

kiranchandramohan:
If pgmath is built in a directory containing /lib/ in the path then
the code for constructing a library name will not work correctly.
The patch fixes this issue by choosing the last occurrence of /lib/.


kiranchandramohan:
Thanks @gklimowicz.

PR #647, opened 2019-01-03
Introduce buffered I/O and replace getc with buffered read

pawosm-arm:
This commit introduces 32k buffer for file I/O operations
and 32k buffer for fgetc() replacemnt routines.

This is follow-up of my work presented in pull request #642.

Some performance gain is observed and 'perf' trace does not
show getc operations at all, shifting the burden to actual
I/O system routines. In order to improve performance of these,
32k buffer is attached to every opened file by calling
setvbuf() function.

Due to suspicious way of passing pointers around in flang runtime,
all of the I/O buffers had to be allocated statically. In order to
prevent memory bloat, only limited number of simultaneously opened
files will be buffered, file descriptor ID is used for buffer
selection which is a thread-safe solution.

The test case I was using is following:
```
! *********************************************************
        program main

        implicit none
!       ---------------------------------------------------
        character(len=500) :: cart
        real(kind=8) :: t1,t2

!       ---------------------------------------------------
        open(unit=9,status='old',file='my_file.txt')
        open(unit=10,file='my_new_file.txt')
        call cpu_time(t1)
        do
!               read each line
                read(9,fmt='(A)') cart
!       ************************************************************
!       ************************************************************
!                       convert  process
!       ************************************************************
!       ************************************************************

                if(cart(1:4)=='/end') then
                        write(10,*) 'this is the end!'
                        exit
                else
                        write(10,*) cart
                endif
        enddo
        call cpu_time(t2)
        close(unit=9)
        close(unit=10)

        print*,' write and read :',t2-t1




!       ---------------------------------------------------

        end program main
! *********************************************************
```
[my_file.txt.gz](https://github.com/flang-compiler/flang/files/2724306/my_file.txt.gz)

Compiled with gfortran it gives much better timing results that when compiled with flang. My patch improves the timing of flang compiled program.

### DIFF
--- a/runtime/flang/backspace.c
+++ b/runtime/flang/backspace.c
@@ -74,20 +74,23 @@ _f90io_backspace(__INT_T *unit, __INT_T *bitv, __INT_T *iostat, int swap_bytes)
 
   if (f->nonadvance) {
     f->nonadvance = FALSE;
+    FIO_FCB_INVALIDATE_GETC_BUFFER(f, return __io_errno());
 #if defined(WINNT)
-    if (__fortio_binary_mode(f->fp))
-      __io_fputc('\r', f->fp);
+    if (__fortio_binary_mode(f->__io_fp))
+      __io_fputc('\r', f->__io_fp);
 #endif
-    __io_fputc('\n', f->fp);
-    if (__io_ferror(f->fp))
+    __io_fputc('\n', f->__io_fp);
+    if (__io_ferror(f->__io_fp))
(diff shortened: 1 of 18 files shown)


pawosm-arm:
Anticipating your questions, setvbuf() is a very standard stdio.h function and you can find its documentation in Linux man pages and Visual Studio reference pages as well.

bryanpkc:
We agreed to close this one at the last call.

PR #658, opened 2019-01-20
Fix in preprocessing for Flexi app

kiranchandramohan:
The preprocessor in fortran detects 1.NE.PP_NUM as a number.
The fix uses the knowledge that a constant number can have
atmost one dot character in it. If more than one dot character
is seen then the scanning has gone too far ahead.
This fix removes PP_NUM from the number detected and allows
for substitution of PP_NUM with the defined preprocessor
value.

### DIFF
--- a/test/llvm_ir_correct/dot_preprocess.f90
+++ b/test/llvm_ir_correct/dot_preprocess.f90
@@ -1,0 +1,11 @@
+! RUN: %flang -cpp -S -emit-llvm %s -o - | FileCheck %s
+#define PP_N 10
+subroutine pp(n)
+implicit none
+integer :: n
+if (n-1.ne.PP_N) then
+  print *, 'Success'
+end if
+end subroutine pp
+
+! //CHECK: {{.*}}  = icmp eq i32 {{.*}}, 11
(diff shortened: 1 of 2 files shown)


gklimowicz (approved):
This works fine on OpenPOWER.

kiranchandramohan:
@shivaramaarao ?

bryanpkc:
This might be a stupid question, but why does the preprocessor think that 1.NE (or any `$num.$ident`) could be a valid number?

gklimowicz:
> This might be a stupid question, but why does the preprocessor think that 1.NE (or any `$num.$ident`) could be a valid number?

This could be an artifact of the preprocessor being shared between the PGI Fortran and C/C++ compilers (and hex floating-point constants). It's also possible that it is there to handle `kind` specifiers on real constants (e.g., `1.e10_REAL64`). (Changes to this file date back to 2009.)

kiranchandramohan:
@bryanpkc please let us know if you need further explanation or if you have concerns.

bryanpkc (approved):
LGTM.

shivaramaarao (approved):
LGTM

PR #659, opened 2019-01-20
Changes to support transpose intrinsic during initialization

kiranchandramohan:
Transpose intrinsic is a special case of the reshape intrinsic
with the order parameter set to /2,1/. We reuse the functions
for reshape intrinsic with minor additions wherever possible.

### DIFF
--- a/test/f90_correct/inc/transpose_init.mk
+++ b/test/f90_correct/inc/transpose_init.mk
@@ -1,0 +1,22 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+EXE=transpose_init.$(EXESUFFIX)
+
+build:  $(SRC)/transpose_init.f90
+	-$(RM) transpose_init.$(EXESUFFIX) core *.d *.mod FOR*.DAT FTN* ftn* fort.*
+	-$(RM) $(OBJ)
+	-$(CC) -c $(CFLAGS) $(SRC)/check.c -o check.$(OBJX)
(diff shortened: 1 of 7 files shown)


test/f90_correct/inc/transpose_init.mk
```diff
@@ -0,0 +1,32 @@
+#
+# Copyright (c) 2016, NVIDIA CORPORATION.  All rights reserved.
```
kiranchandramohan, line 2: Remove copyright and switch to new LLVM license. Here and in all new files.

tools/flang1/flang1exe/semutil2.c
```diff
@@ -9857,11 +9880,16 @@ eval_reshape(ACL *arg, DTYPE dtype)
...
 
   if (orderarg == NULL) {
-    if (src_sz == dest_sz) {
-      return srclist;
-    }
-    for (i = 0; i < rank; i++) {
-      order[i] = i;
+    if (transpose) {
+        order[0] = 1;
+        order[1] = 0;
```
bryanpkc, line 9885: Does the semantic analysis for this initializer need to check that the ranks of the input and output are both 2? Or is that check already performed somewhere else?

kiranchandramohan: Yes, sema does that check here for the input. https://github.com/flang-compiler/flang/blob/27fd9b2ec77be73c9bd90856056dab33af987c39/tools/flang1/flang1exe/semfunc.c#L7058
The output is not strictly checked probably even for non-constant transpose.

pawosm-arm:
@bryanpkc can this one be merged?

tools/flang1/flang1exe/semutil2.c
bryanpkc: Thanks.

bryanpkc (approved):

PR #672, opened 2019-02-13
Support isnan() GNU extension

mvinay:


pawosm-arm:
A small test case would help in well reception of this PR.

mvinay:
> A small test case would help in well reception of this PR.

Done.

tools/flang1/flang1exe/semant.c
```diff
@@ -866,6 +866,15 @@ semant1(int rednum, SST *top)
     sem.is_hpf = scn.is_hpf;
     sem.alloc_std = 0;
     sem.p_dealloc_delete = NULL;
+
+    if (gbl.currsub != 0 && sem.pgphase > PHASE_USE) {
+      if (flg.allow_gnu_extensions &&
+          is_inbuilt_module(SYMNAME(sem.mod_sym)) == 0) {
+      apply_gnu_ext();
```
kiranchandramohan: align code here

test/f90_correct/gnu_isnan.f90
```diff
@@ -0,0 +1,13 @@
+! RUN: %flang -S -emit-llvm %s -o - | FileCheck %s
+
+program test_nan
+  use ieee_arithmetic
```
kiranchandramohan: Thanks for adding the testcase. Can you add one more without ieee_arithmetic.

kiranchandramohan:
Thanks Vinay.

Could you check why the following test fails with this commit.
```
program mn
  call check_r()
contains
  subroutine check_r()
    use ieee_arithmetic,only:ieee_selected_real_kind
    if (ieee_selected_real_kind(r=32)/=ieee_selected_real_kind(4,32)) &
      stop 'FAIL check_r'
  end subroutine
end
```

mvinay:
> Thanks Vinay.
> 
> Could you check why the following test fails with this commit?
> 
> ```
> program mn
>   call check_r()
> contains
>   subroutine check_r()
>     use ieee_arithmetic,only:ieee_selected_real_kind
>     if (ieee_selected_real_kind(r=32)/=ieee_selected_real_kind(4,32)) &
>       stop 'FAIL check_r'
>   end subroutine
> end
> ```

It has been fixed. Sorry for the delayed response.

kiranchandramohan:
We faced a few issues while testing this patch.

-> While building the compiler with this patch, it was felt that this module should be built before any Fortran file in the compiler is built. But this dependency is not expressed in the Makefile. Since we know that isnan is not needed for building the Flang compiler we switched isnan OFF in the makefile of the compiler.

-> Our libraries reported an increase in runtime for one of their testcases with this patch. Unfortunately, this testcase cannot be shared with you. There are a lot of "use only" statements in this testcase. Removing the only qualifier brought down the runtime. Note that this was a huge increase from seconds to minutes.

-> The Quantum Espresso application was failing to build with the isnan patch. It fails with the following error.
```
flang -O3 -ffp-contract=fast -mcpu=native -D__PGI -D__ARM_LIB -I/home/appuser/arm-hpc-compiler-0.0-OpenMPI-3.1.2/quantumespresso/Application//include -I/home/appuser/arm-hpc-compiler-0.0-OpenMPI-3.1.2/quantumespresso/Application//FoX/finclude -I../include/ -I../../iotk/src -I../../Modules -I../../FFTXlib -I../../LAXlib -I../../UtilXlib -I../../KS_Solvers/Davidson -I../../KS_Solvers/CG -c offset_atom_wfc_tmp.f90 -o offset_atom_wfc.o
F90-S-0038-Symbol, npol, has not been explicitly declared (add_vuspsi_tmp.f90)
```

mvinay:
> We faced a few issues while testing this patch.
> 
> -> While building the compiler with this patch, it was felt that this module should be built before any Fortran file in the compiler is built. But this dependency is not expressed in the Makefile. Since we know that isnan is not needed for building the Flang compiler we switched isnan OFF in the makefile of the compiler.
> 
> -> Our libraries reported an increase in runtime for one of their testcases with this patch. Unfortunately, this testcase cannot be shared with you. There are a lot of "use only" statements in this testcase. Removing the only qualifier brought down the runtime. Note that this was a huge increase from seconds to minutes.
> 
> -> The Quantum Espresso application was failing to build with the isnan patch. It fails with the following error.
> 
> ```
> flang -O3 -ffp-contract=fast -mcpu=native -D__PGI -D__ARM_LIB -I/home/appuser/arm-hpc-compiler-0.0-OpenMPI-3.1.2/quantumespresso/Application//include -I/home/appuser/arm-hpc-compiler-0.0-OpenMPI-3.1.2/quantumespresso/Application//FoX/finclude -I../include/ -I../../iotk/src -I../../Modules -I../../FFTXlib -I../../LAXlib -I../../UtilXlib -I../../KS_Solvers/Davidson -I../../KS_Solvers/CG -c offset_atom_wfc_tmp.f90 -o offset_atom_wfc.o
> F90-S-0038-Symbol, npol, has not been explicitly declared (add_vuspsi_tmp.f90)
> ```

My aim was to provide an inbuilt module for gnu extensions so that we would not need to change the internals for other similar extensions. But, adding the implicit use statement for a module is affecting the sema checks. However, I have created an alternate solution for isnan() in https://github.com/flang-compiler/flang/pull/773. Please look at it.

pawosm-arm:
@shivaramaarao would you be able to take ownership of this (namely, publish this PR again and close this one)?

pawosm-arm:
Overridden by PR #1269 so closing this one.

PR #689, opened 2019-03-20
Fix for segfault in multiple entries OpenMP functions (fixes #638 )

### LINKED ISSUE #638: OpenMP: OMP CRITICAL segfaults when subroutine has ENTRY statements
OMP CRITICAL segfaults when subroutine has ENTRY statements. Minimal reproducer can look like following:
```
      SUBROUTINE SUB1
      IMPLICIT NONE
      INTEGER X
      X = 0

      ENTRY SUB2
      WRITE (*, *) "HELLO"
!$OMP CRITICAL
      X = X + 1
!$OMP END CRITICAL
      END SUBROUTINE SUB1

      PROGRAM CRITICAL

      CALL SUB1
      CALL SUB2

      END
```
Rationale: every subroutine containing a CRITICAL section calls __kmpc_global_thread_num() in order to obtain gtid needed by __kmpc_critical() function. Passing invalid gtid to __kmpc_critical() causes segfault as it tries to use gtid value as an offset into threads array - invalid offset results in reading outside the array. Unfortunately, __kmpc_global_thread_num() is called only for the main entry of the subroutine, it's never called for any entry defined by the ENTRY statement.

WORKAROUND: as outlinded functions always get proper gtid as their first parameter, the easiest workaround is to force outlining of the CRITICAL section by enclosing it into OMP PARALLEL region.

HACK: since proposed workaround was rejected due to sheer size of the Fortran codebase that would need to be changed, following hack was successfully tried:
```
--- a/tools/flang2/flang2exe/expsmp.cpp
+++ b/tools/flang2/flang2exe/expsmp.cpp
@@ -3012,7 +3012,15 @@ exp_mp_func_prologue(void)
     }
   }

-  ll_save_gtid_val(bih);
+  if (has_multiple_entries(GBL_CURRFUNC)) {
+    if (bih)
+      ll_save_gtid_val(bih);
+    for (func = gbl.entries; func != NOSYM; func = SYMLKG(func)) {
+      ll_save_gtid_val(BIH_NEXT(BIHNUMG(func)));
+    }
+  } else {
+    ll_save_gtid_val(bih);
+  }
 }

 static void
```

pawosm-arm:
This patch adds missing check for multiple entries in a function
that builds prologue for a Fortran function (or subroutine) that
contains at least one OpenMP directive.

Since ll_save_gtid_val() function usually used for single-entry
Fortran functions can handle only one entry, processing of multiple
entries is handled similarly in a for loop.

A short test case which validates this patch is also added
to the commit.

### DIFF
--- a/tools/flang2/flang2exe/expsmp.cpp
+++ b/tools/flang2/flang2exe/expsmp.cpp
@@ -3029,7 +3029,17 @@ exp_mp_func_prologue(bool process_tp)
     }
   }
 
-  ll_save_gtid_val(bih);
+  if (has_multiple_entries(GBL_CURRFUNC)) {
+    if (bih)
+      ll_save_gtid_val(bih);
+    for (func = gbl.entries; func != NOSYM; func = SYMLKG(func)) {
+      bih = expb.curbih = findEnlabBih(func);
+      if (bih)
+        ll_save_gtid_val(bih);
+    }
+  } else {
+    ll_save_gtid_val(bih);
+  }
 }
 
 static void
--- a/test/llvm_ir_correct/critical_entries.f90
+++ b/test/llvm_ir_correct/critical_entries.f90
@@ -1,0 +1,16 @@
+! RUN: %flang -fopenmp -S -emit-llvm %s -o - | FileCheck %s
+subroutine sub1
+  implicit none
+  integer :: x
+
+  write (*, *) "HELLO1"
+  entry sub2
+  x = 0
+  write (*, *) "HELLO2"
+!$OMP CRITICAL
+  x = x + 1
+!$OMP END CRITICAL
+end subroutine
+! CHECK: call i32 @__kmpc_global_thread_num
+! CHECK: call i32 @__kmpc_global_thread_num
+


tools/flang2/flang2exe/expsmp.cpp
```diff
@@ -3012,7 +3012,15 @@ exp_mp_func_prologue(void)
     }
   }
 
-  ll_save_gtid_val(bih);
+  if (has_multiple_entries(GBL_CURRFUNC)) {
+    if (bih)
+      ll_save_gtid_val(bih);
+    for (func = gbl.entries; func != NOSYM; func = SYMLKG(func)) {
+      ll_save_gtid_val(BIH_NEXT(BIHNUMG(func)));
```
bryanpkc, line 3019: Please reuse `findEnlabBih` here instead of writing `BIH_NEXT(BIHNUMG(func))`.

bryanpkc: I think you need to update `expb.curbih` here as well, similar to line 3004?

pawosm-arm: I agree, this patch needs refresh. Even the test case is in a wrong directory in which it is never executed.

pawosm-arm:
@bryanpkc Thanks for your comments. I've just uploaded a new version.

tools/flang2/flang2exe/expsmp.cpp
bryanpkc: Thanks.

bryanpkc (approved):
LGTM. Tested in our downstream compiler.

kiranchandramohan:
@pawosm-arm We need an approval from AMD (as per process someone has to test on x86_64) for merging code changes. 
Is this patch OK with @shivaramaarao, @SouraVX ?

Issue #461, opened 2018-05-04
[OpenMP] TeaLeaf_ref: local variable became shared in nested parallel region


pawosm-arm:
While running TeaLeaf_ref on larger number of threads unexpected behavior resulting in results validation errors were observed. It turned out, one of local variables (a pointer to a function) became shared within parallel region causing wrong subroutine being called in effect.

Outer region:
```
!$OMP PARALLEL PRIVATE(tile_offset)
!$OMP DO
  DO t=1,tiles_per_task
    SELECT CASE (face)
    CASE (CHUNK_LEFT, CHUNK_RIGHT)
      tile_offset = (chunk%tiles(t)%bottom - chunk%bottom)*depth
    CASE (CHUNK_BOTTOM, CHUNK_TOP)
      tile_offset = (chunk%tiles(t)%left - chunk%left)*depth
      IF (tile_offset .NE. 0) THEN
        tile_offset = tile_offset + depth*depth
      ENDIF
    CASE DEFAULT
      CALL report_error("pack.f90","Invalid face pased to buffer packing")
    END SELECT

    IF (chunk%tiles(t)%tile_neighbours(face) .NE. EXTERNAL_FACE) THEN
      CYCLE
    ENDIF

    CALL pack_all(chunk%tiles(t)%field%x_min,                    &
                  chunk%tiles(t)%field%x_max,                    &
                  chunk%tiles(t)%field%y_min,                    &
                  chunk%tiles(t)%field%y_max,                    &
                  chunk%halo_exchange_depth,                     &
                  chunk%tiles(t)%tile_neighbours,     &
                  chunk%tiles(t)%field%density,        &
                  chunk%tiles(t)%field%energy0,        &
                  chunk%tiles(t)%field%energy1,        &
                  chunk%tiles(t)%field%u,              &
                  chunk%tiles(t)%field%vector_p,       &
                  chunk%tiles(t)%field%vector_sd,      &
                  chunk%tiles(t)%field%vector_rtemp,      &
                  chunk%tiles(t)%field%vector_z,      &
                  chunk%tiles(t)%field%vector_kx,     &
                  chunk%tiles(t)%field%vector_ky,     &
                  chunk%tiles(t)%field%vector_di,     &
                  fields, &
                  depth, &
                  face, &
                  packing, &
                  mpi_buffer,                &
                  offsets, &
                  tile_offset)
  ENDDO
!$OMP END DO NOWAIT
!$OMP END PARALLEL
```
[...]

PR #692, opened 2019-03-25
Fix issue with capturing procedure pointers to OpenMP parallel regions. Fixes TeaLeaf_ref building, see #461

pawosm-arm:
This commit provides additionally a test case which signals the problem
and verifies the solution.

### DIFF
--- a/test/mp_correct/subcallx.f90
+++ b/test/mp_correct/subcallx.f90
@@ -1,0 +1,46 @@
+! RUN: %flang -O0 -fopenmp -S -emit-llvm %s -o - 2>&1 | FileCheck %s
+
+program subcallx
+
+  implicit none
+
+  call sub2
+
+contains
+
+subroutine sub1(x)
+
+  implicit none
+
+  integer :: x
+
+  print *, "Hello from sub1", x
+
+end subroutine sub1
+
+! CHECK: define internal void @subcallx_sub2
+subroutine sub2
+
+  implicit none
+
+  interface
+    subroutine simple_sub(x)
+      implicit none
+      integer :: x
+    end subroutine
+  end interface
+
+  procedure(simple_sub), pointer :: pptr => null() ! CHECK: %"pptr$p_{{[0-9]+}}" = alloca void
+
+  pptr => sub1
+
+! CHECK: @subcallx__{{.+}}_ to i64*
+! CHECK: @__kmpc_fork_call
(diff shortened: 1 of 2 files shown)


vjayathirtha-nv:
@pawosm-arm I tried your patch, it worked with and without an argument. Can you tell me what exactly you see when the subroutine does not take an argument?

pawosm-arm:
I see nothing, I guess something else was fixed too. Changed my comment.

pawosm-arm:
@vjayathirtha-nv are you sure the no-argument code does what it was expected to do? Although there's no runtime segfault anymore, what I see in IR isn't what I expected to see (though I may be wrong).

vjayathirtha-nv:
@pawosm-arm The check inside your lit test
`CHECK-NOT: %"pptr${{.+}}_{{[0-9]+}}" = alloca`

is catching 
`%pptr$sd_305 = alloca [10 x i64], align 8`

But as I understand it was actually intended for this.
`%pptr$p_306 = alloca void (i64*)*, align 8`

The regex should probably be 
`"pptr$p_{{[0-9]+}}"`

vjayathirtha-nv:
My earlier comment was incorrect. There should be no mention of `pptr` at all in the outlined routine.

I actually see `%pptr$sd_305 = alloca [10 x i64], align 8` in both tests (with and without parameters). I believe this is a tangential bug that needs a separate fix. 

Do you not see the same?

pawosm-arm:
I agree this specific case requires separate fix (and I guess separate bug report), yet since original (TeaLeaf_ref) problem is fixed and we do not have any real workload that could face this, it may be hard to justify for me opening it from our side.

vjayathirtha-nv:
That makes sense. About the test itself though, I see `%pptr$sd_305 = alloca [10 x i64]` being generated for the lit test included in this PR. Do you?

vjayathirtha-nv:
@pawosm-arm ["Fixes for OpenMP parallel regions" commit](https://github.com/flang-compiler/flang/commit/e74bd56b48fe7278024227fed0a38ef5f8d39214) includes the original fix from this PR and also fixes the tangential issue discussed above. Could you fix this PR to only have the lit test or introduce a new PR if you like?

pawosm-arm:
I'd rather open new PR with a commit that introduces new test and keep this one (closed) intact for reference. Will do in the (UK) morning.

pawosm-arm:
See PR #710 for the extracted test case.

PR #693, opened 2019-03-26
runtime: ARM64 is a 64-bit architecture

pawosm-arm:
There are places in the runtime code where AArch64 should be treated the same way as x86_64.


SouraVX (approved):

bryanpkc (approved):
I tested this patch on our downstream build. LGTM.

xoviat (approved):

Issue #688, opened 2019-03-13
[ICE][OpenMP] taskloop in subroutine causes ICE in flang2


pawosm-arm:
Following program fails to compile:
```
program t1
    integer :: var1

    var1=4

    !$OMP PARALLEL shared(var1)
    !$OMP MASTER
    call sub(var1)
    !$OMP END MASTER
    !$OMP END PARALLEL

    contains

    subroutine sub(arg1)

    integer :: arg1, i
    real :: var4, var5(10)

    var5 = 1.0
    !$OMP TASKLOOP private(i,var4) shared(arg1,var5)
    do i = 1, arg1
        var4 = 200000.0
        var5(i) = var5(i) + var4
    end do
    !$OMP END TASKLOOP

    write(*,*) "result =", var5

    end subroutine sub
end program
```
Error message:
```
F90-F-0000-Internal compiler error. print_token(): missing token       0
```
Moving the code out of subroutine makes it being able to compile:
```
program t2
    integer :: var1,var5(10),var4

    var1=4

    !$OMP PARALLEL shared(var1)
    !$OMP MASTER

    var5 = 1.0
    !$OMP TASKLOOP private(i,var4) shared(var1,var5)
    do i = 1, var1
        var4 = 200000.0
        var5(i) = var5(i) + var4
    end do
    !$OMP END TASKLOOP

    write(*,*) "result =", var5

    !$OMP END MASTER
    !$OMP END PARALLEL

end program
```

pawosm-arm:
Hi Kiran,

This patch was committed only on your fork of this repo, IMHO this ticket should not be closed...

kiranchandramohan:
I was testing out the github actions workflow for setting up a CI. Merged the fix for this issue in my fork to test triggering the CI which seems to have closed this issue. Reopening as suggested by @pawosm-arm.

PR #701, opened 2019-04-03
Fix for ICE in taskloop

kiranchandramohan:
Add missing case statement for end of Taskloop. Absence of this
case statement leads to some variables being marked as private
(instead of local). This prevents the allocation of these modules
on the stack and their further use in load or store statements.
This patch fixes https://github.com/flang-compiler/flang/issues/688.

### DIFF
--- a/test/mp_correct/inc/taskloop_ice.mk
+++ b/test/mp_correct/inc/taskloop_ice.mk
@@ -1,0 +1,18 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+build: taskloop_ice.$(OBJX)
+
+run:
+	@echo ------------ executing test $@
+	-$(RUN2) ./taskloop_ice.$(EXESUFFIX) $(LOG)
+
+verify: ;
+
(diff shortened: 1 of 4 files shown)


test/mp_correct/inc/taskloop_ice.mk
```diff
@@ -0,0 +1,29 @@
+#
+# Copyright (c) 2019, Arm Ltd.  All rights reserved.
```
SouraVX, line 2: Do you mind removing them ? This will make tests less noisy and uniform.
Perhaps, if time permits and everyone agrees we should remove these from every test file?
@gklimowicz any comments on this ?

kiranchandramohan: Sure, I will remove.

gklimowicz (approved):
I've run this change on OpenPOWER for LLVM 9 and 10. Looks good to me.

Let's just remove the unnecessary copyright notices before merging, as Kiran mentioned he will do.

bryanpkc (approved):
LGTM.

pawosm-arm (approved):

kiranchandramohan:
ping @SouraVX. I have made the changes you requested.

SouraVX:
Thanks @kiranchandramohan! we are already in process of validating this PR. Initial testing seems good, running more internal tests, will get back soon.

shivaramaarao (approved):

SouraVX (approved):
LGTM, Thanks!

PR #709, opened 2019-04-30
Pull 2019-04-30T16-05 Recent NVIDIA Changes

gklimowicz:
Simpler pull request of recent changes than #708.

### DIFF
--- a/runtime/flangrti/trace.c
+++ b/runtime/flangrti/trace.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2006-2017, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 2006-2019, NVIDIA CORPORATION.  All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


runtime/flangrti/trace.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2006-2017, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 2006-2019, NVIDIA CORPORATION.  All rights reserved.
```
sscalpone: No change?

tools/flang2/flang2exe/ll_write.cpp
```diff
@@ -400,6 +400,25 @@ ll_write_instruction(FILE *out, LL_Instruction *inst, LL_Module *module, int no_
...
+      fprintf(out, "(");
+      for(i=3; i<noperands; i++) {
+        fprintf(out, "%s %s", inst->operands[i]->type_struct->str, inst->operands[i]->data);
+        if(i<(noperands-1))
+          fprintf(out, ",");
+      }
+      fprintf(out, ")");
+    }
+  }
+  break;
```
schweitzpgi: This should not be needed in Flang.

runtime/flangrti/trace.c
gklimowicz: This appears to be an artifact of merging multiple Perforce changes into a single git commit. It looks like some of the Perforce changelists inserted changes and then backed them out, and the only change remaining was the copyright update.

I will try to keep these from showing up in Flang in the future.

tools/flang2/flang2exe/ll_write.cpp
gklimowicz: Thanks, Eric. I don't necessarily have a good way of knowing what changes to the bridge are needed or unneeded in Flang. Let's talk about how I can identify that so we can limit the changes to Flang.

PR #716, opened 2019-05-07
Make the NCAR Kernels tests into longtests

RichBarton-Arm:
Many of these tests take tens of seconds and a few take minutes. Mark the
whole subdirectory as longtests so they can be disabled for make check-all
in development. We need to enable long-tests in CI.

Also use more modern lit cmdline processing utils in lit.cfg.

Signed-off-by: Richard Barton <richard.barton@arm.com>

### DIFF
--- a/test/lit.cfg
+++ b/test/lit.cfg
@@ -344,8 +344,7 @@ flags = flags.replace(',', ' ')
 config.substitutions.append(('%flags', '\"' + flags + '\"'))
 
 # Option to keep files generated while running tests.
-keep_file = lit_config.params.get('keep', None)
-if keep_file != None:
+if lit.util.pythonize_bool(lit_config.params.get('keep', None)):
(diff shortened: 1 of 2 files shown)


gklimowicz:
@RichBarton-Arm When applying this change, I now get the following results when running "make check-flang". Is the "Unsupported tests" line what you're expecting?

```-- Testing: 1258 tests, 56 threads --
Testing: 0 .. 10.. 20.. 30.. 40.. 50.. 60.. 70.. 80.. 90.
Testing Time: 44.16s
  Expected Passes    : 1225
  Unsupported Tests  : 33
Built target check-flang```

RichBarton-Arm:
Hi Gary

Yes, the effect of the longtests patch is to make the tests in the NCAR_kernels directory UNSUPPORTED. I have not checked that 33 is the exact right number, but you should expect to see an uptick in this number with the patch.

PR #740, opened 2019-06-22
Pull 2019-06-21T21-07 Recent NVIDIA Changes

gklimowicz:


lib/scutil/pgnewfil.c
```diff
@@ -79,7 +79,12 @@ addn(char *p, unsigned long val, int n)
   return p;
 } /* addn */
 
-static long pgrand = 0;
+#ifdef HOST_WIN
+#define LONG long long
```
sscalpone: Note to future developer -- doesn't flang already have a define for 64-bit int somewhere?

PR #751, opened 2019-07-12
Fix for len intrinsic returning int*8 in some cases

kiranchandramohan:
Flang's implementation of len intrinsic returns the length attribute.
After large array support was introduced, this attribute is stored in
an integer of kind=8. The standard stipulates that the integer
returned should be of the default (kind=4). Fix is to convert to the
right kind.


gklimowicz (approved):

test/f90_correct/inc/fix_len_returning_i8.mk
```diff
@@ -0,0 +1,36 @@
+#
+# Copyright (c) 2019, Arm Ltd.  All rights reserved.
```
kiranchandramohan, line 2: Change/Remove copyright before submission.

shivaramaarao (approved):
LGTM

test/f90_correct/len_intrinsic.f90
```diff
@@ -0,0 +1,47 @@
+!** Copyright (c) 2019, Arm Ltd.  All rights reserved.
```
bryanpkc, line 1: This file should be moved to test/llvm_ir_correct/. Use the same license notice as in other files.

test/f90_correct/len_intrinsic.f90
```diff
@@ -0,0 +1,47 @@
...
+! RUN: %flang -S -emit-llvm %s -o - | FileCheck %s -check-prefix=NO_CONVERT
+! RUN: %flang -fdefault-integer-8 -S -emit-llvm %s -o - | FileCheck %s -check-prefix=CONVERT_I8
+! RUN: %flang -fdefault-integer-8 -S -emit-llvm %s -o - | FileCheck %s -check-prefix=NO_CONVERT_I8
+subroutine s1(x)
+  character(len=x) :: s1_str
+  integer :: s1_test_len
+  s1_test_len = len(s1_str)
+  print *, s1_test_len
+end subroutine
+! //CONVERT: define void @s1_
```
bryanpkc, line 27: The `//` should be deleted. The correct FileCheck comment prefix should be `! CONVERT:`.

kiranchandramohan: done,

test/f90_correct/len_intrinsic.f90
kiranchandramohan: done.

test/f90_correct/inc/fix_len_returning_i8.mk
kiranchandramohan: done.

bryanpkc (approved):
Thanks, LGTM now.

PR #752, opened 2019-07-12
Fix for ICE in atomic instruction generation

kiranchandramohan:
Using non-standard types (logical or int with kind=1 and logical
with kind=8) in code with atomic instructions were leading to ICEs due
to a switch statement not handling these types. Adding these types to
the switch statement fixes the issue.


kiranchandramohan:
Superseded by PR 951

Issue #745, opened 2019-07-04
Array constructor use results in internal compiler error. interf:new_symbol_and_link, symbol not found


peterwaller-arm:
Reproducer:

```
$ flang -c bad_module.f90
$ flang use_module.f90
F90-F-0000-Internal compiler error. interf:new_symbol_and_link, symbol not found     901  (use_module.f90: 5)
```

Note that this can be turned into a working example by substituting `INTEGER*4` for the type of `i`.

#### `bad_module.f90`

```f90
MODULE bad_module
IMPLICIT NONE

INTEGER*8, PRIVATE :: i

INTEGER*8, PARAMETER :: some_array(2) = (/ (i, i = 901, 902) /)

END MODULE bad_module
```

#### `use_module.f90`

```f90
PROGRAM use_module
USE bad_module
IMPLICIT NONE
END PROGRAM use_module

```

---

I have made a bit of progress investigating this.

In this function during the array constructor `const_eval` was returning a literal `901` instead of a symbol table for an `TY_INT8`, which `mk_cval1` expects:

https://github.com/flang-compiler/flang/blob/d01767099530eb2e3fe2314cf1d7c826877c88fc/tools/flang1/flang1exe/semutil2.c#L6470-L6471

Inside const_eval, the value of the do variable index is returned `p->currval`:

https://github.com/flang-compiler/flang/blob/d01767099530eb2e3fe2314cf1d7c826877c88fc/tools/flang1/flang1exe/semutil2.c#L6180

The issue, I think, is that `p->currval` represents the current value and not a symbol table - I infer this by seeing that currval is incremented by step, so it can't represent a symbol table entry for a TY_INT8:

https://github.com/flang-compiler/flang/blob/d01767099530eb2e3fe2314cf1d7c826877c88fc/tools/flang1/flang1exe/dinit.c#L447

I conclude that `p->currval` should be fed through `mk_cval(p->currval, A_DTYPEG(ast));` so that it can be inserted into the symbol table if the ast type is INT8.

Does that sound correct?

gklimowicz:
@mleair Mark, can you take a quick look at this?

mleair:
[...]

PR #756, opened 2019-07-23
Fix ICE interf:new_symbol_and_link, symbol not found

peterwaller-arm:
Using a parameter array from a module which is constructed with implied-do
notation resulted in an ICE:

```
INTEGER*8, PARAMETER :: some_array(3) = (/ (i, i = 4241, 4243) /)
```

`const_eval()` computes the value of the index `i` in the implied do of the
array constructor. This function was returning a number literal, which was then
being fed into `mk_cval1()`.  This latter function expects a symbol table entry
when the number is a TY_INT8, and a number literal when it is a TY_INT.

This resulted in the number literal as it appeared in the source being
incorrectly treated as a symbol table index.

The fix is to use `mk_cval` which inserts the constant into the symbol table
and calls `mk_cval1` with what it expects.

This commit adds a test which sums an array which lives in another module and
checks that the sum is correct. `INTEGER` is used with `-fdefault-integer-8`,
which is how the issue was first reported.

Fixes https://github.com/flang-compiler/flang/issues/745

### DIFF
--- a/test/f90_correct/inc/array_constructor.mk
+++ b/test/f90_correct/inc/array_constructor.mk
@@ -1,0 +1,25 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+build:  $(SRC)/array_constructor.f90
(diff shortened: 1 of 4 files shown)


peterwaller-arm:
Rebased against master.

bryanpkc (approved):
LGTM.

gklimowicz (approved):
Passes my tests on OpenPOWER. Tested against both LLVM 9 and 10.

peterwaller-arm:
Force updated to new project license headers.

shivaramaarao (approved):
LGTM.

PR #759, opened 2019-07-26
In order to prevent mysterious crashes during parallel building process, build static modules in separate include-static directory

pawosm-arm:
This patch fixes mysterious build-time problem with parallel builds we were facing for way too long.

### DIFF
--- a/runtime/flang/CMakeLists.txt
+++ b/runtime/flang/CMakeLists.txt
@@ -465,6 +465,11 @@ add_flang_library(flang_static
   )
 set_property(TARGET flang_static PROPERTY OUTPUT_NAME flang)
 
+set_target_properties(flang_static
+  PROPERTIES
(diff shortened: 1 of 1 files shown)


sscalpone:
Way too long.  No kidding.  We have a fix that's working it's way into open source.  Same makefile:
```
#
# Seralize the building of flang_shared and flang_static to eliminate
# conflicts with the same module files from the shared and static builds
# being created/recreated in the common directory
# ${CMAKE_Fortran_MODULE_DIRECTORY}.
#
# Note: building of each library is still parallelized.
#
add_dependencies(flang_shared flang_static)
```

gklimowicz:
@pawosm-arm This should be addressed in #792 commit b166f13

pawosm-arm:
Closing this one as the problem has been handled different way.
Thanks @gklimowicz for pointing this out.

PR #760, opened 2019-07-26
Rename getcpu() to get_rutime() for compatibility with glibc 2.29 and later

pawosm-arm:
glibc 2.29 (as e.g. in Gentoo Linux 17.0) introduces a new function, getcpu(). This collides at compile time with the getcpu() function defined in flang.

### DIFF
--- a/include/legacy-util-api.h
+++ b/include/legacy-util-api.h
@@ -88,7 +88,7 @@ char *mkperm(char *pattern, const char *oldext, const char *newext);
 #define ASMFILE		".s"		/* asm source */
 
 /* Measures user+system CPU milliseconds that elapse between calls. */
-unsigned long getcpu(void);
+unsigned long get_rutime(void);
 
 #ifdef __cplusplus
 }
--- a/lib/scutil/cpu-stopwatch.c
+++ b/lib/scutil/cpu-stopwatch.c
@@ -26,7 +26,7 @@
 #include "scutil.h"
 
 unsigned long
-getcpu(void)
+get_rutime(void)
 {
   static long ticks_per_second = -1;
   static unsigned long last = 0;
--- a/tools/flang1/flang1exe/main.c
+++ b/tools/flang1/flang1exe/main.c
@@ -156,7 +156,7 @@ int
 main(int argc, char *argv[])
 {
   int savescope, savecurrmod = 0;
-  getcpu();
+  get_rutime();
   init(argc, argv); /* initialize */
   if (gbl.fn == NULL)
     gbl.fn = gbl.src_file;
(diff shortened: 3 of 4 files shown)


include/legacy-util-api.h
```diff
@@ -88,7 +88,7 @@ char *mkperm(char *pattern, const char *oldext, const char *newext);
 #define ASMFILE		".s"		/* asm source */
 
 /* Measures user+system CPU milliseconds that elapse between calls. */
-unsigned long getcpu(void);
+unsigned long flang_getcpu(void);
```
sscalpone: I didn't like the old name.  I don't like the new name.  Given that we have the opportunity to pick a good name, how about something that is more descriptive?  And drop the flang_ prefix :)

pawosm-arm: It's not my native language so I'm not good at proposing names either. What about getMillis()? or getCPUMillis()? or getElapsed()?

d-parks:
get_rutime()?

pawosm-arm:
> get_rutime()?

Sounds too common, google finds "get_runtime()" used in places, e.g. https://ani.cebitec.uni-bielefeld.de/mllib/mllib2/html/svm__common_8c.html, but if you insist on that name, it's fine by me.

d-parks:
Suggestion only based upon the documentation for *nix getrusage() system call.  Will defer to the collective opinion.

pawosm-arm:
As there were no other proposals, the winner is `get_runtime()`.

d-parks:
Hi.  Not a problem - but I had suggested "get_rutime" as in:
GETRUSAGE(2)                                                        Linux Programmer's Manual                                                       GETRUSAGE(2)

NAME
       getrusage - get resource usage

SYNOPSIS
       #include <sys/time.h>
       #include <sys/resource.h>

       int getrusage(int who, struct rusage *usage);

DESCRIPTION
       getrusage() returns resource usage measures for who, which can be one of the following:

<SNIP>
       RUSAGE_THREAD (since Linux 2.6.26)
              Return resource usage statistics for the calling thread.

       The resource usages are returned in the structure pointed to by usage, which has the following form:

           struct rusage {
               struct timeval ru_utime; /* user CPU time used */
               struct timeval ru_stime; /* system CPU time used */

pawosm-arm:
done

janusw:
@gklimowicz  flang can not be compiled on recent distros (Ubuntu 19.04 etc) without e1e13b4. The fix has been waiting for more than a month. Can it be merged, please?

gklimowicz:
@janusw @pawosm-arm This should be addressed with pull request #805; this should be OK to close, yes?

PR #762, opened 2019-07-30
Force #undef HUGE in runtime/libpgmath/lib/x86_64/libm_amd.h

ibara:
Prevents preprocessor redefined warnings on the BSDs. macOS probably also suffers from this issue.

I've noticed this for a while now in the OpenBSD package builds; just finally getting around to doing something about it.

### DIFF
--- a/runtime/libpgmath/lib/x86_64/libm_amd.h
+++ b/runtime/libpgmath/lib/x86_64/libm_amd.h
@@ -89,6 +89,8 @@ typedef unsigned long long __UINT8_T;
 # define TLOSS          5
 # define PLOSS          6
 
+/* The BSDs (and probably macOS) define HUGE in math.h.  */
+# undef HUGE
 # define HUGE           3.40282347e+38F
 
 #endif


runtime/libpgmath/lib/x86_64/libm_amd.h
```diff
@@ -89,6 +89,8 @@ typedef unsigned long long __UINT8_T;
 # define TLOSS          5
 # define PLOSS          6
 
+/* The BSDs (and probably macOS) define HUGE in math.h.  */
+# undef HUGE
 # define HUGE           3.40282347e+38F
```
ararslan: Would it make more sense to use the system-provided `HUGE` value and fall back to a custom definition? That is, something like
```c
#ifndef HUGE
#define HUGE 3.40282347e+38F
#endif
```

runtime/libpgmath/lib/x86_64/libm_amd.h
```diff
@@ -89,6 +89,8 @@ typedef unsigned long long __UINT8_T;
 # define TLOSS          5
 # define PLOSS          6
 
+/* The BSDs (and probably macOS) define HUGE in math.h.  */
```
ararslan: Also I've confirmed that it is the case that `HUGE` is defined in `math.h` on macOS.

d-parks:
The cpp object macro HUGE does not appear used/referenced in the libpgmath source.  I'll look at deleting the macro altogether.  Thank you for pointing this out.

ibara:
Thank you!

gklimowicz:
This should be resolved with @d-parks last changes. Please let me know if you find otherwise.

PR #767, opened 2019-08-01
Pull 2019-08-01T10-37 Recent NVIDIA Changes

gklimowicz:


tools/flang2/flang2exe/lldebug.cpp
```diff
@@ -518,6 +519,11 @@ lldbg_create_global_variable_mdnode(LL_DebugInfo *db, LL_MDRef context,
   llmd_add_i32(mdb, make_dwtag(db, DW_TAG_variable));
   llmd_add_i32(mdb, 0);
   llmd_add_md(mdb, context);
+#ifdef FLANG_LLVM_EXTENSIONS
```
sscalpone: Why isn't this always on?

gklimowicz: A while back, we had a build issue because some extensions weren't in our LLVM for some versions of LLVM. There was a lag between when they were done for LLVM 6, 7 and 8. We added this guard to prevent those build errors.

schweitzpgi: There was a request a long, long time ago that flang be buildable with "off the shelf" LLVM rather than LLVM with any modifications.

We could change the default to assume that flang is being built with flang's LLVM and any user's that still want to use a generic LLVM source tree have to add special build flags.

PR #768, opened 2019-08-01
Pull 2019-08-01T12-56 Recent NVIDIA Changes

gklimowicz:
Add support for libpgmath Windows and OSX builds with clang.

Majority of CmakeLists.txt and lit*.cfg* changes attributable
to PR#719: https://github.com/flang-compiler/flang/pull/719

1) Differentiate between CMAKE_SYSTEM_PROCESSOR and
   LIBPGMATH_SYSTEM_PROCESSOR throughout libpgmath configuration
   scripts.

2) Windows version of CMAKE_SYSTEM_PROCESSOR is defined as "AMD64".
   If building for Windows, set LIBPGMATH_SYSTEM_PROCESSOR to "x86_64".
   This eliminates any need for most conditional logic that applies
   to both Linux and Windows Intel 64 source/object files.

3) Cleanup various L1 and L2 definitions (C preprocessor -D<def>)
   eliminating duplicates.

4) Special case building of DLL in rte/libpgmath/lib/CMakeLists.txt
   for the MSYS environment.

5) To support Window's unique complex variable declarations of
   _Fcomplex and _Dcomplex, globally change declarations of the form:
	float _Complex f
	double _Complex d
   to:
	float_complex_t
	double_complex_t

  This also consolidates the C++ declarations. Specifically, C++
  typedefs complex_float and complex_double have been eliminated along
  with any CPP function macro that used those typedefs. Basically
  conditional __cplusplus logic in rte/pgm/port/src_new/mthdecls.h has
  been eliminated.

5) Introduce two forms of complex variable assignment:
   a) float/double static/constants use the CPP function macro
      PGMATH_CMPLX_CONST
   b) Complex variables composed of real and imaginary variables
      (not constants) use:
      float_complex_t pgmath_cmplxf(float, float);
      double_complex_t pgmath_cmplx(double, double);

Additional libpgmath mods to build rte/libpgmath.

1) bin/build-library
   add logic to build libpgmath.
   Note: logic disabled.

2) bin/win64/tccmath_cmake
   Windows+cmake specific version of tccmath.

3) rte/libc/hammer/include-win64/complex.h
   Add missing complex intrinsics, i.e., cacos, casin, ...

4) rte/libpgmath/*/makefile
   Makefiles consistent in structure/form with legacy library builds
   that configure and build using cmake.

4) rte/libpgmath/*cleanup
   Cleanup dependency files.

5) rte/pgm/port/src_new/mthdecls.h
   Eliminate CPP function macros for calling host versions of various
   math functions. These macros (example: CACOSF) were needed because
   Windows at some point in the past only had entry points for the
   double precision version of the intrinsic. Windows C runtime library
   now has both the single and double precision verions of the
   needed functions.

6) rte/pgc/port/src/
   rte/pgm/port/src_new/
   With #5 above, change (example) from:

	float_complex_t __mth_i_ccosh(float_complex_t carg)
	{
		CMPLXARGS_C;
		float_complex_t f;
		f = CCOSHF(carg);
		CRETURN_C(f);
	}
  to:
	float_complex_t __mth_i_ccosh(float_complex_t carg)
	{
		CMPLXARGS_C;
		float_complex_t f;
		f = ccoshf(carg);
		CRETURN_C(f);
	}

7) rte/libpgmath/lib/CMakeLists.txt
   *Temporary* Windows: only build static version of libpgmath.


runtime/libpgmath/CMakeLists.txt
```diff
@@ -117,3 +122,11 @@ include(LibmathUtils)
 # Setup Source Code And Tests
 add_subdirectory(lib)
 add_subdirectory(test)
+
+if(0)
```
sscalpone: Why an if(0)?

gklimowicz: The developer said it was left in for debug. I can have it removed.

sscalpone: No need to do anything special.

PR #770, opened 2019-08-01
Pull 2019-08-01T15-45 Recent NVIDIA Changes

gklimowicz:
Add support for libpgmath Windows and OSX builds with clang.

Majority of CmakeLists.txt and lit*.cfg* changes attributable
to PR#719: https://github.com/flang-compiler/flang/pull/719

1) Differentiate between CMAKE_SYSTEM_PROCESSOR and
   LIBPGMATH_SYSTEM_PROCESSOR throughout libpgmath configuration
   scripts.

2) Windows version of CMAKE_SYSTEM_PROCESSOR is defined as "AMD64".
   If building for Windows, set LIBPGMATH_SYSTEM_PROCESSOR to "x86_64".
   This eliminates any need for most conditional logic that applies
   to both Linux and Windows Intel 64 source/object files.

3) Cleanup various L1 and L2 definitions (C preprocessor -D<def>)
   eliminating duplicates.

4) Special case building of DLL in runtime/libpgmath/lib/CMakeLists.txt
   for the MSYS environment.

5) To support Window's unique complex variable declarations of
   _Fcomplex and _Dcomplex, globally change declarations of the form:
	float _Complex f
	double _Complex d
   to:
	float_complex_t
	double_complex_t

6) Introduce two forms of complex variable assignment:
   a) float/double static/constants use the CPP function macro
      PGMATH_CMPLX_CONST
   b) Complex variables composed of real and imaginary variables
      (not constants) use:
      float_complex_t pgmath_cmplxf(float, float);
      double_complex_t pgmath_cmplx(double, double);

7) runtime/libpgmath/lib/CMakeLists.txt
   Windows: only build static version of libpgmath.


runtime/libpgmath/CMakeLists.txt
```diff
@@ -117,3 +122,11 @@ include(LibmathUtils)
 # Setup Source Code And Tests
 add_subdirectory(lib)
 add_subdirectory(test)
+
+if(0)
```
cbezault: Is this left-over debugging code?

gklimowicz: Yes, the developer here wanted to leave it in for the time being.

PR #773, opened 2019-08-08
Support for isnan GNU extension: Generate isnanf() & isnan() libm calls

mvinay:


tools/flang1/utils/symtab/symini_ftn.n
```diff
@@ -1297,6 +1297,8 @@ CM Fortran dotproduct intrinsic.
 .AT inquiry
 .PD PROCESSORS_SHAPE - -
 .AT inquiry
+.PD ISNAN - -
+.AT elemental x
```
kiranchandramohan: The following code shows link errors. Is there no support for the elemental version?
program isnan01
  real :: i(5)
  i = 0.0
  print *, isnan(i)
end program

tools/flang1/flang1exe/semfunc.c
```diff
@@ -5786,6 +5786,16 @@ ref_pd(SST *stktop, ITEM *list)
       XFR_ARGAST(1);
     }
     break;
+  case PD_isnan:
```
kiranchandramohan: I am assuming that support for isnan during initialization is deliberately left out. right?
real :: x = isnan(0.0)
F90-S-0155-Intrinsic not supported in initialization: isnan

tools/flang1/utils/symtab/symini_ftn.n
mvinay: This is currently not supported.

tools/flang1/flang1exe/semfunc.c
mvinay: Yes, this is not supported.

tools/flang1/utils/symtab/symini_ftn.n
kiranchandramohan: Can you remove the elemental description if elemental is not supported?

test/gnu_extensions/inc/isnan01.mk
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
kiranchandramohan: Can you remove the Nvidia copyright for new files you are adding?

tools/flang1/utils/symtab/symini_ftn.n
mvinay: I have added the support for elemental version.

tools/flang1/flang1exe/func.c
```diff
@@ -4416,6 +4416,8 @@ mk_result_sptr(int func_ast, int func_args, int *subscr, int elem_dty, int lhs,
     temp_sptr = mk_shape_sptr(shape, subscr, elem_dty);
     *retval = mk_id(temp_sptr);
     break;
+  case I_ISNAN:
+     break;
```
kiranchandramohan: Is this case necessary? If this case falls in the execution path then some uninitialized value will be returned. Else if this case is necessary can you add a test which hits this path?

tools/flang1/flang1exe/semfunc.c
```diff
@@ -5817,6 +5817,21 @@ ref_pd(SST *stktop, ITEM *list)
...
+    if (count != 1 || get_kwd_args(list, count, KWDARGSTR(pdsym)))
+      goto bad_args;
```
kiranchandramohan: bad_args (here and also below) does not seem to be erroring out if provided incorrect parameters. This usually results in a linking error. 
Using E74_CNT, E74_ARG might be more appropriate here. Also add a test which checks these.

tools/flang1/utils/symtab/symini.cpp
```diff
@@ -1541,6 +1541,7 @@ const char *SyminiFE90::init_names0[] = {
     "any",
     "compl",
     "count",
+    "isnan",
```
kiranchandramohan: Do you know whether isnan should be added to all these init_names arrays or only one of them is necessary?

pawosm-arm: I'm seeing it added to other init_names too.

pawosm-arm (changes requested):
@vinaym-amd we really need this PR to be merged. Downstream we were using it and found no visible issues with it, so the only remaining problem is with the licensing. Can you change licensing of the test cases to LLVM-compliant? I'd approve and merge it after that.

pawosm-arm:
Overridden by PR #1269 so closing this one

PR #775, opened 2019-08-11
Introduce a LIBPGMATH_SYSTEM_NAME variable.

ibara:
This is a genericized CMAKE_SYSTEM_NAME, modelled after
LIBPGMATH_SYSTEM_PROCESSOR. This permits ease of maintenance of
other Unix operating systems, like OpenBSD, who can just "pretend"
to be Linux for libpgmath.


gklimowicz:
I believe this is fixed in Flang. Please reopen if you find otherwise.

PR #778, opened 2019-08-13
Fixing NCAR test problems with error tolerance lower than E-12.

CarolineConcatto:
The NCAR test use selected_real_kind(12), therefore the  tolerance tests should also compare values until E-12.
This patch adds a tolerance  value TOL according to the architecture.
 The consequence is to make the tests more robust


sscalpone:
Please consider pushing these changes directly to NCAR.  You might start with github.com/NCAR.  @gklimowicz may be able to give you a pointer to the project from where these benchmarks live.

PR #795, opened 2019-09-11
Pull 2019-09-11T13-10 Recent NVIDIA Changes

gklimowicz:


tools/flang2/flang2exe/kmpcutil.cpp
```diff
@@ -38,7 +38,7 @@
 #include "ll_structure.h"
 #include "llmputil.h"
 #include "llutil.h"
-#ifdef OMP_OFFLOAD_LLVM
+#if defined(OMP_OFFLOAD_LLVM) || defined(OMP_OFFLOAD_PGI)
```
sscalpone: Is this the same as "#if 1" ?

gklimowicz:
Maybe not for Arm's CI, based on the build failure that just happened. It looks like they don't define it in their builds. And it looks like we may have a misplaced #endif in ompaccel.cpp. I'll investigate.

RichBarton-Arm:
Hi @gklimowicz - what's the update on this? Until the code builds again, the bot is essentially useless for checking the correctness of merges.

Do I understand right from the above comments that building with -DOMP_OFFLOAD_LLVM would fix the buildbot again? If so, we can make that change fairly quickly but as Kiran and Steve say, we should change the official build instructions to also include this setting. Are you able to merge a patch for that quickly?

RichBarton-Arm:
@gklimowicz  Alternatively, if I misunderstood, and we require additional patches to revert to a passing build would it work to revert this merge to get the builder passing again then re-commit with the additional changes that you mention above included?

gklimowicz:
Im trying to work my way through the changes that have been queued up that address this. I've been running into merge conflicts if I try to pull them too far forward in the queue. You could revert this merge and see if it improves things temporarily, but I have not attempted it.

PR #801, opened 2019-09-20
Enable build when OMP_OFFLOAD_LLVM and OMP_OFFLOAD_PGI are both

ibara:
not defined, for example the official OpenBSD package.

### DIFF
--- a/tools/flang2/flang2exe/ompaccel.cpp
+++ b/tools/flang2/flang2exe/ompaccel.cpp
@@ -2412,7 +2412,6 @@ init_test()
   init_tgtutil();
 }
 
-#endif
 /* Expander - OpenMP Accelerator Model */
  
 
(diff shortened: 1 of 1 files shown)


ibara:
This is needed to keep the package building on OpenBSD (and I assume other platforms that don't yet ship libomp).

ibara:
Note that this breakage began with 74949ae.

gklimowicz:
@ibara Yes, there was a problem with this commit. I have a fix in queue that should resolve it as soon as our gatekeeper reviews and passes the latest pull request I made. (It is large and affected a lot of the OpenMP implementation; it was a big refactor of the work.) It should be the next pull request that gets merged into Flang.

ibara:
Great, thanks!

gklimowicz:
@ibara PR #800 should resolve this issue for you, Brian.

PR #816, opened 2019-10-11
LLVM9 debuginfo change

RichBarton-Arm:
LLVM9 Changes
LLVM9 moves some bitfields from the DIFlags to a subprogram-specific field called DISPFlags. Teach flang2 to emit these, whilst staying backwards compatible.

### DIFF
--- a/tools/flang2/flang2exe/lldebug.cpp
+++ b/tools/flang2/flang2exe/lldebug.cpp
@@ -75,11 +75,19 @@ inline SPTR GetParamSptr(int dpdsc, int i) {
 #define DW_TAG_vector_type 0x103
 #endif
 
-const int DIFLAG_ARTIFICIAL = 1 << 6;
-const int DIFLAG_ISMAINPGM = 1 << 21;
+static int DIFLAG_ARTIFICIAL;
+static int DIFLAG_ISMAINPGM;
 static int DIFLAG_PURE;
 static int DIFLAG_ELEMENTAL;
 static int DIFLAG_RECUSIVE;
+
+static int DISPFLAG_LOCALTOUNIT;
+static int DISPFLAG_DEFINITION;
+static int DISPFLAG_OPTIMIZED;
+static int DISPFLAG_PURE;
+static int DISPFLAG_ELEMENTAL;
+static int DISPFLAG_RECUSIVE;
+static int DISPFLAG_ISMAINPGM;
 
 typedef struct {
   LL_MDRef mdnode; /**< mdnode for block */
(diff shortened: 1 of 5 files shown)


gklimowicz:
@jiel-nv Can you look these over for us?

tools/flang2/flang2exe/lldebug.cpp
```diff
@@ -181,11 +189,20 @@ void
 InitializeDIFlags(const LL_IRFeatures *feature)
 {
 #ifdef FLANG_LLVM_EXTENSIONS
-  if (ll_feature_debug_info_ver70(feature)) {
+  DIFLAG_ARTIFICIAL = 1 << 6;
```
jiel-nv: It seems to me the DIFLAG_ARTIFICIAL = 1 << 6 should be pulled out of the FLANG_LLVM_EXTENSIONS guard.

tools/flang2/flang2exe/lldebug.cpp
```diff
@@ -447,18 +487,23 @@ lldbg_create_subprogram_mdnode(
...
+  }
   llmd_add_i32(mdb, virtuality);
   llmd_add_i32(mdb, vindex);
-  llmd_add_null(mdb);
+  llmd_add_null(mdb); // containingType field
   llmd_add_i32(mdb, flags);
-  llmd_add_i1(mdb, is_optimized);
+  if (ll_feature_debug_info_ver90(&db->module->ir))
+    llmd_add_i32(mdb, spFlags);
+  if (!ll_feature_debug_info_ver90(&db->module->ir))
```
jiel-nv: Could be an "else" branch?

tools/flang2/flang2exe/lldebug.cpp
```diff
@@ -75,12 +75,20 @@ inline SPTR GetParamSptr(int dpdsc, int i) {
...
 static int DIFLAG_ELEMENTAL;
 static int DIFLAG_RECUSIVE;
 
+static int DISPFLAG_LOCALTOUNIT;
+static int DISPFLAG_DEFINITION;
+static int DISPFLAG_OPTIMIZED;
+static int DISPFLAG_PURE;
+static int DISPFLAG_ELEMENTAL;
+static int DISPFLAG_RECUSIVE;
+static int DISPFLAG_ISMAINPGM;
```
schweitzpgi: These flags should all be `const` since they do not change in prior versions of LLVM IR. (They didn't exist or were not used if they did, so there is no backward compatibility issue.)

tools/flang2/flang2exe/lldebug.cpp
```diff
@@ -195,6 +212,29 @@ InitializeDIFlags(const LL_IRFeatures *feature)
...
+    DISPFLAG_LOCALTOUNIT = 0;
+    DISPFLAG_DEFINITION = 0;
+    DISPFLAG_OPTIMIZED = 0;
+    DISPFLAG_PURE = 0;
+    DISPFLAG_ELEMENTAL = 0;
+    DISPFLAG_RECUSIVE = 0;
+    DISPFLAG_ISMAINPGM = 0;
+  }
+}
+
```
schweitzpgi: This function isn't needed if we make the DISP flags constants.

tools/flang2/flang2exe/lldebug.cpp
schweitzpgi: An `else` would make this a tad clearer to read. (The `!` is less clear.)

tools/flang2/flang2exe/lldebug.cpp
```diff
@@ -2008,23 +2064,27 @@ lldbg_emit_outlined_subprogram(LL_DebugInfo *db, int sptr, int findex,
   mips_linkage_name = func_name;
   is_def = DEFDG(sptr);
   is_def |= (STYPEG(sptr) == ST_ENTRY);
+  if (ll_feature_has_diextensions(&db->module->ir))
+    flags = set_disubprogram_flags(sptr);
```
schweitzpgi: I think it would be clearer if `set_disubprogram_flags` was customized to omit the DIFLAGs that are no longer in that flags word rather than introducing more pseudo-constants and relying on them being reset to 0.

tools/flang2/flang2exe/lldebug.cpp
RichBarton-Arm: Ack

RichBarton-Arm:
Thanks @jiel-nv and @schweitzpgi for the reviews. 

I have moved all the flag values back to being const variables. The ones that changed value between LLVM versions are now different variables. The functions that initialse them are gone and instead the function that sets the flag value uses the right values for the LLVM version being generated for.

I also fixed the small comment on else vs !

I have uploaded a third commit which should address them all, but I am not sure if you would prefer me to squash the commits together before the final pull. Let me know if you want me to do this.

sscalpone:
Yes, please squash.

RichBarton-Arm:
Seems like this got merged in in other commits, so closing.

Issue #844, opened 2019-12-11
Internal Compiler Error when compiling arithmetic if code with -i8 flag


kiranchandramohan:
Code with arithmetic if when compiled using the -i8 flag shows an internal compiler error.
F90-F-0000-Internal compiler error. addili: illegal IL_TYPE(opc)       0
```
IF (J) 85,85,105
85 I = J
105 I = J
END 
```

This is a regression introduced (https://github.com/flang-compiler/flang/commit/a4012a93a3d9efebe5de1289d8900decf515fb2b) when adding half precision support. The issue happens because the aif array size was increased from 4 to 5 and the index for integer*8 increased from 3 to 4 but the entry for half precision was not added. Hence the index for integer*8 pointing to an incorrect location.
Relevant file: tools/flang2/flang2exe/exp_ftn.cpp

Is the correct fix to 
1) Add the entry for half precision
or
2) Revert to the original array size and index for integer*8.

PR #845, opened 2019-12-12
Fix ICE in ArithmeticIF for INT*8 type

kiranchandramohan:
When half precision reals were introduced the table (aif) used in the
ILI codegen for ArithmeticIF was increased in size and the offset for
INT*8 increased by one. But the entry for half precision was not
inserted, this causes an ICE for INT*8 type. The fix is to introduce
the entry for the half precision type before the INT*8 type.

Fix for https://github.com/flang-compiler/flang/issues/844

### DIFF
--- a/test/f90_correct/inc/arithmetic_if.mk
+++ b/test/f90_correct/inc/arithmetic_if.mk
@@ -1,0 +1,26 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+$(TEST): run
+	
+
+build:  $(SRC)/$(TEST).f90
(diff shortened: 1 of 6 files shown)


kiranchandramohan:
Thanks @gklimowicz for having a look. Also, appreciate the quick response.

kiranchandramohan:
@shivaramaarao FYI, I believe 859 (is it from AMD?) also addresses the same issue. But I feel this patch's approach of adding the missing entry is probably better than deleting the entry. Let me know in case you think differently.
https://github.com/flang-compiler/flang/pull/859

shivaramaarao:
> @shivaramaarao FYI, I believe 859 (is it from AMD?) also addresses the same issue. But I feel this patch's approach of adding the missing entry is probably better than deleting the entry. Let me know in case you think differently.
> #859

We checked this patch and it works fine with our testing. LGTM

shivaramaarao (approved):
LGTM

kiranchandramohan:
@gklimowicz Since you have approved it previously is it safe to assume that you have run the tests?

gklimowicz:
Yes, I probably ran them in December, but let me make sure and run them again specifically on Power.

PR #859, opened 2020-01-23
Fixed internal compiler error. addili: illegal IL_TYPE(opc)

nikhil-mk:
Accessing the correct index

Change-Id: I8a8103f71266366bfb32c658df395ce61e3065f8

### DIFF
--- a/tools/flang2/flang2exe/exp_ftn.cpp
+++ b/tools/flang2/flang2exe/exp_ftn.cpp
@@ -3442,7 +3442,7 @@ exp_bran(ILM_OP opc, ILM *ilmp, int curilm)
     break;
 
   case IM_KAIF: /* integer*8 arithmetic IF */
-    type = 4;
+    type = 3;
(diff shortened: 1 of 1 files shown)


kiranchandramohan:
Is this a duplicate of https://github.com/flang-compiler/flang/issues/844 and the fix in https://github.com/flang-compiler/flang/pull/845 ?

nikhil-mk:
> Is this a duplicate of #844 and the fix in #845 ?

No.

kiranchandramohan:
@nikhil-mk this was the first fix I thought of when trying to fix the issue https://github.com/flang-compiler/flang/issues/844. But then I realised that the index (=4) is correct if the entry for half-precision is there at index (=3) is there in the aif array in exp_bran in exp_ftn.cpp. The fix in https://github.com/flang-compiler/flang/pull/845 adds that entry.

mleair:
Looks good to me!

PR #863, opened 2020-02-06
Support for -gdwarf-5 in flang2

alokkrsharma:
Summary:
  flang2 doesnt produce dwarf version 5 for -gdwarf-5.

  Currently when -gdwarf-5 is passed to flang (driver), that doesnt
pass any corresponding (-x 120) option as it passes in case of
-gdwarf-2/3. flang2 binary picks -gdwarf-4 by default.
   # flang -gdwarf-5 test.f90
   # llvm-dwarfdump a.out | grep version
0x00000000: Compile Unit: length = 0x0000008e version = 0x0004

  Now flang2 will identify gdwarf-4 and gdwarf-5 by -x 120 0x1000000
and -x 120 0x2000000, flang(driver) need to pass these option to flang

  flang-driver changes will be pushed separately. After that -gdwarf-5
will work from glang driver.

Testing:
  - check-llvm
  - check-debuginfo

Change-Id: Ibb70438f0a024bb7a6425d8b022c9353d138fa98

### DIFF
--- a/tools/flang2/flang2exe/ll_structure.cpp
+++ b/tools/flang2/flang2exe/ll_structure.cpp
@@ -394,10 +394,12 @@ compute_ir_feature_vector(LLVMModuleRef module, enum LL_IRVersion vers)
     module->ir.dwarf_version = LL_DWARF_Version_2;
   } else if (XBIT(120, 0x4000)) {
     module->ir.dwarf_version = LL_DWARF_Version_3;
+  } else if (XBIT(120, 0x1000000)) {
+    module->ir.dwarf_version = LL_DWARF_Version_4;
+  } else if (XBIT(120, 0x2000000)) {
+    module->ir.dwarf_version = LL_DWARF_Version_5;
   } else if (true) { // FIXME - need a new bit
     module->ir.dwarf_version = LL_DWARF_Version_4;
-  } else {
-    module->ir.dwarf_version = LL_DWARF_Version_5;
   }
 
   if (ll_feature_versioned_dw_tag(&module->ir)) {
@@ -434,6 +436,8 @@ ll_feature_dwarf_version(const LL_IRFeatures *feature)
     return 3;
   case LL_DWARF_Version_4:
     return 4;
+  case LL_DWARF_Version_5:
+    return 5;
   }
 }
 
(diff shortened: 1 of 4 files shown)


tools/flang2/flang2exe/ll_structure.cpp
```diff
@@ -394,10 +394,12 @@ compute_ir_feature_vector(LLVMModuleRef module, enum LL_IRVersion vers)
     module->ir.dwarf_version = LL_DWARF_Version_2;
   } else if (XBIT(120, 0x4000)) {
     module->ir.dwarf_version = LL_DWARF_Version_3;
-  } else if (true) { // FIXME - need a new bit
+  } else if (XBIT(120, 0x1000000)) {
     module->ir.dwarf_version = LL_DWARF_Version_4;
-  } else {
+  } else if (XBIT(120, 0x2000000)) {
     module->ir.dwarf_version = LL_DWARF_Version_5;
+  } else if (true) { // FIXME - need a new bit
```
schweitzpgi: Can you make this a simple `else` clause? And add a comment that DWARF 4 is the default? (This will fix the FIXME. Yay.)

schweitzpgi:
@jiel-nv Should we have some sort of test for this, to make sure DWARF-5 is really being output?

tools/flang2/flang2exe/ll_structure.cpp
alokkrsharma: Thanks a lot for your comment. I shall be updating with your suggestion.

jiel-nv:
> @jiel-nv Should we have some sort of test for this, to make sure DWARF-5 is really being output?

Yes, I think so. I will work it out.

nextsilicon-itay-bookstein:
I noticed that the other bits controlling the DWARF version are used for instance in `tools/flang2/flang2exe/main.cpp` in the definitions of `IS_DWARF` and `IS_COFFDBG`. Should they be modified to include these two new possibilities as well?

alokkrsharma:
> 
> 
> I noticed that the other bits controlling the DWARF version are used for instance in `tools/flang2/flang2exe/main.cpp` in the definitions of `IS_DWARF` and `IS_COFFDBG`. Should they be modified to include these two new possibilities as well?

Thanks for bringing this to notice. I also noticed that these defines are unused. Can you are someone else suggest whether these defines need update for newer bits ?

nextsilicon-itay-bookstein:
If it were up to me I'd delete the dead code rather than update it, but I defer to the project's maintainers in deciding this - I don't know the history/context behind that piece of dead code :)

gklimowicz (approved):
I tested this PR along with #888 and #921, and the Flang-driver PR#92 and PR#93. These pass tests on OpenPOWER. The entire collection looks good to me.

kiranchandramohan (approved):
LGTM.

bryanpkc (approved):
LGTM.

PR #886, opened 2020-04-20
Replaced deprecated function stime which is removed in glibc 2.31

arjunsuresh1987:


kiranchandramohan:
Thanks @arjunsuresh1987 for this patch. 

Can you add a pointer to the removal announcement/description?
Is this fixing a flang build failure or is the stime runtime function called in flang generated code?

wyphan:
I can confirm this also affects AOCC 2.1 on Ubuntu 20.04 LTS (which ships glibc 2.31). I emailed AMD toolchain support about this, and the latest reply describes the following workaround:
>The issue is because the library function stime is removed from GLIBC version of ubuntu 20.04 and flang is referencing the library function.
>
>One alternative is to create a C function `stime` and link this C file with `flang`.
>
>-- `stime.c` ---
>
>```
>#include <time.h>
>int stime(const time_t *t)
>{
>  struct timespec ts = {};
>  ts.tv_sec = *t;
>  return clock_settime(CLOCK_REALTIME, &ts);
>}
>```
>`$ flang test.f90 stime.c`

kiranchandramohan:
Can a test be added?

bryanpkc:
@kiranchandramohan Adding a test is probably difficult since setting the system time requires root access.

paul-mchugh:
I had run into this problem while building flang.  Thanks for the fix.

runtime/flang/stime3f.c
```diff
@@ -18,9 +18,9 @@
 int ENT3F(STIME, stime)(int *tp)
 {
   int i;
-  time_t t = *tp;
-
-  if ((i = stime(&t)))
+  struct timespec ts = {};
```
kiranchandramohan, line 21: LGTM.

I observed the following warning with gcc -pedantic.
warning: ISO C forbids empty initializer braces [-Wpedantic]
   struct timespec ts = {}

arjunsuresh1987: Thank you for the reviews. @kiranchandramohan  I have fixed the empty initializer warning now.

kiranchandramohan:
@shivaramaarao assuming this patch is from AMD. Please "Rebase and merge" after the minor issue that @gklimowicz is pointing out here.

Issue #90, opened 2017-06-07
Debug Build Type cannot be built


pawosm-arm:
I've built LLVM (release_40 branch, SHA f3d3277bb713bb8aced9a7ac2e9b05c52d2844ee) with Clang from flang-compiler (flang_release_40 branch, SHA 1210030915d1e1441d62eea54976f4ced7f6ad88). For this build I've selected -DCMAKE_BUILD_TYPE=Debug in CMake. Then I tried to build Flang (master branch), also with -DCMAKE_BUILD_TYPE=Debug set in CMake. It failed with following error:
```
[ 36%] Building Fortran object runtime/flang/CMakeFiles/flang_static.dir/ieee_arithmetic.F95.o
fragment covers entire variable
  tail call void @llvm.dbg.value(metadata i32 %1, i64 0, metadata !1064, metadata !1065), !dbg !1059
!1064 = !DILocalVariable(name: "ix", scope: !1060, file: !3, line: 630, type: !28)
!1065 = !DIExpression(DW_OP_LLVM_fragment, 0, 32)
.
.
.
fragment covers entire variable
  tail call void @llvm.dbg.value(metadata i32 %2, i64 0, metadata !1330, metadata !1064), !dbg !1326
!1330 = !DILocalVariable(name: "ix", scope: !1327, file: !3, line: 936, type: !28)
!1064 = !DIExpression(DW_OP_LLVM_fragment, 0, 32)
clang-4.0: lib/IR/Verifier.cpp:4498: virtual bool {anonymous}::VerifierLegacyPass::doFinalization(llvm::Module&): Assertion `!V->hasBrokenDebugInfo() && "Module contains invalid debug info"' failed.
```

WORKAROUND is to set following environmental variables before running `cmake`:
FFLAGS="-O0"
FCFLAGS="-O0"

UNFORTUNATELY, Flang built that way fails to compile many of the source files when started with -g on the optimization level 2 and higher, due to the same assertion as above ("Module contains invalid debug info").
Example workload that fails to build with -g -O2 using Debug Build Type Flang:
https://codesign.llnl.gov/lulesh/LULESH_serial.F90

schweitzpgi:
This bug was filed against LLVM’s SROA. See: https://bugs.llvm.org/show_bug.cgi?id=32687

The LLVM IR passes verification before SROA and fails after SROA in LLVM.

We haven’t had an opportunity to pursue the response as yet.

pawosm-arm:
[...]

Issue #89, opened 2017-06-06
Support for gfortran ABI ?


cournape:
I built a recent master of flang and tried to build a non trivial piece of fortran code, scipy. The build went well, and I could even run some of the tests, but then the test suite crashes when calling BLAS`dotu` function. I suspect this is because the default BLAS/LAPACK on my debian system is built with the gfortran ABI. Is there a way to control the ABI for emitted code by flang ?

jeffhammond:
I recommend just compiling BLAS/LAPACK with Flang.  They are open-source.  OpenBLAS and BLIS both accept a Fortran compiler argument if you want a high-performance BLAS library.

It is possible that BLIS does not yet know about the Flang ABI for DOT but that's a trivial bug to fix if it's reported (e.g. https://github.com/flame/blis/pull/637).

vchuravy:
Is there documentation on the Flange ABI? The original post mentioned scipy. Users like Julia and Python need to know the ABI in order to call fortran applications.

If the ABI is consistent with gfortran that makes our lives easier.

jeffhammond:
At least part of it is described in https://releases.llvm.org/11.0.1/tools/flang/docs/Calls.html but I don't think that answers all the questions, in particular, the one here.

PR #888, opened 2020-05-10
flang gen-exec does not show routine variables with parameter attribute

alokkrsharma:
This patch enables printing of routine scoped variables with parameter
attributes and enum values under debuggers.

- array and derived types with parameter attribute are fixed by
re-naming corresponding array constant values var$ac to var.

- scalar types with parameter attribute are fixed by generating
dbg.value intrinsics.

Example:
```
program main
  integer (kind=8) :: svar
  integer (kind=8) :: avar(5)
  integer (kind=8), parameter :: spar = 99
  integer (kind=8), parameter :: apar(5) = (/99, 98, 97, 96, 95/)
  svar = spar
  avar = apar

  print *, svar, avar, spar, apar

end program main
```

For above test program, now gdb is able to show the value of variables
with parameter attribute.
```
. . . . .
(gdb) p spar
$1 = 99
(gdb) p apar
$2 = (99, 98, 97, 96, 95)
. . . . .
```
Note: #921 should be committed first.

Testing
- make check-flang

### DIFF
--- a/test/debug_info/parameter.f90
+++ b/test/debug_info/parameter.f90
@@ -1,0 +1,17 @@
+!RUN: %flang -g -S -emit-llvm %s -o - | FileCheck %s
+
+!CHECK: call void @llvm.dbg.value(metadata i64 99, metadata [[SPAR:![0-9]+]], metadata !DIExpression())
+!CHECK: distinct !DIGlobalVariable(name: "apar"
+!CHECK: [[SPAR]] = !DILocalVariable(name: "spar"
+
+program main
+  integer (kind=8) :: svar
+  integer (kind=8) :: avar(5)
+  integer (kind=8), parameter :: spar = 99
+  integer (kind=8), parameter :: apar(5) = (/99, 98, 97, 96, 95/)
+  svar = spar
+  avar = apar
+
+  print *, svar, avar, spar, apar
+
+end program main
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -1299,8 +1299,8 @@ remove_dead_instrs(void)
 static void process_params(void) {
   unsigned smax = stb.stg_avail;
   for (SPTR sptr = get_symbol_start(); sptr < smax; ++sptr) {
-    if (STYPEG(sptr) == ST_PARAM && should_preserve_param(sptr)) {
-      DTYPE dtype = DTYPEG(sptr);
+    DTYPE dtype = DTYPEG(sptr);
+    if (STYPEG(sptr) == ST_PARAM && should_preserve_param(dtype)) {
       if (DTY(dtype) == TY_ARRAY || DTY(dtype) == TY_STRUCT) {
         /* array and derived types have 'var$ac' constant variable
          * lets use that, by renaming that to 'var'.
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -116,6 +116,7 @@ static char *fn_sig_ptr = NULL;
 static void insert_entry_label(int);
 static void insert_jump_entry_instr(int);
 static void store_return_value_for_entry(OPERAND *, int);
+static void insertLLVMDbgValue(OPERAND *, LL_MDRef, SPTR, LL_Type *);
 
 static int openacc_prefix_sptr = 0;
 static unsigned addressElementSize;
@@ -1293,6 +1294,38 @@ remove_dead_instrs(void)
 }
 
 /**
+   \brief process debug info of variables with parameter attribute.
+ */
+void
+process_params(void)
+{
+  unsigned smax = stb.stg_avail;
+  for (SPTR sptr = get_symbol_start(); sptr < smax; ++sptr) {
+    if (STYPEG(sptr) == ST_PARAM && should_preserve_param(sptr)) {
+      DTYPE dtype = DTYPEG(sptr);
+      if (DTY(dtype) == TY_ARRAY || DTY(dtype) == TY_STRUCT) {
+        /* array and derived types have 'var$ac' constant variable
+         * lets use that, by renaming that to 'var'.
+         */
+	SPTR new_sptr = (SPTR) CONVAL1G(sptr);
+	NMPTRP(new_sptr, NMPTRG(sptr));
+      } else {
+	LL_DebugInfo *di = cpu_llvm_module->debug_info;
+	int fin = BIH_FINDEX(gbl.entbih);
+	LL_Type *type = make_lltype_from_dtype(dtype);
+	OPERAND *ld = make_operand();
+	ld->ot_type = OT_MDNODE;
+	ld->val.sptr = sptr;
+	LL_MDRef lcl = lldbg_emit_local_variable(di, sptr, fin, true);
+
+        /* lets generate llvm.dbg.value intrinsic for it.*/
+	insertLLVMDbgValue(ld, lcl, sptr, type);
+      }
+    }
+  }
+}
+
+/**
    \brief Perform code translation from ILI to LLVM for one routine
  */
 void
@@ -1458,6 +1491,14 @@ restartConcur:
   for (; bih; bih = BIH_NEXT(bih))
     for (ilt = BIH_ILTFIRST(bih); ilt; ilt = ILT_NEXT(ilt))
       build_csed_list(ILT_ILIP(ilt));
+
+  /* process variables with parameter attribute */
+  if (!XBIT(49, 0x10)
+#if defined(OMP_OFFLOAD_PGI) || defined(OMP_OFFLOAD_LLVM)
+      && !gbl.ompaccel_isdevice
+#endif
+     )
+    process_params();
 
   merge_next_block = false;
   bih = BIH_NEXT(0);
@@ -5186,6 +5227,7 @@ insertLLVMDbgValue(OPERAND *load, LL_MDRef mdnode, SPTR sptr, LL_Type *type)
   callOp->next = oper = make_operand();
   oper->ot_type = OT_MDNODE;
   oper->tmps = load->tmps;
+  oper->val = load->val;
   oper->ll_type = type;
   oper->flags |= OPF_WRAPPED_MD;
   oper = make_constval_op(ll_create_int_type(mod, 64), 0, 0);
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -1307,6 +1307,7 @@ static void process_params(void) {
          */
         SPTR new_sptr = (SPTR)CONVAL1G(sptr);
         NMPTRP(new_sptr, NMPTRG(sptr));
+        CCSYMP(new_sptr, 0);
       } else {
         LL_DebugInfo *di = cpu_llvm_module->debug_info;
         int fin = BIH_FINDEX(gbl.entbih);
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -1296,7 +1296,9 @@ remove_dead_instrs(void)
 /**
    \brief process debug info of constants with parameter attribute.
  */
-static void process_params(void) {
+static void
+process_params(void)
+{
   unsigned smax = stb.stg_avail;
   for (SPTR sptr = get_symbol_start(); sptr < smax; ++sptr) {
     DTYPE dtype = DTYPEG(sptr);
@@ -5194,8 +5196,10 @@ gen_gep_index(OPERAND *base_op, LL_Type *llt, int index)
   return gen_gep_op(0, base_op, llt, make_constval32_op(index));
 }
 
-static void insert_llvm_dbg_value(OPERAND *load, LL_MDRef mdnode, SPTR sptr,
-                                  LL_Type *type) {
+static void
+insert_llvm_dbg_value(OPERAND *load, LL_MDRef mdnode,
+                      SPTR sptr, LL_Type *type)
+{
   static bool defined = false;
   OPERAND *callOp;
   OPERAND *oper;
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -1490,8 +1490,9 @@ restartConcur:
     for (ilt = BIH_ILTFIRST(bih); ilt; ilt = ILT_NEXT(ilt))
       build_csed_list(ILT_ILIP(ilt));
 
-  /* process variables with parameter attribute */
-  if (!XBIT(49, 0x10)
+  /* Process variables with parameter attribute to generate debug info, if
+     debug is on. */
+  if (!XBIT(49, 0x10) && flg.debug
 #if defined(OMP_OFFLOAD_PGI) || defined(OMP_OFFLOAD_LLVM)
       && !gbl.ompaccel_isdevice
 #endif
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -116,7 +116,7 @@ static char *fn_sig_ptr = NULL;
 static void insert_entry_label(int);
 static void insert_jump_entry_instr(int);
 static void store_return_value_for_entry(OPERAND *, int);
-static void insertLLVMDbgValue(OPERAND *, LL_MDRef, SPTR, LL_Type *);
+static void insert_llvm_dbg_value(OPERAND *, LL_MDRef, SPTR, LL_Type *);
 
 static int openacc_prefix_sptr = 0;
 static unsigned addressElementSize;
@@ -1294,11 +1294,9 @@ remove_dead_instrs(void)
 }
 
 /**
-   \brief process debug info of variables with parameter attribute.
+   \brief process debug info of constants with parameter attribute.
  */
-void
-process_params(void)
-{
+static void process_params(void) {
   unsigned smax = stb.stg_avail;
   for (SPTR sptr = get_symbol_start(); sptr < smax; ++sptr) {
     if (STYPEG(sptr) == ST_PARAM && should_preserve_param(sptr)) {
@@ -1307,19 +1305,19 @@ process_params(void)
         /* array and derived types have 'var$ac' constant variable
          * lets use that, by renaming that to 'var'.
          */
-	SPTR new_sptr = (SPTR) CONVAL1G(sptr);
-	NMPTRP(new_sptr, NMPTRG(sptr));
+        SPTR new_sptr = (SPTR)CONVAL1G(sptr);
+        NMPTRP(new_sptr, NMPTRG(sptr));
       } else {
-	LL_DebugInfo *di = cpu_llvm_module->debug_info;
-	int fin = BIH_FINDEX(gbl.entbih);
-	LL_Type *type = make_lltype_from_dtype(dtype);
-	OPERAND *ld = make_operand();
-	ld->ot_type = OT_MDNODE;
-	ld->val.sptr = sptr;
-	LL_MDRef lcl = lldbg_emit_local_variable(di, sptr, fin, true);
+        LL_DebugInfo *di = cpu_llvm_module->debug_info;
+        int fin = BIH_FINDEX(gbl.entbih);
+        LL_Type *type = make_lltype_from_dtype(dtype);
+        OPERAND *ld = make_operand();
+        ld->ot_type = OT_MDNODE;
+        ld->val.sptr = sptr;
+        LL_MDRef lcl = lldbg_emit_local_variable(di, sptr, fin, true);
 
         /* lets generate llvm.dbg.value intrinsic for it.*/
-	insertLLVMDbgValue(ld, lcl, sptr, type);
+        insert_llvm_dbg_value(ld, lcl, sptr, type);
       }
     }
   }
@@ -1497,7 +1495,7 @@ restartConcur:
 #if defined(OMP_OFFLOAD_PGI) || defined(OMP_OFFLOAD_LLVM)
       && !gbl.ompaccel_isdevice
 #endif
-     )
+  )
     process_params();
 
   merge_next_block = false;
@@ -5194,9 +5192,8 @@ gen_gep_index(OPERAND *base_op, LL_Type *llt, int index)
   return gen_gep_op(0, base_op, llt, make_constval32_op(index));
 }
 
-static void
-insertLLVMDbgValue(OPERAND *load, LL_MDRef mdnode, SPTR sptr, LL_Type *type)
-{
+static void insert_llvm_dbg_value(OPERAND *load, LL_MDRef mdnode, SPTR sptr,
+                                  LL_Type *type) {
   static bool defined = false;
   OPERAND *callOp;
   OPERAND *oper;
@@ -5252,7 +5249,7 @@ consLoadDebug(OPERAND *ld, OPERAND *addr, LL_Type *type)
     LL_DebugInfo *di = cpu_llvm_module->debug_info;
     int fin = BIH_FINDEX(gbl.entbih);
     LL_MDRef lcl = lldbg_emit_local_variable(di, sptr, fin, true);
-    insertLLVMDbgValue(ld, lcl, sptr, type);
+    insert_llvm_dbg_value(ld, lcl, sptr, type);
   }
 }
 
--- a/tools/flang2/flang2exe/llutil.cpp
+++ b/tools/flang2/flang2exe/llutil.cpp
@@ -2198,10 +2198,22 @@ metadata_args_need_struct(void)
  * This function returns true for the types supported
  * in function make_param_op
  */
-bool should_preserve_param(const DTYPE dtype) {
+bool
+should_preserve_param(const DTYPE dtype)
+{
   switch (DTY(dtype)) {
   // handled cases
   case TY_ARRAY:
+    {
+      ADSC *ad = AD_DPTR(dtype);
+      SPTR size_sptr = AD_NUMELM(ad);
+      ISZ_T size = ad_val_of(size_sptr);
+      /* dont preserve zero-sized array, which would be optimized out later */
+      if (size == 0)
+        return false;
+      else
+        return true;
+    }
   case TY_STRUCT:
   case TY_BLOG:
   case TY_SLOG:
@@ -2230,7 +2242,9 @@ bool should_preserve_param(const DTYPE dtype) {
   }
 }
 
-OPERAND *make_param_op(SPTR sptr) {
+OPERAND *
+make_param_op(SPTR sptr)
+{
   OPERAND *oper;
   DTYPE dtype = DTYPEG(sptr);
 
--- a/tools/flang2/flang2exe/llutil.cpp
+++ b/tools/flang2/flang2exe/llutil.cpp
@@ -2208,7 +2208,7 @@ should_preserve_param(const DTYPE dtype)
       ADSC *ad = AD_DPTR(dtype);
       SPTR size_sptr = AD_NUMELM(ad);
       ISZ_T size = ad_val_of(size_sptr);
-      /* do not preserve zero-sized array, which would be optimized out later */
+      /* Do not preserve zero-sized array, which would be optimized out later */
       if (size == 0)
         return false;
       else
--- a/tools/flang2/flang2exe/llutil.cpp
+++ b/tools/flang2/flang2exe/llutil.cpp
@@ -2195,6 +2195,86 @@ metadata_args_need_struct(void)
 }
 
 /**
+ * This function returns true for the types supported
+ * in function make_param_op
+ */
+bool
+should_preserve_param(SPTR sptr)
+{
+  bool should_preserve = false;
+  DTYPE dtype = DTYPEG(sptr);
+
+  switch (DTY(dtype)) {
+    case TY_ARRAY:
+    case TY_STRUCT:
+    case TY_BLOG:
+    case TY_SLOG:
+    case TY_LOG:
+    case TY_BINT:
+    case TY_SINT:
+    case TY_INT:
+    case TY_REAL:
+    case TY_INT8:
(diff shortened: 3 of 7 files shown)


alokkrsharma:
Gentle reminder !!!

alokkrsharma:
Gentle reminder.

mleair:
Looks good to me.

tools/flang2/flang2exe/llutil.cpp
```diff
@@ -2377,9 +2457,13 @@ write_operand(OPERAND *p, const char *punc_string, int flags)
...
-        new_op = make_var_op(p->val.sptr);
-        if (p->ll_type)
-          new_op->ll_type = ll_get_pointer_type(p->ll_type);
+	if (STYPEG(p->val.sptr) == ST_PARAM) {
+	  new_op = make_param_op(p->val.sptr);
+	} else {
+	  new_op = make_var_op(p->val.sptr);
+	  if (p->ll_type)
+	    new_op->ll_type = ll_get_pointer_type(p->ll_type);
+	}
```
kiranchandramohan, line 2466: Alignment.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1292,6 +1293,38 @@ remove_dead_instrs(void)
...
+	LL_DebugInfo *di = cpu_llvm_module->debug_info;
+	int fin = BIH_FINDEX(gbl.entbih);
+	LL_Type *type = make_lltype_from_dtype(dtype);
+	OPERAND *ld = make_operand();
+	ld->ot_type = OT_MDNODE;
+	ld->val.sptr = sptr;
+	LL_MDRef lcl = lldbg_emit_local_variable(di, sptr, fin, true);
+
+        /* lets generate llvm.dbg.value intrinsic for it.*/
+	insertLLVMDbgValue(ld, lcl, sptr, type);
```
kiranchandramohan, line 1322: Alignment.

tools/flang2/flang2exe/llutil.cpp
```diff
@@ -2194,6 +2194,86 @@ metadata_args_need_struct(void)
...
+ */
+bool
+should_preserve_param(SPTR sptr)
+{
+  bool should_preserve = false;
+  DTYPE dtype = DTYPEG(sptr);
+
+  switch (DTY(dtype)) {
+    case TY_ARRAY:
+    case TY_STRUCT:
```
kiranchandramohan, line 2209: Are these two handled as part of this PR. If yes, please add a comment where else can they be dropped (also in process_params) from this PR

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1459,6 +1492,14 @@ schedule(void)
     for (ilt = BIH_ILTFIRST(bih); ilt; ilt = ILT_NEXT(ilt))
       build_csed_list(ILT_ILIP(ilt));
 
+  /* process variables with parameter attribute */
+  if (!XBIT(49, 0x10)
```
kiranchandramohan, line 1496: can this flag be added to xflag.n? What is its status by default?

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1292,6 +1293,38 @@ remove_dead_instrs(void)
...
+{
+  unsigned smax = stb.stg_avail;
+  for (SPTR sptr = get_symbol_start(); sptr < smax; ++sptr) {
+    if (STYPEG(sptr) == ST_PARAM && should_preserve_param(sptr)) {
+      DTYPE dtype = DTYPEG(sptr);
+      if (DTY(dtype) == TY_ARRAY || DTY(dtype) == TY_STRUCT) {
+        /* array and derived types have 'var$ac' constant variable
+         * lets use that, by renaming that to 'var'.
+         */
+	SPTR new_sptr = (SPTR) CONVAL1G(sptr);
```
schweitzpgi, line 1310: Free casting.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1292,6 +1293,38 @@ remove_dead_instrs(void)
   }
 }
 
+/**
+   \brief process debug info of variables with parameter attribute.
```
schweitzpgi, line 1297: PARAMETERs are constants, not variables.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1292,6 +1293,38 @@ remove_dead_instrs(void)
...
+    if (STYPEG(sptr) == ST_PARAM && should_preserve_param(sptr)) {
+      DTYPE dtype = DTYPEG(sptr);
+      if (DTY(dtype) == TY_ARRAY || DTY(dtype) == TY_STRUCT) {
+        /* array and derived types have 'var$ac' constant variable
+         * lets use that, by renaming that to 'var'.
+         */
+	SPTR new_sptr = (SPTR) CONVAL1G(sptr);
+	NMPTRP(new_sptr, NMPTRG(sptr));
+      } else {
+	LL_DebugInfo *di = cpu_llvm_module->debug_info;
```
schweitzpgi, line 1313: Indentation looks strange here.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1292,6 +1293,38 @@ remove_dead_instrs(void)
   }
 }
 
+/**
+   \brief process debug info of variables with parameter attribute.
+ */
+void
```
schweitzpgi, line 1299: `static` ?

tools/flang2/flang2exe/llutil.cpp
```diff
@@ -2194,6 +2194,86 @@ metadata_args_need_struct(void)
...
+    case TY_INT:
+    case TY_REAL:
+    case TY_INT8:
+    case TY_LOG8:
+    case TY_DBLE:
+    case TY_QUAD:
+    case TY_CMPLX:
+    case TY_DCMPLX:
+    case TY_CHAR:
+      should_preserve = true;
```
schweitzpgi, line 2224: Why not `return true`?

tools/flang2/flang2exe/llutil.cpp
```diff
@@ -2194,6 +2194,86 @@ metadata_args_need_struct(void)
...
+      break;
+  }
+  return should_preserve;
```
schweitzpgi, line 2227: Why not
```c
default:
  return false;
```

tools/flang2/flang2exe/upper.cpp
```diff
@@ -6500,6 +6500,12 @@ lookup_modvar_alias(SPTR sptr)
   return NULL;
 }
 
+SPTR
+get_symbol_start(void)
+{
+  return (SPTR) (oldsymbolcount +1);
```
schweitzpgi, line 6506: What is this magic function doing?
Free casting.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -116,6 +116,7 @@ static char *fn_sig_ptr = NULL;
 static void insert_entry_label(int);
 static void insert_jump_entry_instr(int);
 static void store_return_value_for_entry(OPERAND *, int);
+static void insertLLVMDbgValue(OPERAND *, LL_MDRef, SPTR, LL_Type *);
```
vjayathirtha-nv, line 119: Using camelCase here breaks the pattern that other functions use (underscore naming).

tools/flang2/flang2exe/cgmain.cpp
vjayathirtha-nv: Why is it under an xflag to begin with? Wouldn't we want to have this debug information always available?

tools/flang2/flang2exe/llutil.cpp
vjayathirtha-nv, line 2466:
```suggestion
	else if (STYPEG(p->val.sptr) == ST_PARAM) {
	  new_op = make_param_op(p->val.sptr);
	} else {
	  new_op = make_var_op(p->val.sptr);
	  if (p->ll_type)
	    new_op->ll_type = ll_get_pointer_type(p->ll_type);
	}
```

tools/flang2/flang2exe/cgmain.cpp
gklimowicz: I think I agree that we wouldn't want to bother putting this under an xflag. Once you do that, we have to make sure the driver sets it up as well, don't we? Now I have two pull requests.

tools/flang2/flang2exe/llutil.cpp
```diff
@@ -2194,6 +2194,86 @@ metadata_args_need_struct(void)
...
+}
+
+OPERAND *
+make_param_op(SPTR sptr)
+{
+  OPERAND *oper;
+  DTYPE dtype = DTYPEG(sptr);
+
+  assert(should_preserve_param(sptr), "make_param_op(): unexpected sptr",
```
vjayathirtha-nv, line 2236: Instead of having another check inside of `should_preserve_param()` you could skip this assert and just emit an error in the `default` case of the following switch statement. This would eliminate the need for `should_preserve_param()` function. Why have the knowledge of "supported types" listed separately over 2 different switch statements? Any changes to one will require a change to the other, which seems like  sub-optimal design.

alokkrsharma:
> 
> 
> Thanks for working on this.
> Can some tests be added?

Thanks for suggestion. I am little unaware of flang test framework. I shall understand and add a testcase.

tools/flang2/flang2exe/llutil.cpp
alokkrsharma: Thanks for your comment. Incorporating it.

tools/flang2/flang2exe/cgmain.cpp
alokkrsharma: Thanks for your comment. Incorporating it.

tools/flang2/flang2exe/llutil.cpp
alokkrsharma: Yes it is handled in the PR. I shall add the comment.

tools/flang2/flang2exe/cgmain.cpp
alokkrsharma: Sorry for the confusion here. The changes work with default option, No additional options are needed to make it work (Please notice NOT operator "!"). I just noticed some changes under this macro in semant.c and semant2.c and that suggests it to be some experimental option in absence of that which is by default the current PR would work. Please let me know if you want this to be removed.

tools/flang2/flang2exe/cgmain.cpp
alokkrsharma: Thanks. Incorporating.

tools/flang2/flang2exe/cgmain.cpp
alokkrsharma: Thanks for pointing this out. I have now using clang-format.

tools/flang2/flang2exe/cgmain.cpp
alokkrsharma: Thanks. Incorporating.

tools/flang2/flang2exe/llutil.cpp
alokkrsharma: Yes that would look great. Incorporating this comment.

tools/flang2/flang2exe/llutil.cpp
alokkrsharma: Thanks. That would look great. Incorporating this now.

tools/flang2/flang2exe/upper.cpp
alokkrsharma: This is to get the start of symbols of current routine.

tools/flang2/flang2exe/cgmain.cpp
alokkrsharma: It is an existing function so did not notice earlier. Thanks for pointing it out. Correcting it now.

tools/flang2/flang2exe/llutil.cpp
alokkrsharma: Thanks for suggestion. Incorporating now.

tools/flang2/flang2exe/llutil.cpp
alokkrsharma: Yes I agree that it looks a little odd. But it has to be done to identify the (now) unsupported cases and let it generate an IR without (out of scope) parameter debug-info in place of producing an error which it would have done without this PR.

tools/flang2/flang2exe/llutil.cpp
```diff
@@ -2198,77 +2198,70 @@ metadata_args_need_struct(void)
...
-OPERAND *
-make_param_op(SPTR sptr)
-{
+OPERAND *make_param_op(SPTR sptr) {
   OPERAND *oper;
   DTYPE dtype = DTYPEG(sptr);
 
-  assert(should_preserve_param(sptr), "make_param_op(): unexpected sptr",
-         0, ERR_Fatal);
+  assert(should_preserve_param(sptr), "make_param_op(): unexpected sptr", 0,
```
vjayathirtha-nv, line 2231: > Yes I agree that it looks a little odd. But it has to be done to identify the (now) unsupported cases and let it generate an IR without (out of scope) parameter debug-info in place of producing an error which it would have done without this PR.


You can add unsupported cases in this switch statement qualifying them with a comment which mentions that they are unsupported. These cases need not generate an error. The default case can safely generate errors how any type which are not expected to be handled.

```
switch(DTY(dtype)) {
// handled cases
case TY_BLOG:
.
.
// unsupported cases
case TY_<foo>:
case TY_<foo2>:
    break;
default:
   <error message>
```

tools/flang2/flang2exe/upper.cpp
schweitzpgi: OK. Please document what this is doing someplace. Your explanation and the name of the function don't seem perfectly aligned in my reading.

By the way, this seems like it should already be "a thing" someplace. Is this a duplicate of another function or macro?

Note on free-casting of types: we discourage the use of casts like this because we'd rather that people use well-defined access methods, defined properly in one place, rather than write (TYPE) (variable + int-const) all over the code base. It's a large engineering cost (and bug multiplier) to fix the latter should something change.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1292,6 +1293,36 @@ remove_dead_instrs(void)
...
 }
 
+/**
+   \brief process debug info of constants with parameter attribute.
+ */
+static void process_params(void) {
+  unsigned smax = stb.stg_avail;
+  for (SPTR sptr = get_symbol_start(); sptr < smax; ++sptr) {
+    if (STYPEG(sptr) == ST_PARAM && should_preserve_param(sptr)) {
+      DTYPE dtype = DTYPEG(sptr);
```
kiranchandramohan, line 1303: can should_preserve_param take dtype as argument? If so you hoist the getting of dtype.

tools/flang2/flang2exe/llutil.cpp
```diff
@@ -2194,6 +2194,79 @@ metadata_args_need_struct(void)
   return ll_feature_metadata_args_struct(&llvm_get_current_module()->ir);
 }
 
+/**
+ * This function returns true for the types supported
+ * in function make_param_op
+ */
+bool should_preserve_param(SPTR sptr) {
```
kiranchandramohan, line 2201: can this be marked as a constant function?

tools/flang2/flang2exe/llutil.cpp
alokkrsharma: That looks great. I shall incorporate that.

tools/flang2/flang2exe/cgmain.cpp
alokkrsharma: Thanks for the comment. It will be incorporated.

tools/flang2/flang2exe/llutil.cpp
alokkrsharma: Sure, I shall make that.

alokkrsharma:
> 
> 
> Does this change require dwarf version set to 4? I see that you have a PR for setting it to 4 in the flang-driver.

The current PR doesn't depend on dwarf version. the other PR you are referring is not related. It was to align flang to clang and llvm back-end which have 4 as default dwarf version.

> 
> Do you plan to work on constant variables in modules?

Yes. That would be separate PR.

kiranchandramohan:
> > Does this change require dwarf version set to 4? I see that you have a PR for setting it to 4 in the flang-driver.
> 
> The current PR doesn't depend on dwarf version. the other PR you are referring is not related. It was to align flang to clang and llvm back-end which have 4 as default dwarf version.

This PR worked for me only after setting the dwarf version to 4. Hence I asked.
 
> > Do you plan to work on constant variables in modules?
> 
> Yes. That would be separate PR.

OK.

kiranchandramohan (changes requested):
I think this requires atleast Dwarf 4. Currently if -g is specified the dwarf version is set to 2 in the emitted LLVM IR. I believe the following PR from @alokkrsharma should be merged first. @gklimowicz @shivaramaarao @bryanpkc is it OK to merge the following PR before continuing with this one?
https://github.com/flang-compiler/flang-driver/pull/91

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1292,6 +1293,36 @@ remove_dead_instrs(void)
...
+        SPTR new_sptr = (SPTR)CONVAL1G(sptr);
+        NMPTRP(new_sptr, NMPTRG(sptr));
+      } else {
+        LL_DebugInfo *di = cpu_llvm_module->debug_info;
+        int fin = BIH_FINDEX(gbl.entbih);
+        LL_Type *type = make_lltype_from_dtype(dtype);
+        OPERAND *ld = make_operand();
+        ld->ot_type = OT_MDNODE;
+        ld->val.sptr = sptr;
+        LL_MDRef lcl = lldbg_emit_local_variable(di, sptr, fin, true);
```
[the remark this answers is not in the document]
alokkrsharma: Thanks for the comment. I shall update the patch for this.

test/debug_info/parameter.f90
```diff
@@ -0,0 +1,17 @@
+!RUN: %flang -g -S -emit-llvm %s -o - | FileCheck %s
+
+!CHECK: call void @llvm.dbg.value(metadata i64 99, metadata [[SPAR:![0-9]+]], metadata !DIExpression())
+!CHECK: distinct !DIGlobalVariable(name: "apar"
```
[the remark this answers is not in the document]
alokkrsharma: I am not sure what I am missing here. But for me it shows the expected result regardless of dwarf version.

 ----------------------------------
# flang parameter.f90 -gdwarf-2 -S -emit-llvm
# grep -e "spar" -e "apar" -e "Dwarf Version" parameter.ll
!8 = distinct !DIGlobalVariable(name: "apar", scope: !2, file: !3, type: !9, isLocal: true, isDefinition: true)
!15 = !{i32 2, !"Dwarf Version", i32 2}
!17 = !DILocalVariable(name: "spar", scope: !18, file: !3, type: !10)
  ------------------------------------
I am using this driver.

$ git branch
  70_mod
  master
  release_60
* release_70
$ git remote -v
origin  https://github.com/flang-compiler/flang-driver.git (fetch)
origin  https://github.com/flang-compiler/flang-driver.git (push)
----------------- 
Can you please share the details or your sources (open) ?
In case it works for you on dwarf-4 (with DWARF related PRs), it would be better to first commit below PRs
#863 #89 #90 #91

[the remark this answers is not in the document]
kiranchandramohan: There are two separate issues here.
1) Not able to see !8 = distinct !DIGlobalVariable(name: "apar", scope: !2, file: !3, type: !9, isLocal: true, isDefinition: true)
This is because i am building with FLANG_LLVM_EXTENSIONS. If i do not use that cmake variable then I can see the debug for apar.

2) Not able to see values of spar in gdb with default dwarf version =2.
```
Breakpoint 1, main () at param.f90:6
6         svar = spar
(gdb) info locals
svar = 0
spar = <error reading variable spar (Cannot access memory at address 0x63)>
apar = (99, 98, 97, 96, 95)
avar = (0, 0, 0, 0, 0)
```
If i manually change the dwarf version to 4 in the LLVM IR generated by flang then I get the following in gdb.
```
Breakpoint 1, main () at param.f90:6
6         svar = spar
(gdb) info locals
svar = 0
spar = 99
apar = (99, 98, 97, 96, 95)
avar = (0, 0, 0, 0, 0)
```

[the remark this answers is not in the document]
alokkrsharma: Thanks, this helps. 
Regarding first issue, I never tested with FLANG_LLVM_EXTENSION. Now I am able to reproduce this issue. I shall shortly push the fix to make it work with the FLANG_LLVM_EXTENSION as well.

kiranchandramohan (commented):
Thanks, the flang unit tests pass for me now. Now will run some internal tests.

kiranchandramohan:
Saw a few failures from the tests. Could you have a look?
```
program t31
  character(*), parameter :: x1 = achar(33)//achar(1)//achar(35)
  print *, x1
end
```
```
program t32
  use iso_c_binding
  character(*),parameter :: compiler = 'flang'//c_null_char
  print *, compiler
end
```
```
program t33
  integer, parameter :: indx(0) = 4
  real, parameter :: arr(10) = [ 0.0,1.0,2.0,3.0,4.0,5.0,6.0,7.0,8.0,9.0 ]
  real :: y(10) = arr
  call internal
  print *, "success"
contains
  subroutine internal
    y(indx0) = -4
  end subroutine
end program
```

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1292,6 +1293,37 @@ remove_dead_instrs(void)
   }
 }
 
+/**
+   \brief process debug info of constants with parameter attribute.
+ */
+static void process_params(void) {
```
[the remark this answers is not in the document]
alokkrsharma: Thanks for your comment. I shall fix this in next patch.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -5153,9 +5194,8 @@ gen_gep_index(OPERAND *base_op, LL_Type *llt, int index)
   return gen_gep_op(0, base_op, llt, make_constval32_op(index));
 }
 
-static void
-insertLLVMDbgValue(OPERAND *load, LL_MDRef mdnode, SPTR sptr, LL_Type *type)
-{
+static void insert_llvm_dbg_value(OPERAND *load, LL_MDRef mdnode, SPTR sptr,
+                                  LL_Type *type) {
```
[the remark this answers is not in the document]
alokkrsharma: Thanks for your comment. I shall fix this in next patch.

tools/flang2/flang2exe/llutil.cpp
```diff
@@ -2194,6 +2194,82 @@ metadata_args_need_struct(void)
   return ll_feature_metadata_args_struct(&llvm_get_current_module()->ir);
 }
 
+/**
+ * This function returns true for the types supported
+ * in function make_param_op
+ */
+bool should_preserve_param(const DTYPE dtype) {
```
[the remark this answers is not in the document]
alokkrsharma: Thanks for your comment. I shall fix this in next patch.

tools/flang2/flang2exe/llutil.cpp
```diff
@@ -2194,6 +2194,82 @@ metadata_args_need_struct(void)
...
+  case TY_HOLL:
+  case TY_NCHAR:
+    return false;
+  default:
+    assert(0, "should_preserve_param(dtype): unexpected DTYPE", 0, ERR_Fatal);
+    return false;
+  }
+}
+
+OPERAND *make_param_op(SPTR sptr) {
```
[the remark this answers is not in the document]
alokkrsharma: Thanks for your comment. I shall fix this in next patch.

alokkrsharma:
> 
> 
> Saw a few failures from the tests. Could you have a look?
> 
> ```
> program t31
>   character(*), parameter :: x1 = achar(33)//achar(1)//achar(35)
>   print *, x1
> end
> ```
Yes, the issue related to string (character) constants are due to existing issue related to bad representation of character constants. It has a separate PR #921 which should be committed before #888 .
> 
> ```
> program t32
>   use iso_c_binding
>   character(*),parameter :: compiler = 'flang'//c_null_char
>   print *, compiler
> end
> ```
> 
Yes, the issue related to string (character) constants are due to existing issue related to bad representation of character constants. It has a separate PR #921 which should be committed before #888 .
> ```
> program t33
>   integer, parameter :: indx(0) = 4
>   real, parameter :: arr(10) = [ 0.0,1.0,2.0,3.0,4.0,5.0,6.0,7.0,8.0,9.0 ]
>   real :: y(10) = arr
>   call internal
>   print *, "success"
> contains
>   subroutine internal
>     y(indx0) = -4
>   end subroutine
> end program
> ```
It has two issues, one related to zero sized array (which is optimized out), which is fixed in latest patch. 
Second one is when variable array is initialized with parameter array, the parameter is optimized out. in case two variables are initialized with same parameter array, one parameter and one variable is optimized out. This is separate issue and i am working on that but the current PR should be able to go in with or without that fix in case there are no more issue.

alokkrsharma:
The order of PRs before current (#888) is as follows.
1. #863 (flang) - approved
2. #89 (driver) - approved
3. #90 (driver) - approved
4. #91 (driver) - approved
5. #921 (flang)

kiranchandramohan:
> 2. #89 (driver) - approved
> 3. #90 (driver) - approved
> 4. #91 (driver) - approved

The driver links are not correct. Corrected links below.
https://github.com/flang-compiler/flang-driver/pull/89
https://github.com/flang-compiler/flang-driver/pull/90
https://github.com/flang-compiler/flang-driver/pull/91

gklimowicz (approved):
I tested this PR along with #863 and #921, and the Flang-driver PR#92 and PR#93. These pass tests on OpenPOWER.

kiranchandramohan:
LGTM.
I will now run some internal tests and get back to you in a couple of days.

kiranchandramohan (approved):
Internal tests passed. Thanks for your work and patience.

Please squash and use "Rebase and merge" for all PRs you are committing.

shivaramaarao:
@bryanpkc : I believe this change is also good to go. please confirm otherwise.

tools/flang2/flang2exe/llutil.cpp
```diff
@@ -2194,6 +2194,96 @@ metadata_args_need_struct(void)
...
+should_preserve_param(const DTYPE dtype)
+{
+  switch (DTY(dtype)) {
+  // handled cases
+  case TY_ARRAY:
+    {
+      ADSC *ad = AD_DPTR(dtype);
+      SPTR size_sptr = AD_NUMELM(ad);
+      ISZ_T size = ad_val_of(size_sptr);
+      /* dont preserve zero-sized array, which would be optimized out later */
```
bryanpkc, line 2211: Is this true even when compiling at `-O0`?

tools/flang2/flang2exe/llutil.cpp
```diff
@@ -2194,6 +2194,96 @@ metadata_args_need_struct(void)
...
+  }
+}
+
+OPERAND *
+make_param_op(SPTR sptr)
+{
+  OPERAND *oper;
+  DTYPE dtype = DTYPEG(sptr);
+
+  switch (DTY(dtype)) {
```
bryanpkc, line 2251: Isn't this `switch` supposed to handle `TY_ARRAY` and `TY_STRUCT` as well?

bryanpkc: OK, now I understand why. It would be useful to add a comment here to explain the inconsistency between this function and `should_preserve_param` immediately above.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1459,6 +1493,15 @@ schedule(void)
     for (ilt = BIH_ILTFIRST(bih); ilt; ilt = ILT_NEXT(ilt))
       build_csed_list(ILT_ILIP(ilt));
 
+  /* Process variables with parameter attribute to generate debug info, if
+     debug is on. */
+  if (!XBIT(49, 0x10) && flg.debug
```
bryanpkc, line 1498: Could you document this use of `XBIT(49, 0x10)` in `tools/flang2/docs/xflag.n`?

tools/flang2/flang2exe/llutil.cpp
bryanpkc: BTW, please spell the first word with an apostrophe, or write "Do not".

tools/flang2/flang2exe/cgmain.cpp
alokkrsharma: This is an existing option you can find other usage of it in files semant.c and semant2.c, which was probably earlier introduced to guard (possibly unfinished) changes.
the current condition is to make sure that experimental (and undocumented) option is not mentioned. if user is not aware of that option this condition will always pass.

tools/flang2/flang2exe/llutil.cpp
alokkrsharma: Yes, you are right. I shall add comment here.

tools/flang2/flang2exe/llutil.cpp
alokkrsharma: Yes that is optimized out in front-end itself. I shall update your comment. Thanks.

tools/flang2/flang2exe/cgmain.cpp
bryanpkc: My understanding of xflags is that they should already be considered internal options. Unlike driver options, they are not designed to be used by end users. Not having a definitive list of xflags and a version-controlled way to reserve bits makes it very confusing for developers, IMHO. I think we should discuss this "current condition" but it is not a problem that should block this pull request.

bryanpkc (approved):
LGTM. Thanks for your patience.

PR #895, opened 2020-05-16
[DebugInfo]Fix for redundant lexical block in debug info

SouraVX:
Flang was creating a redundant lexical block at the beginning of of every subroutine debug info. This was causing debuggers to incorrectly assume the scope of a variable to be this lexical block.

This patch fixes this incorrect behavior by replacing all the debug info references of this redundant lexical block to the actual subroutine.

Brief explanation and correct behavior can be found here:
http://lists.llvm.org/pipermail/flang-dev/2020-April/000277.html
https://patchwork.ozlabs.org/patch/436971/

After this, GDB is correctly able to show formal arguments for the case of pass by value parameters.

### DIFF
--- a/tools/flang2/flang2exe/lldebug.cpp
+++ b/tools/flang2/flang2exe/lldebug.cpp
@@ -1810,7 +1810,9 @@ lldbg_emit_lexical_block(LL_DebugInfo *db, int sptr, int lineno, int findex,
   NEEDB((db->blk_idx + 1), db->blk_tab, BLKINFO, db->blk_tab_size,
         (db->blk_tab_size + 64));
   db->blk_tab[db->blk_idx].mdnode =
-      lldbg_create_block_mdnode(db, parent_blk_mdnode, lineno, 1, findex, ID++);
+      STYPEG(sptr) == ST_BLOCK ? lldbg_create_block_mdnode(db,
+      parent_blk_mdnode, lineno, 1, findex, ID++)
+      : parent_blk_mdnode;
   db->blk_tab[db->blk_idx].sptr = sptr;
   db->blk_tab[db->blk_idx].startline = startline;
   db->blk_tab[db->blk_idx].endline = endline;
--- a/test/debug_info/redundant-lexicalblock.f90
+++ b/test/debug_info/redundant-lexicalblock.f90
@@ -1,0 +1,22 @@
+!RUN: %flang -g -S -emit-llvm %s -o - | FileCheck %s
+
+!Ensure that there is no redundant LexicalBlock created with scope
+!pointing to Subprogram.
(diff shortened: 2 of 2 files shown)


SouraVX:
Gentle Ping @Jiel-nv!

SouraVX:
Thank you for your reviewing this!

> Actually, instead of checking STYPE on each cur_line_mdnode, it could be easier to set the root block mdnode to be the subprogram itself. Fortran doesn't have lexical block and nested lexical block. So it seems to me, a simple one line change might do the trick.

With Fortran 2008 we can have lexical blocks with BLOCK statement:
`BLOCK`
`  REAL BLOCK_LOCAL`
`  ....`
` END BLOCK`

Anyways I've flipped the approach similar to what you mentioned and fortunately this covers this above test scenarios also. 
So we're all set to no redundant block in LLVM IR and the resultant DWARF. And correct Lexical Block debug info(From IR to DWARF) for the BLOCK statement💯

SouraVX:
Rebased on top of master, added test case.
@kiranchandramohan , please have look!
Thanks!

tools/flang2/flang2exe/lldebug.cpp
```diff
@@ -1810,7 +1810,9 @@ lldbg_emit_lexical_block(LL_DebugInfo *db, int sptr, int lineno, int findex,
   NEEDB((db->blk_idx + 1), db->blk_tab, BLKINFO, db->blk_tab_size,
         (db->blk_tab_size + 64));
   db->blk_tab[db->blk_idx].mdnode =
-      lldbg_create_block_mdnode(db, parent_blk_mdnode, lineno, 1, findex, ID++);
+      STYPEG(sptr) == ST_BLOCK ? lldbg_create_block_mdnode(db,
+		      parent_blk_mdnode, lineno, 1, findex, ID++)
```
kiranchandramohan, line 1814: There are tabs here I think. Could you check and remove?

kiranchandramohan (commented):
One minor formatting comment.
I am running some internal tests and will get back to you in a couple of days.

tools/flang2/flang2exe/lldebug.cpp
SouraVX: Not visible from diff, but after setting list there is some special character "^I". Thanks for mentioning this!
Anyway should I squash(all the commits) it or do an extra commit for this formatting correction ?

kiranchandramohan: Please go ahead and squash.

kiranchandramohan (approved):
LGTM. Internal tests passed.

SouraVX:
Although this PR is already approved by @jiel-nv . @gklimowicz could you please run internal tests ? so that we can move ahead.

gklimowicz:
Testing today...

PR #900, opened 2020-06-04
[DebugInfo] Basic testing infra for debug-info testing

SouraVX:
This patch adds minimal support for debug-info testing
infrastructure.

Post this debug-info related development should use this for
adding incremental test cases for regression and verification.

Discussion GitHub Issue:
https://github.com/flang-compiler/flang/issues/899

PR #909, opened 2020-07-06
Debugger is not able to print correct variable values from modules of library

NitikaAchra:
Test case used:

filename: library-module-main.f90

use lib
call lib_func
if (var_i .ne. 2) call abort
var_i = var_i
end

filename: library-module-lib.f90

module lib
        integer :: var_i = 1
contains
        subroutine lib_func
        var_i = 2
        var_i = var_i
        end subroutine lib_func
end module lib

flang -gdwarf-4  -fpic -c -o library-module-lib.f90.o library-module-lib.f90
flang library-module-lib.f90.o -gdwarf-4  -shared -Wl,-soname,library-module-lib.so -g -lm -o library-module-lib.so
flang -Wl,--no-as-needed library-module-main.f90 library-module-lib.so -gdwarf-4 -g -Wl,-rpath,$ORIGIN -lm -o library-module

gdb -q library-module -ex "start" -ex "b library-module-lib.f90:7" -ex "continue" 

Reading symbols from library-module...
Temporary breakpoint 1 at 0x400854: -qualified MAIN. (2 locations)
Starting program: /home/nitachra/library-module
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Temporary breakpoint 1, main
 (argc=1,
argv=0x7ffffffee208) at /home/nitachra/aocc/llvm-project/flang/runtime/flangmain/flangmain.c:46
46        int i = 0;
Breakpoint 2 at 0x7fffff1f0692: file library-module-lib.f90, line 7.
Continuing.

Breakpoint 2, lib::lib_func () at library-module-lib.f90:7
7               end subroutine lib_func
(gdb) p var_i
$1 = 1
(gdb)

This is because DW_AT_linkage_name is not getting emitted in debug info global
variables. This patch fixes this.

This patch has the dependency on https://github.com/flang-compiler/flang/pull/900/commits/211ee32c223f04178656882c3306b56b3850d046
for adding test case.

### DIFF
--- a/test/debug_info/fail.f90
+++ b/test/debug_info/fail.f90
@@ -1,0 +1,7 @@
+! Simple test to check for a failing case.
+
+! RUN: %flang -g -S -emit-llvm %s -o - | FileCheck %s
+! CHECK: NULL
+
+subroutine foo
+end subroutine
--- a/test/debug_info/linkage_name.f90
+++ b/test/debug_info/linkage_name.f90
@@ -1,0 +1,12 @@
+
+! RUN: %flang -gdwarf-4  -fpic -c  %s -o %t
+! RUN: llvm-dwarfdump -debug-info %t | FileCheck %s
+! CHECK:  DW_TAG_variable
+! CHECK:  DW_AT_linkage_name ("_lib_8_")
+module lib
+        integer :: var_i = 1
+contains
+        subroutine lib_func
+        var_i = 2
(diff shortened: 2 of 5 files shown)


SouraVX:
This patch incorrectly puts same `linkage_name`, closing this after internal discussion.

PR #912, opened 2020-07-27
libflang: Fix strcmp_klen return value type

nextsilicon-itay-bookstein:
Ftn_strcmp_klen return value type should be int like
its similar sibling functions and like the compiler
assumes when generating a call to this function.

Signed-off-by: Itay Bookstein <itay.bookstein@nextsilicon.com>

### DIFF
--- a/runtime/flang/ftncharsup.c
+++ b/runtime/flang/ftncharsup.c
@@ -493,7 +493,7 @@ int Ftn_str_index(  const unsigned char * const a1,
  */
 /* ***********************************************************************/
 
-int64_t
+int
 Ftn_strcmp_klen(    const unsigned char * const a1,
                     const unsigned char * const a2,
                     int64_t a1_len,


pawosm-arm (approved):
considering this function always returns one of { -1, 0, 1 } the chosen return type was an overkill.

bryanpkc:
What's the difference between this function and `Ftn_kstrcmp_klen` in `ftni64misc.c`? Do their prototypes need to remain identical?

Did you check corner cases to make sure that this will produce correct behaviour? The code in `exp_rte.cpp` seems to suggest that the return value could, in some cases, be a 64-bit integer:

https://github.com/flang-compiler/flang/blob/master/tools/flang2/flang2exe/exp_rte.cpp#L5393-L5398

nextsilicon-itay-bookstein:
Ping

bryanpkc (approved):

PR #902, opened 2020-06-06
flang gen-exec doesnt print allocatable array in debugger

alokkrsharma:
Note: This modification makes use of recent upgrade to DISubrange
    https://reviews.llvm.org/rGd20bf5a7258d4b6a7f017a81b125275dac1aa166

    This PR has two check-ins, first one is up for review as https://github.com/flang-compiler/flang/pull/901 . I intend to get only second check-in to get reviewed here. once other review is completed i shall re-base current one.

    Now, below testcase now gives desired results with debugger.
    --------------
    program main
      integer, allocatable  :: arr(:)
      allocate (arr(2:10))
      arr(3) = 9
    end program main
    --------------
    (gdb) pt arr
    type = integer (2:10)
    (gdb) p arr
    $1 = (0, 9, 0, 0, 0, 0, 0, 0, 0)
    --------------
    And below case works as well
    --------------
    program main
    type dt
      integer :: var1
      integer :: var2
      integer, allocatable :: arr1 (:,:)
      integer :: var3
    end type dt
    type(dt) :: dvar1
    dvar1%var1 = 21
    dvar1%var2 = 22
    dvar1%var3 = 23
    allocate (dvar1%arr1(2:9, 3:11))
    dvar1%arr1(5,4) = 11
    end program main
    --------------
    (gdb) pt dvar1
    type = Type dt
    integer :: var1
    integer :: var2
    integer :: arr1(2:9,3:11)
    integer :: var3
    End Type dt
    (gdb) p dvar1%arr1
    $2 = (( 0, 0, 0, 0, 0, 0, 0, 0) ( 0, 0, 0, 11, 0, 0, 0, 0) ( 0, 0, 0, 0, 0, 0, 0, 0) ( 0, 0, 0, 0, 0, 0, 0, 0) ( 0, 0, 0, 0, 0, 0, 0, 0) ( 0, 0, 0,
 0, 0, 0, 0, 0) ( 0, 0, 0, 0, 0, 0, 0, 0) ( 0, 0, 0, 0, 0, 0, 0, 0) ( 0, 0, 0, 0, 0, 0, 0, 0) )
    ---------------

PR #901, opened 2020-06-06
flang gen-exec doesnt show assumed shape array in debugger

alokkrsharma:
Note: This modification makes use of recent upgrade to DISubrange
    https://reviews.llvm.org/rGd20bf5a7258d4b6a7f017a81b125275dac1aa166

  Currently flang generates 0 sized DISubrange metadata for assumed sahpe arrays.

  ---------
  test case
  ---------
```
  subroutine sub(assume,n1,n2)
    dimension assume(n1,n2)
  end subroutine sub
```
  -------
  LLVM IR
  -------
```
  !29 = !DILocalVariable(name: "assume", arg: 1, scope: !26, file: !3, type: !30)
  !30 = !DICompositeType(tag: DW_TAG_array_type, baseType: !10, size: 32, align: 32, elements: !31)
  !31 = !{!32, !32}
  !32 = !DISubrange(count: 0, lowerBound: 1)
```
  -------
  This is fixed by upgrading DISubrange as it is upgraded in LLVM to be able to
  dump proper DISubrange for assumed shape array.
  -------
```
  !29 = !DILocalVariable(name: "assume", arg: 1, scope: !26, file: !3, type: !30)
  !30 = !DICompositeType(tag: DW_TAG_array_type, baseType: !10, size: 32, align: 32, elements: !31)
  !31 = !{!32, !36}
  !32 = !DISubrange(lowerBound: 1, upperBound: !33)
  !33 = distinct !DILocalVariable(scope: !34, file: !3, type: !35, flags: DIFlagArtificial)
  !36 = !DISubrange(lowerBound: 1, upperBound: !37)
  !37 = distinct !DILocalVariable(scope: !34, file: !3, type: !35, flags: DIFlagArtificial)
```
  -------

PR #913, opened 2020-07-29
flang gen-exec does not generate DW_AT_associated or DW_AT_allocated

alokkrsharma:
This patch makes use of commited llvm patch
          [DebugInfo] Support for DW_AT_associated and DW_AT_allocated.
          https://reviews.llvm.org/rG2d10258a31a6d05368dfd4442c4aaa7b54614f1e

        This PR has multiple check-ins, first four are already up for review as https://github.com/flang-compiler/flang/pull/901 and #902 . I intend to get only fifth check-in to get reviewed here. once other reviews are completed, the current one will be re-based.

        This fix is needed for allocatable/pointer arrays.

        for pointer array (before allocation/association)
        Before fix
        ------
        (gdb) pt ptr
        type = integer (140737345375288:140737354129776)
        (gdb) p ptr
        value requires 35017956 bytes, which is more than max-value-size
        ------
        After fix
        ------
        (gdb) pt ptr
        type = integer (:)
        (gdb) p ptr
        $1 = <not associated>
        ------
        for allocatable array (before allocation)
        Before fix
        ------
        (gdb) pt arr
        type = integer (140737345375288:140737354129776)
        (gdb) p arr
        value requires 35017956 bytes, which is more than max-value-size
        ------
        After fix
        ------
        (gdb) pt arr
        type = integer, allocatable (:)
        (gdb) p arr
        $1 = <not allocated>
        ------

    Testing:
        - check-flang

### DIFF
--- a/test/debug_info/allocated.f90
+++ b/test/debug_info/allocated.f90
@@ -1,7 +1,7 @@
 !RUN: %flang -g -S -emit-llvm %s -o - | FileCheck %s
 
 !CHECK: !DILocalVariable(name: "arr", scope: {{![0-9]+}}, file: {{![0-9]+}}, type: [[TYPE:![0-9]+]])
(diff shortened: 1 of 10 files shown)


alokkrsharma:

PR #916, opened 2020-08-25
[runtime] Fix off-by-one error in minimum integers

erjin:
runtime/flang/const.c defines min_int{1,2,4,8} to be -max_int{1,2,4,8}, which is wrong for the two's complement representation. As a result, the MAXLOC runtime function will compare each element in the input array with an incorrect minimum value. If the array contains -huge(integer)-1, MAXLOC will return an unexpected result.

### DIFF
--- a/test/f90_correct/src/maxloc14.f90
+++ b/test/f90_correct/src/maxloc14.f90
@@ -1,0 +1,15 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+! maxloc with minimum value of integer type as elements
+
+program t
+  integer, dimension(2) :: rslt, expect
+  integer, dimension(2, 2) :: m
+  m = -huge(integer) - 1
+  rslt = maxloc(m, dim = 1)
+  expect = (/1, 1/)
+  call check(rslt, expect, 2)
+end
--- a/runtime/flang/const.c
+++ b/runtime/flang/const.c
@@ -677,10 +677,10 @@ __fort_init_consts()
 
   max_str = (__STR_T) 255;
 
-  min_int1 = -max_int1;
-  min_int2 = -max_int2;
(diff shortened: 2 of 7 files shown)


gklimowicz:
From a practical perspective (how many ones-complement machines are there?), this seems OK.

It's not clear to me without more digging, though, if this is legal Fortran. The definition of model integers in the standard (F2018 subclause 16.4) defines model integers only through the equivalent of -HUGE(the integer kind under consideration).

bryanpkc:
@gklimowicz I read that part of the language standard and I was wondering about the same thing. However, neither flang nor gfortran complains about the use of `-HUGE(integer) - 1`, and gfortran actually generates MAXLOC code that runs correctly for the test case in this PR.

gklimowicz:
@bryanpkc Yeah. It's hard to argue against that. I will stop. :-)

test/f90_correct/src/maxloc14.f90
```diff
@@ -0,0 +1,16 @@
...
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+! maxloc with minimum value of integer type as elements
+
+program t
+  integer, dimension(2) :: rslt, expect
+  integer, dimension(2, 2) :: m
+  m(1, :) = (/-huge(integer) - 1, -huge(integer) - 1/)
+  m(2, :) = (/-huge(integer) - 1, -huge(integer) - 1/)
```
kiranchandramohan, line 12: m = -huge(integer) - 1 would work here right?

kiranchandramohan (approved):
LGTM. A trivial comment inline (can be ignored).

test/f90_correct/src/maxloc14.f90
bryanpkc: Yes, it would. Thanks for the suggestion.

shivaramaarao (approved):
LGTM. thanks

kiranchandramohan:
@bryanpkc please merge if no further changes required. You have approval from everyone i guess.

bryanpkc:
Thanks @kiranchandramohan.

PR #917, opened 2020-08-27
f90_correct: fix rank mismatch in dt17.f90

bryanpkc:
Line 45 in `test/f90_correct/src/dt17.f90` contains a rank mismatch which causes Flang to emit a warning:
```
F90-W-0155-The number of subscripts is less than the rank of ia (test/f90_correct/src/dt17.f90: 45)
  0 inform,   1 warnings,   0 severes, 0 fatal for check2
```
Its sibling test, `dt17b.f90`, uses the correct rank for `ia` on the same line. Presumably this is just a typo in the first test case; this PR fixes the typo.

Incidentally, gfortran gives an error on this test case:
```
dt17.f90:45:81:

   45 |       write(*,'(a,a,i1,a,i1,a,i4,a,i4)') s,'a(',i,',',j,')%i=',a(i,j)%i,' != ',ia(i)
      |                                                                                 1
Error: Rank mismatch in array reference at (1) (1/2)
```

Is there any reason why Flang only gives a warning and not a severe error?

### DIFF
--- a/test/f90_correct/src/dt17.f90
+++ b/test/f90_correct/src/dt17.f90
@@ -42,7 +42,7 @@
 	  do i = lbound(a,1),ubound(a,1)
 	   do j = lbound(a,2),ubound(a,2)
 	    if(a(i,j)%i .ne. ia(i,j) )then
-	     write(*,'(a,a,i1,a,i1,a,i4,a,i4)') s,'a(',i,',',j,')%i=',a(i,j)%i,' != ',ia(i)
+	     write(*,'(a,a,i1,a,i1,a,i4,a,i4)') s,'a(',i,',',j,')%i=',a(i,j)%i,' != ',ia(i,j)
 	     r = 0
 	    endif
 	    do k = lbound(ja,1),ubound(ja,1)


gklimowicz:
Looking back at the history of this change in the original PGI compiler, it looks like this was an intentional extension. When you compile with `-Mstandard`, it flags this as an error.

bryanpkc:
I tried adding `-Mstandard` already; Flang still just prints a warning and not a severe error.

gklimowicz:
Oh, sorry, yes. The only real difference is that with `-Mstandard`, the warning comes with a different warning number and the message that this use is an extension.

I think in general, even under `-Mstandard`, flang treats non-standard Fortran or use of extensions as warning and not as severe errors. That was probably considered a _feature_ at the time the code was written. 

@mleair may be able to provide some additional insights into the underlying philosophy here.

bryanpkc:
In `ref_array` (tools/flang1/flang1exe/semutil.c), I think I understand the code that handles "overindexed subscript reference" (the Cray Fortran documentation states that `nooverindex` is the default). But I don't understand the part that handles "a member reference of a subscripted derived type"; I wasn't able to construct a rank mismatch example where `SST_IDG(stktop) == S_DERIVED`. Could you give me some pointer?

PR #918, opened 2020-08-28
Updating the test script to work with python 3.x

shivaramaarao:
The script check_compilation.py is written for python 2.x and many tests fail when python 3.x is used. 
The print statements used in the script such as print title are not compatible with python 3.x 
The script is updated to support both python 2.x and pyton 3.x.

### DIFF
--- a/test/tools/check_compilation.py
+++ b/test/tools/check_compilation.py
@@ -22,6 +22,7 @@ not match the original source info.
 """
 
 import re
+import sys
 import argparse
 from collections import namedtuple
 from os.path import basename
@@ -101,7 +102,8 @@ def getLogErrors(log):
 			elif terminatedError.search(line):
 				terminated = True
 			else:
-				print line,
+				sys.stdout.write (line)
+				sys.stdout.write (" ")
 	return terminated, errorList
 
 def getFileSet(source, logErrors):
@@ -166,14 +168,14 @@ def checkErrorsMatch(terminated, logErrors, sourceErrors, compiler):
 	for e in logErrors:
 		if not removeMatch(e, sourceErrors, compiler):
 			fail = True
-			print "Unexpected:", e.sev, e.file, e.lineno, e.text
+			print ("Unexpected:", e.sev, e.file, e.lineno, e.text)
 	for e in sourceErrors:
 		fail = True
-		print "Missed:", e.sev, e.file, e.lineno, e.text
+		print ("Missed:", e.sev, e.file, e.lineno, e.text)
 	if fail or terminated:
-		print "FAIL"
+		print ("FAIL")
 	else:
-		print "PASS"
+		print ("PASS")
 
 def removeMatch(e, errors, compiler):
 	"""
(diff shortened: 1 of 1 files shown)


test/tools/check_compilation.py
```diff
@@ -101,7 +101,7 @@ def getLogErrors(log):
 			elif terminatedError.search(line):
 				terminated = True
 			else:
-				print line,
+				print (line, end=" ")
```
kiranchandramohan, line 104: This line shows issues with python 2.
  File "flang/test/f90_correct/lit/../../tools/check_compilation.py", line 104
    print (line, end=" ")

bryanpkc: `sys.stdout.write` would work.

shivaramaarao: thanks for pointing it out. will fix it and submit new patch.

test/tools/check_compilation.py
```diff
@@ -101,7 +102,7 @@ def getLogErrors(log):
 			elif terminatedError.search(line):
 				terminated = True
 			else:
-				print line,
+				sys.stdout.write (line)
```
kiranchandramohan, line 105: Not a python expert but wouldn't we lose the space between printed lines if we directly use sys.stdout.write?

bryanpkc: The original Python 2 statement `print line,` with a trailing comma, would have skipped the newline. `sys.stdout.write` gives the same behavior, and works with both Python 2 and 3.

kiranchandramohan: Newline is fine and it works with both 2 and 3. The question is about space.

I was trying the python program below and with python 2 it gave the following output. Where the print with comma added space but sys.stdout.write does not add a space.
one two three
onetwothree
```
import sys
mylist = ["one", "two", "three"]
for l in mylist:
    print l,
print ""
for l in mylist:
    sys.stdout.write(l)
```

shivaramaarao: yes. this may be a problem and print (line, endl=" ") is the correct code for python 3.x. 
Looks like I need to add additional sys.stdout.write(" ")  line. would that be fine?

kiranchandramohan (approved):
LGTM

sscalpone:
Who tested this patch with python2 and python3?

shivaramaarao:
> Who tested this patch with python2 and python3?

Hi Steve, I tested with python 2.7 and python 3.6.  do you think some more testing need to be done?

sscalpone:
@shivaramaarao Hi, that's great.  Thanks

kiranchandramohan:
I tested on python2 and on python3 (by changing the invocation to use python3 and also setting python3 as alias for python).

```
openmp_examples/makefile:COMP_CHECK = python3 $(BASE_DIR)/../tools/check_compilation.py
f90_correct/makefile:COMP_CHECK=python3 $(HOMEQA)/../tools/check_compilation.py
```

gklimowicz:
I also tested on Power with Python2 and Python3 by changing the scripts that set up the location of the Python executable.

Issue #670, opened 2019-02-11
Internal compiler error. mk_id: invalid symbol table index 0


bcfriesen:
Consider the following code:

```fortran
program main
  implicit none
  integer,parameter :: kind_param = 4
  integer(kind_param),parameter :: x = Z"FF"
  integer(kind_param) :: n
  integer(kind_param),parameter :: a(*) = [ Z"d76aa478", Z"e8c7b756"]
  enum, bind(c)
     enumerator :: param_foo = Z"2000" ! in C, 0x0200
     enumerator :: param_bar = Z"4000" ! in C, 0x4000
  end enum
  integer :: buffer = 5000, switch = 2

   n = O"034"
   print *, "n = ",n

   if (buffer > Z"3F") then
      print *, "Buffer is bigger"
   end if

   select case (switch)
      case (Z"01")
         print *, "bad select case branch"
      case (Z"02")
         print *, "Good select case branch"
   end select

 end program main
```

This code relies on non-standard treatment of boz literal constants which have been adopted in some compilers. So a compile-time error would be appropriate behavior. However, flang encounters an ICE:

```bash
> flang --version
clang version 6.0.1 (https://github.com/flang-compiler/flang-driver.git f90d905b99ef497cc654b4a8433f1ac35a870f11) (https://github.com/flang-compiler/llvm.git e759b6cedf4ea8d3072eda356e1ded21d1341a37)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/local/flang/bin
> flang main.f90
F90-S-0084-Illegal use of symbol a - a named constant array must have constant extents (main.f90: 6)
F90-S-0000-Internal compiler error. mk_id: invalid symbol table index       0  (main.f90: 6)
F90-S-0000-Internal compiler error. sym_of_ast: unexpected ast      26  (main.f90: 6)
  0 inform,   0 warnings,   3 severes, 0 fatal for main
```

Incidentally, the PGI 18.10 compiler encounters the same ICE; it is being tracked as NVIDIA bug 2508942.

bryanpkc:
This is no longer happening with the latest master branch.

PR #919, opened 2020-08-28
Fix issue #670: Internal compiler error

gklimowicz:
The issue as reported is about the use of BOZ constants,
but the internal compiler error was generated by the handling
of the array constructor in the parameter definition.

This fix from the NVIDIA compiler addresses the issue.

Added two tests as well, from the original issue, and
a cut-down version used internally to identify the fix.

### DIFF
--- a/tools/flang1/flang1exe/semant.c
+++ b/tools/flang1/flang1exe/semant.c
@@ -13286,7 +13286,7 @@ fixup_param_vars(SST *var, SST *init)
     sptr1 = get_param_alias_var(sptr, dtype);
   } else if (DTY(dtype) == TY_ARRAY) {
     ad = AD_DPTR(dtype);
-    if (AD_ASSUMSZ(ad) || AD_ADJARR(ad) || AD_DEFER(ad)) {
+    if (AD_ADJARR(ad) || AD_DEFER(ad)) {
       error(84, 3, gbl.lineno, SYMNAME(sptr),
             "- a named constant array must have constant extents");
       return;
--- a/test/f90_correct/inc/issue670_00.mk
+++ b/test/f90_correct/inc/issue670_00.mk
@@ -1,0 +1,28 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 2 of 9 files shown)


tools/flang1/flang1exe/semant.c
```diff
@@ -13286,7 +13286,7 @@ fixup_param_vars(SST *var, SST *init)
     sptr1 = get_param_alias_var(sptr, dtype);
   } else if (DTY(dtype) == TY_ARRAY) {
     ad = AD_DPTR(dtype);
-    if (AD_ASSUMSZ(ad) || AD_ADJARR(ad) || AD_DEFER(ad)) {
+    if (AD_ADJARR(ad) || AD_DEFER(ad)) {
```
kiranchandramohan, line 13289: Flang would have given an error here for the following code previously, with this change the error does not appear.

```
program test
  integer, parameter :: asumed_sz(*) = 1
  print *, assumed_sz(1)
end program
```

kiranchandramohan (commented):
LGTM. Passed a test suite internally.

Have a question about a negative test. See inline comment.

tools/flang1/flang1exe/semant.c
bryanpkc: If the initializer in your example was changed to `(/1/)`, the shape should be implied to be `DIMENSION(1)`, but Flang currently would still complain incorrectly. gfortran can compile that case successfully.

Setting an assumed-shape array to a scalar integer has unclear semantics, and Flang should complain about this case.

It seems to me that the code should differentiate these cases and produce errors (or not) accordingly.

gklimowicz:
There's some kind of interaction with this fix in Flang that doesn't appear in the NVIDIA compiler. I need to dig further to see if there are other changes that are also required to me merged. (I didn't see anything when I first looked, but something here can't be explained.)

bryanpkc:
I have checked with the latest master branch, and the problem is no longer occurring, thanks to a fix that landed as 97527abd723a85154e54337b5a1f3e1e5db1f86d. Both test cases in this PR run correctly, and the compiler behaves correctly [in the case mentioned by Kiran](https://github.com/flang-compiler/flang/pull/919#discussion_r479819924).

bryanpkc:
Closing stale PR.

PR #921, opened 2020-09-08
Correction of representation of string (character type) constants.

alokkrsharma:
The current representation fails when character constant has '\0' in between.
Current representation terminates string there which is not correct.
It is fixed by representing character type with I8 type array.

Note: This issue was exposed after support of debug info for parameter
attribute. Below test case would work only after parameter attribute.
After #888 commit FUTURE-CHECK are changed to CHECK in test case.

test case
```
program main
  character(10),parameter :: x = "abc"//achar(0)//"def"
  print *,x
end
```
Current IR
```
call void @llvm.dbg.value (metadata [10 x i8] c"abc\00", metadata !17, metadata !18)
```
Fixed IR
```
call void @llvm.dbg.value (metadata [10 x i8] c"abc\00def   ", metadata !17, metadata !18)
```

### DIFF
--- a/test/debug_info/conststring.f90
+++ b/test/debug_info/conststring.f90
@@ -1,0 +1,23 @@
+!! This should be enabled only after #888
+!! check if conststrings are stored correctly for special characters
+!RUN: %flang -g -S -emit-llvm %s -o - | FileCheck %s
+
+!! \0 = 0 , " = 34 = 0x22
+!FUTURECHECK: call void @llvm.dbg.value(metadata [10 x i8] c"a\00d\22g     ", metadata [[CONSTR1:![0-9a-f]+]], metadata !DIExpression())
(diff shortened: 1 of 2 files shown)


jiel-nv (approved):

alokkrsharma:
Hi @jiel-nv , thanks for approving this.

gklimowicz (approved):
I reviewed this PR combined with #863 and #888 and Flang-driver PR#92 and PR#93. All tests passed on OpenPOWER as expected.

kiranchandramohan (approved):
LGTM.

shivaramaarao:
@bryanpkc , Is this patch set OK to merge?
863, 921, 92,93 and 888

shivaramaarao:
@bryanpkc : Is it ok to merge this code.

bryanpkc:
@shivaramaarao I didn't get a chance to test this patch, but it looks good to merge. Sorry for the delays.

Issue #889, opened 2020-05-13
Problem with rename-list on USE statement


RichBarton-Arm:
The following reproducer is reduced from the [BigDFT application](http://bigdft.org/Wiki/index.php?title=BigDFT_website). Flang is not able to compile this and gives an error on line 10:

>flang -O0 big_dft1.f90     
F90-S-0087-Non-constant expression where constant expression required (big_dft1.f90: 50)
F90-S-0081-Illegal selector - KIND parameter has unknown value for data type  (big_dft1.f90: 50)
  0 inform,   0 warnings,   2 severes, 0 fatal for potential_from_charge_multipoles

In the reproducer, we have an integer type _dp_ used as the kind to initialise a real _x_ in the function _potential_from_charge_multipoles_. _dp_ is defined in _module_defs_ and is also defined through a complicated sequence of module use statements from both _f_harmonics_ and _Poisson_Solver_ but neither of these definitions should be visible at the definition of _x_ and the definition from _module_defs_ should be used. This is the gfortran behaviour, so x is a sp real and solid_harmonic is a dp real.

The reproducer is very fragile and removing a number of things will cause it to build with flang and create a program that prints the expected output. I have tried to point all these out in comments

```
module f_precisions
  implicit none
  integer, parameter, public :: f_double = 8
end module f_precisions

module Poisson_Solver
   use f_precisions
   implicit none
   integer, parameter, public :: dp=f_double
end module Poisson_Solver

module module_types
  ! Removing this use statement and defining dp in this module fixes our issue
  use Poisson_Solver
  implicit none
  type, public :: DFT_local_fields
  end type DFT_local_fields
end module module_types

module f_harmonics
  use f_precisions, only: dp => f_double
  implicit none
  private ! Removing this fixes the issue
  public :: solid_harmonic
  contains
    pure real(dp) function solid_harmonic()
        solid_harmonic = x'eeeeeeee'
    end function solid_harmonic
end module f_harmonics

module module_defs
  implicit none
  private
  integer, parameter :: f_double = 4
  integer, parameter, public :: dp=f_double
end module module_defs

module multipole
  use module_defs ! defines dp=4
  ! Removing this use statement (and use of solid_harmonic) fixes the issue
  use f_harmonics, only: solid_harmonic ! returns dp=8 defined in f_precisions
  contains
    subroutine potential_from_charge_multipoles(denspot)
      ! Removing this use statement (and use of denspot) fixes the issue
      use module_types, only: DFT_local_fields
[...]

PR #922, opened 2020-09-08
Fix private flag overwrite in find_def_in_most_recent_scope()

### LINKED ISSUE #890: Problem with conflicting use statements
This is similar to issue #889 and from the same [BigDFT application](http://bigdft.org/Wiki/index.php?title=BigDFT_website). 

The below code causes flang to emit the following message:
> F90-S-0155-gp is use-associated from modules box and module_defs, and cannot be accessed (big_dft2.f90: 36)

I think that the definition in _module_defs_ should be visible because the definition in _box_ is private to the module and the definition from _Poisson_Solver_ is renamed in the _USE_ statement.

```
module f_precisions
  implicit none
  integer, parameter, public :: f_double = selected_real_kind(15, 307)
end module f_precisions

module box
  use f_precisions, gp=>f_double
  implicit none
  private
end module box

module Poisson_Solver
  use f_precisions
  implicit none
  integer, parameter, public :: gp=f_double
end module Poisson_Solver

module module_defs
  use f_precisions
  implicit none  
  integer, parameter, public :: gp=f_double
end module module_defs

module module_base
  use module_defs
  implicit none  
end module module_base

subroutine PSolver(eh)
  use module_base ! This has a public gp
  ! Removing the next line makes gp visible
  use box ! This has a private gp
  ! Removing the next line makes gp visible
  use Poisson_Solver, except_gp => gp ! This has a public gp
  implicit none
  real(gp), intent(out) :: eh ! Should be able to use gp from module_base
END SUBROUTINE PSolver
```
The reproducer is slightly unstable in that removing either of _box_ or _Poisson_Solver_ _USE_ statements causes flang to compile this with no error. Removing the _USE_ of _module_base_ does correctly cause an error because _gp_ is unknown.

michalpasztamobica:
Fixes #890 .

`find_def_in_most_recent_scope()` sets the private bit to 0 when it found an alias. I assume that this is unintended behaviour, as the function is just supposed to find the symbol and not modify it. The private bit for an alias would be misleading, as explained in the comment a few lines later, so it gets cleared, but I don't think it should be cleared forever...

I suggest we save the original value of the flag and restore it at the end of the loop.

@kiranchandramohan , @RichBarton-Arm , please review. I am not sure I got the purpose of the `find_def_in_most_recent_scope()` right, but if I did we could consider doing the exact same thing for "hidden" flag, which gets overwritten in the next line.

### DIFF
--- a/test/f90_correct/inc/use00.mk
+++ b/test/f90_correct/inc/use00.mk
@@ -1,0 +1,20 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+########## Make rule for test use00  ########
+
+build:
+	@echo ------------------------------------ building test $(TEST)
+	-$(CC) -c $(CFLAGS) $(SRC)/check.c -o check.$(OBJX)
+	-$(FC) -c $(FFLAGS) $(LDFLAGS) $(SRC)/$(TEST).f90 -o $(TEST).$(OBJX)
+	-$(FC) $(FFLAGS) $(LDFLAGS) $(TEST).$(OBJX) check.$(OBJX) $(LIBS) -o $(TEST).$(EXESUFFIX)
+
+run:
+	@echo ------------------------------------ executing test $(TEST)
+	./$(TEST).$(EXESUFFIX)
+
+verify: ;
+
--- a/test/f90_correct/src/use00.f90
+++ b/test/f90_correct/src/use00.f90
@@ -1,0 +1,39 @@
+module f_precisions
+  integer, parameter, public :: f_double = selected_real_kind(15, 307)
+end module f_precisions
+
+module box
+  use f_precisions, gp=>f_double
+  private
+  public :: get_gp
+contains
+  subroutine get_gp(out_val)
+    integer, intent(out) :: out_val
+    out_val = gp
+  end subroutine get_gp
+end module box
+
+module Poisson_Solver
+  use f_precisions
+  integer, parameter, public :: gp=f_double
+end module Poisson_Solver
+
+module module_defs
+  use f_precisions
+  integer, parameter, public :: gp=f_double
+end module module_defs
+
+subroutine PSolver(eh)
+  use module_defs ! This has a public gp
+  use box ! This has a private gp
+  use Poisson_Solver, except_gp => gp ! This has a public gp
+  real(gp), intent(out) :: eh ! Should be able to use gp from module_base
+end subroutine PSolver
+
+program main
+  use box
+  use Poisson_Solver
+  integer :: box_gp
+  call get_gp(box_gp)
+  call check(box_gp, gp, 1) ! box's gp and Solver's gp alias the same variable
+end program
--- a/test/f90_correct/lit/use00.sh
+++ b/test/f90_correct/lit/use00.sh
@@ -1,0 +1,9 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Shared lit script for each tests. Run bash commands that run tests with make.
+
+# RUN: KEEP_FILES=%keep FLAGS=%flags TEST_SRC=%s MAKE_FILE_DIR=%S/.. bash %S/runmake | tee %t 
+# RUN: cat %t | FileCheck %S/runmake
--- a/tools/flang1/flang1exe/module.c
+++ b/tools/flang1/flang1exe/module.c
@@ -503,7 +503,6 @@ find_def_in_most_recent_scope(int sptr, int save_sem_scope_level)
     if (NMPTRG(sptr1) != NMPTRG(sptr))
       continue;
     if (STYPEG(sptr1) == ST_ALIAS && aliased_sym_visible(sptr1)) {
-      PRIVATEP(sptr1, 0);
       HIDDENP(SYMLKG(sptr1), 0);
     }
     if (STYPEG(sptr1) == ST_ALIAS) {


michalpasztamobica:
I have run the tests with lit and got:
```
Testing Time: 232.50s
  Expected Passes    : 1261
  Unsupported Tests  : 33
```
But to be honest, I think the existing tests do not cover the scenario as they passed the same way when I just removed the flag clearing, most likely breaking `find_def_in_most_recent_scope()` function in at least some cases...
I will try to add a test case for this scenario.

gklimowicz:
@mleair and @wangzpgi may want to take a look at this after there are tests added. (Mark, also let me know if there are recent NVIDIA Fortran fixes that might address this issue. I know there are a lot of NVIDIA bug fixes and enhancements that we haven't pulled into Flang.)

michalpasztamobica:
Added `use00` test, based on the example described in #890 . 
Without the fix the test fails to build as described in the issue:
```
F90-S-0155-gp is use-associated from modules box and module_defs, and cannot be accessed (PATH/test/f90_correct/lit/../src/use00.f90: 24)
  0 inform,   0 warnings,   1 severes, 0 fatal for psolver
PATH/test/f90_correct/lit/../inc/use00.mk:12: recipe for target 'build' failed
make: *** [build] Error 1
------------------------------------ executing test use00
...
Failing Tests (1):
    Flang :: f90_correct/lit/use00.sh

  Expected Passes    : 1261
  Unsupported Tests  : 33
  Unexpected Failures: 1
```

With the fix I got one pass more:
```
Testing Time: 195.95s
  Expected Passes    : 1262
  Unsupported Tests  : 33
```

test/f90_correct/inc/use00.mk
```diff
@@ -0,0 +1,20 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+########## Make rule for test use00  ########
+
+FFLAGS+= -Mallocatable=03
```
RichBarton-Arm, line 9: Why is this needed for the test?

test/f90_correct/src/use00.f90
```diff
@@ -0,0 +1,29 @@
...
+module module_defs
+  use f_precisions
+  integer, parameter, public :: gp=f_double
+end module module_defs
+
+subroutine PSolver(eh)
+  use module_defs ! This has a public gp
+  use box ! This has a private gp
+  use Poisson_Solver, except_gp => gp ! This has a public gp
+  real(gp), intent(out) :: eh ! Should be able to use gp from module_base
```
RichBarton-Arm, line 24: Given that these are runnable tests, it would be nice to make this a test that actually runs a program which passes iff the Fortran standard is followed in this case. Is it possible to make one of these for the bug that we have fixed? I am not sure it is because I think it might only occur if the private `gp` in `box` and the public `gp` in `Poisson_Solver` alias the same underlying variable. Perhaps a statement that prints out `eh` in `PSolver` to show it is the right precision might work.

In addition, my original bug report contains a second code snippet which is a permutation of the same original code but showing a slightly different bug. This bug is reproducable with a running program as it is the compiler incorrectly selecting the underlying alias of `gp` to use and can be seen in a print statement. A runnable testcase which prints output and checks it against an expected value should be possible for that bug. 

Also - does your patch fix that related bug?

RichBarton-Arm (changes requested):

test/f90_correct/src/use00.f90
michalpasztamobica: I tried to cut the example as short as possible and still keep it failing without the fix and passing with the fix and this is what I came down to. The bug is revealed at compilation time, so I didn't think any particular checks are required, but now that you pointed out the depth of the problem I'll try to add a check for `eh` precision being the same as `gp`'s.

The permuted example failed to compile for me just like the non-permuted one, so I didn't take care of it any more and I must admit I forgot about it. I just checked now that it also compiles fine once the fix is there.

michalpasztamobica:
I force pushed changes reuested by @RichBarton-Arm :
* removed unnecessary compilation flags
* modified the test to perform a check between the `box`'es private `gp` variable and the public `gp` from `Poisson_solver`.

test/f90_correct/inc/use00.mk
michalpasztamobica: Sorry, copy & paste from another test. Removed now :)

test/f90_correct/src/use00.f90
michalpasztamobica: I updated the test case code adding a public function allowing us to check the value of `box`'es private `gp` (it's private, so no direct way to access it) and comparing its return value to the publicly available `gp`. The test is passing. I hope I got it right, @RichBarton-Arm ?

Have you managed to reproduce the other issue that you mentioned: different incorrect behaviour with slightly permuted test?

mleair:
The proposed solution looks hacky and confusing. As @michalpasztamobica pointed out in #890, the bug occurs with the setting of PRIVATEP(sptr1, 0) at line 507. I believe that's a bug. I think a better fix is to remove that line unless someone has a test that fails with it not being there. I reproduced the bug with nvfortran. I then ran some tests on nvfortran with that change and did not see any failures. Therefore, I believe the correct solution is to remove line 507.

wangzpgi:
I agree with @mleair . That line was put in 5+ years ago, might already be outdated. Since there is no regression found, I assume it's safe to just remove that line.

michalpasztamobica:
Thanks for your remarks, @mleair and @wangzpgi .
I am concerned about the rest of the `find_def_in_most_recent_scope()` function. It relies on the PRIVATEG macro later on in [L510](https://github.com/flang-compiler/flang/blob/5755e102155f81bbef525390cab84f655da396a7/tools/flang1/flang1exe/module.c#L510). 

I removed the line, as you suggested, but I also added a local variable that stores its value, so the `find_def_in_most_recent_scope()` works the same as it used to.

There also `PRIVATEP` being called in [L548](https://github.com/flang-compiler/flang/blob/5755e102155f81bbef525390cab84f655da396a7/tools/flang1/flang1exe/module.c#L548), but the `ng` should point to the actual variable, not the aliased one, so it should have its own PRIVATE setting anyway.

Tests are still passing fine.

mleair:
I don't understand why you need the local variable sptr1_prv. I don't think you need it. The PRIVATE field does not change for sptr1 in that loop. I confirmed that with your test case by turning PRIVATEP into a function instead of a macro (this allows you to add a "debug" print statement in the PRIVATEP function or set a break point in PRIVATEP with a debugger). PRIVATEP is not called when that loop is executed. Therefore, you can just remove line 507 (the PRIVATEP(sptr1, 0)), and rely on the PRIVATEG "getter" macro at line 510.

michalpasztamobica:
The private flag is being overwritten with `PRIVATEP(sptr1, 0)`. It doesn't change if it was already `0`, but if it was `1` it would change and influence [the behaviour of line 510](https://github.com/flang-compiler/flang/blob/5755e102155f81bbef525390cab84f655da396a7/tools/flang1/flang1exe/module.c#L510), where PRIVATEG reads out the private flag we've just switched from 1 to 0.

I believe that it actually happens that the flag gets overwritten from 1 to 0 in the example code from issue #890 . I checked this with the debugger attached:
image
Here I am debugging a simple `flang PATH/use00.f90` command, I commented out the `call check` for simplicity.
You can see the `sptr1_prv` is set to `1` when we are just about to set it to `0` (this is what was previously done with a `PRIVATEP(sptr1, 0)` call).

mleair:
We both agree that the bug is the setting of PRIVATEP(sptr1, 0) at line 507. So, that would suggest that the loop behavior was buggy too. What you’re proposing removes the setting of PRIVATEP(sptr1, 0), which we both agree is correct, but your use of the sptr1_prv temporary variable preserves the buggy behavior of the loop. Removing just PRIVATEP(sptr1, 0) did not cause any of our tests to fail (I ran over 10,000 tests and did not see any failures).  Do you have an example that fails if one just removes PRIVATEP(sptr1, 0) at line 507? If not, the fix should just be removing line 507.

michalpasztamobica:
You are right, I do not have a test case that would fail if we just remove the line. I might be being too cautious here :). I now submitted the code compliant with your suggestion and it passes all the tests including the new one. 

Just to fully explain my standpoint: I was trying to apply a "minimum impact fix", leaving the logic of the function intact and just preventing it from a lasting change of the flag, which clearly caused an issue in a totally different place in the code. You do have a point, that we should trust the tests and remove any code that keeps the tested functionality working. 

Once I get more experience with the project I hope to come up with more accurate fixes. Thanks for being patient with me! :smiley:

mleair (approved):

mleair:
Thank you for your fix.

wangzpgi (approved):

RichBarton-Arm (approved):
My changes have been incorporated so happy to approve from my end.
I also cannot reproduce the second 'bug' I reported, so I'll edit the description of 890 and no further action needed here.

kiranchandramohan:
Thanks @michalpasztamobica for the fix and thanks @mleair and @wangzpgi for the reviews.

https://github.com/flang-compiler/flang/issues/889 is a similar issue. Does this patch fix that issue? Should we check and confirm that 889 is a separate issue before merging this one?

michalpasztamobica:
Hi @kiranchandramohan . I had similar hopes and checked, but unfortunately #889 is still reproducible. I will take it up next.

kiranchandramohan:
@bryanpkc @shivaramaarao @gklimowicz This is next from Arm. The patch is already approved by engineers from Nvidia.

kiranchandramohan:
ping @bryanpkc @shivaramaarao @gklimowicz.

gklimowicz (approved):
These pass on OpenPower for LLVM 9 and LLVM 10.

test/f90_correct/src/use00.f90
```diff
@@ -0,0 +1,39 @@
...
+END SUBROUTINE PSolver
```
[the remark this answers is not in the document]
michalpasztamobica: Fixed. Thanks, @bryanpkc !

test/f90_correct/src/use00.f90
```diff
@@ -0,0 +1,39 @@
...
+
+program main
+  use box
+  use Poisson_Solver
+  integer :: box_gp
+  call get_gp(box_gp)
+  call check(box_gp, gp, 1) ! box's gp and Solver's gp alias the same variable
+end
```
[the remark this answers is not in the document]
michalpasztamobica: Fixed. Thanks, @bryanpkc !

bryanpkc (approved):
This PR passed tests on my end. Thanks.

shivaramaarao (approved):

kiranchandramohan:
@michalpasztamobica can you squash?

michalpasztamobica:
@kiranchandramohan . Sure, squashed.

PR #925, opened 2020-09-18
[DebugInfo] Flang should generate correct IR for assumed size array

alokkrsharma:
This is to support debugging of assumed size array. 
It needs below LLVM patch.
https://reviews.llvm.org/D87500 ([DebugInfo] DISubrange support for fortran assumed size array)

Dependency: #901, #902, #913


alokkrsharma:

PR #930, opened 2020-09-30
Fix HTML docs generation

michalpasztamobica:
* The `ast.n` file was moved to a different location causing cmake to fail to generate most of the existing documentation.
* Additionally, the unused rst files are now added to the list, so they are now also visible in the generated HTML.
* Sphinx build depends on generating all necessary rst files.
* Directories must be created before putting files into them.
* `gram.lis` file dependency added (thanks @bryanpkc!).
* renamed Cmake target `gen_sphinx_docs` to `flang2_gen_sphinx_docs` (to match existing `flang1_gen_sphinx_docs`)

@RichBarton-Arm , @kiranchandramohan , please review.

### DIFF
--- a/tools/flang1/utils/n2rst/CMakeLists.txt
+++ b/tools/flang1/utils/n2rst/CMakeLists.txt
@@ -20,6 +20,7 @@ file(MAKE_DIRECTORY ${FLANG1_DOC_BIN_DIR})
 
 add_custom_command(
   OUTPUT ${FLANG1_DOC_BIN_DIR}/ast.rst
+         ${FLANG1_DOC_BIN_DIR}/commopt.rst
          ${FLANG1_DOC_BIN_DIR}/comms.rst
          ${FLANG1_DOC_BIN_DIR}/controller.rst
          ${FLANG1_DOC_BIN_DIR}/dinit.rst
@@ -32,7 +33,8 @@ add_custom_command(
          ${FLANG1_DOC_BIN_DIR}/scanner.rst
          ${FLANG1_DOC_BIN_DIR}/semant.rst
          ${FLANG1_DOC_BIN_DIR}/transform.rst
-  COMMAND ${CMAKE_BINARY_DIR}/bin/fen2rst -v ${FLANG1_DOC_SRC_DIR}/ast.n
+  COMMAND ${CMAKE_BINARY_DIR}/bin/fen2rst -v ${UTILS_AST_DIR}/ast.n
+                                             ${FLANG1_DOC_SRC_DIR}/commopt.n
                                              ${FLANG1_DOC_SRC_DIR}/comms.n
                                              ${FLANG1_DOC_SRC_DIR}/controller.n
                                              ${FLANG1_DOC_SRC_DIR}/dinit.n
@@ -46,7 +48,8 @@ add_custom_command(
                                              ${FLANG1_DOC_SRC_DIR}/semant.n
                                              ${FLANG1_DOC_SRC_DIR}/transform.n
   WORKING_DIRECTORY ${FLANG1_DOC_BIN_DIR}
-  DEPENDS fen2rst ${FLANG1_DOC_SRC_DIR}/ast.n
+  DEPENDS fen2rst ${UTILS_AST_DIR}/ast.n
+                  ${FLANG1_DOC_SRC_DIR}/commopt.n
                   ${FLANG1_DOC_SRC_DIR}/comms.n
                   ${FLANG1_DOC_SRC_DIR}/controller.n
                   ${FLANG1_DOC_SRC_DIR}/dinit.n
@@ -63,6 +66,7 @@ add_custom_command(
 
 add_custom_target(flang1_gen_sphinx_docs
   SOURCES ${FLANG1_DOC_BIN_DIR}/ast.rst
+          ${FLANG1_DOC_BIN_DIR}/commopt.rst
           ${FLANG1_DOC_BIN_DIR}/comms.rst
           ${FLANG1_DOC_BIN_DIR}/controller.rst
           ${FLANG1_DOC_BIN_DIR}/dinit.rst
@@ -76,6 +80,8 @@ add_custom_target(flang1_gen_sphinx_docs
           ${FLANG1_DOC_BIN_DIR}/semant.rst
           ${FLANG1_DOC_BIN_DIR}/transform.rst
   )
+
+add_dependencies(flang1_gen_sphinx_docs gen_frontend_parsetable)
 
 # Local Variables:
 # mode: cmake
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -141,6 +141,16 @@ if (SPHINX_FOUND)
         "${SPHINX_BASE_DIR}"
         "${SPHINX_HTML_DIR}"
     COMMENT "Building HTML documentation with Sphinx")
+
+   add_dependencies(sphinx-flang
+                    flang1_gen_sphinx_docs
+                    gen_frontend_machar
+                    gen_frontend_parsetable
+                    gen_frontend_symtab
+                    gen_frontend_symini
+                    flang2_gen_sphinx_docs
+                    gen_backend_symtab
+                    gen_backend_symini)
 endif()
 
 if (LLVM_ENABLE_SPHINX)
--- a/docs/doxygen.cfg.in
+++ b/docs/doxygen.cfg.in
@@ -687,7 +687,7 @@ CITE_BIB_FILES         =
 # messages are off.
 # The default value is: NO.
 
-QUIET                  = NO
+QUIET                  = YES
 
 # The WARNINGS tag can be used to turn on/off the warning messages that are
 # generated to standard error ( stderr) by doxygen. If WARNINGS is set to YES
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -44,6 +44,8 @@ Front-end Design
    :maxdepth: 1
 
    flang1/intro
+   flang1/commopt
+   flang1/comms
    flang1/controller
    flang1/scanner
    flang1/parser
@@ -88,6 +90,7 @@ Back-end Internals
    ILI Definitions <flang2/ilitp>
    Intrinsics & Generics <flang2/symini>
    flang2/errmsg
+   flang2/register
 
 Source code
 -----------
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -34,8 +34,9 @@ Common
    flang2/coding
    Target Machine <flang1/machar>
    flang2/error
-   flang1/dinit
    flang2/fin
+   flang2/xflag
+   flang2/xref
 
 Front-end Design
 ----------------
@@ -63,6 +64,7 @@ Front-end Internals
 
    Grammar <flang1/grammar>
    flang1/ast
+   flang1/dinit
    flang1/symtab
    Intrinsics & Generics <flang1/symini>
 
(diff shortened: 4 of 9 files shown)


kiranchandramohan:
@gklimowicz, @bryanpkc, @shivaramaarao : Please help with review/suggestions.

Note: In general we are interested in improving the documentation of Classic Flang upstream. We will try to share what we have.

tools/flang1/utils/n2rst/CMakeLists.txt
```diff
@@ -46,7 +48,8 @@ add_custom_command(
                                              ${FLANG1_DOC_SRC_DIR}/semant.n
                                              ${FLANG1_DOC_SRC_DIR}/transform.n
   WORKING_DIRECTORY ${FLANG1_DOC_BIN_DIR}
-  DEPENDS fen2rst ${FLANG1_DOC_SRC_DIR}/ast.n
+  DEPENDS fen2rst ${UTILS_AST_DIR}/ast.n
+                  ${FLANG1_DOC_SRC_DIR}/commopt.n
```
bryanpkc, line 52: My build is failing with this error message:
```
[ 14%] Built target fen2rst
make[2]: *** No rule to make target 'tools/flang1/utils/prstab/gram.lis', needed by 'docs/sphinx/flang1/ast.rst'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:1806: tools/flang1/utils/n2rst/CMakeFiles/flang1_gen_sphinx_docs.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

bryanpkc (changes requested):
I needed to apply this patch before I could configure flang with the additional `-DLLVM_INCLUDE_DOCS=on -DLLVM_BUILD_DOCS=on -DLLVM_ENABLE_DOXYGEN=on` options:

```diff
diff --git a/docs/CMakeLists.txt b/docs/CMakeLists.txt
index ec2bfd84..e1a4ef3e 100644
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -9,6 +9,8 @@ find_package(Sphinx)

 if (DOXYGEN_FOUND)
 if (LLVM_ENABLE_DOXYGEN)
+  add_custom_target(doxygen ALL)
+
   set(abs_srcdir ${CMAKE_CURRENT_SOURCE_DIR})
   set(abs_builddir ${CMAKE_CURRENT_BINARY_DIR})

```

tools/flang1/utils/n2rst/CMakeLists.txt
bryanpkc: I fixed the above error with this patch:
```diff
diff --git a/tools/flang1/utils/prstab/CMakeLists.txt b/tools/flang1/utils/prstab/CMakeLists.txt
index 59a81011..c9837a51 100644
--- a/tools/flang1/utils/prstab/CMakeLists.txt
+++ b/tools/flang1/utils/prstab/CMakeLists.txt
@@ -21,7 +21,7 @@ add_executable(prodstr
 # Generate parse tables

 add_custom_command(
-  OUTPUT ${UTILS_PARSETABLE_BIN_DIR}/gramdf.h ${UTILS_PARSETABLE_BIN_DIR}/gramsm.h ${UTILS_PARSETABLE_BIN_DIR}/gramtk.h ${UTILS_PARSETABLE_BIN_DIR}/proddf.h ${UTILS_PARSETABLE_BIN_DIR}/tokdf.h ${UTILS_PARSETABLE_BIN_DIR}/gramnt.h
+  OUTPUT ${UTILS_PARSETABLE_BIN_DIR}/gramdf.h ${UTILS_PARSETABLE_BIN_DIR}/gramsm.h ${UTILS_PARSETABLE_BIN_DIR}/gramtk.h ${UTILS_PARSETABLE_BIN_DIR}/proddf.h ${UTILS_PARSETABLE_BIN_DIR}/tokdf.h ${UTILS_PARSETABLE_BIN_DIR}/gramnt.h ${UTILS_PARSETABLE_BIN_DIR}/gram.lis
   # Copy to current build directory, otherwise header will be generated in
   # original directory
   COMMAND ${CMAKE_COMMAND} -E copy ${UTILS_PARSETABLE_DIR}/gram.txt ${UTILS_PARSETABLE_BIN_DIR}
```

michalpasztamobica: Thanks for pointing this out @bryanpkc. Indeed, I managed to reproduce the issue when I `rm -rf`-ed the build directory and started from scratch.
Unfortunately your patch did not improve the situation.  
It might matter that I build with `-DFLANG_INCLUDE_DOCS=ON`, which is the flag I found to be used in the CMake files. I admit I am currently looking mainly at HTML docs generated by Sphinx and this flag did the job for me. I see no explicit documentation on how docs should be built - perhaps that would be an idea for another improvement? I doubt that the nroff/rst files influence the doxygen build.
When I used the flags you mentioned I got a different error. But then... maybe I should build `llvm` with these flags as well?

Anyway, regarding `gram.lis` - this file doesn't exist anywhere in the source tree and is not produced during build process.  Also - the sphinx docs will eventually use `flang1/grammar.rst` and `gram.rst` would remain unused. `git blame` shows that [most of this file remained intact](https://github.com/flang-compiler/flang/blame/master/tools/flang1/utils/n2rst/CMakeLists.txt) since the beginning of this repo, so perhaps it was never really built?

I suggest we remove the lines which mention `gram.list` and `gram.rst` from the CMakeLists.txt. Would that be OK, @bryanpkc ?

bryanpkc: @michalpasztamobica I wasn't discussing Doxygen at all.

I am building master branch, plus the patch above, with `-DLLVM_INCLUDE_DOCS=on` and `make -j4 install`, after deleting the build directory, and I get the correct Sphinx files. There is currently a race condition in the Sphinx command which requires running `make -j4 install` a second time in order to generate all the files and index.html correctly. I haven't needed to build LLVM any differently (I did not add any docs-related option when running `cmake` for llvm-project), but I should add that I am using release_100 classic-flang-llvm-project.

`gram.lis` is still being generated by the `lr` command in my build. You can find it under `$CMAKE_BINARY_DIR/tools/flang1/utils/prstab/`; it contains the actual text for the grammar, and Sphinx generates `gram.html` from it. In contrast, `tools/flang1/docs/grammar.n`contains just a pointer to `gram.html`. Do you have a different `grammar.n` than the master branch?

P.S. I agree we should update the "Building Flang" wiki page with a sentence or two on how to build the docs.

michalpasztamobica: It took me a while, but now I can reliably build all the docs in a single run even with `make -j6`.

You were right, @bryanpkc : `gram.html` is being generated and used to produce `grammar.html`. I included your patch. Thanks!

Regarding the two DOCS flags and the wiki update... I am not entirely sure but to my understanding  `LLVM_INCLUDE_DOCS` [is being set if](https://github.com/flang-compiler/flang/blob/a6d7701ea4f4ea37b09ac83d8351623d0292c489/CMakeLists.txt#L417) `FLANG_INCLUDE_DOCS` is set. The latter is also a valid CMake option. I don't have the privileges to edit [wiki with build instructions](https://github.com/flang-compiler/flang/wiki/Building-Flang), but perhaps you or @kiranchandramohan could put up a sentence saying "To build the the HTML and Doxygen documentation add `-DFLANG_INCLUDE_DOCS` to the cmake command" somewhere in point 7 of the wiki page.

bryanpkc: Thanks @michalpasztamobica!

> LLVM_INCLUDE_DOCS is being set if FLANG_INCLUDE_DOCS is set.

It's the other way around... The default value for `FLANG_INCLUDE_DOCS` is the value of `LLVM_INCLUDE_DOCS`. Setting either will work. Setting the latter would be useful if we were building the Flang project in-tree (but we are not).

michalpasztamobica: You're right @bryanpkc , I wrote it the wrong way round, thanks for correction.

bryanpkc: I have updated the [Building-Flang](https://github.com/flang-compiler/flang/wiki/Building-Flang) page.

bryanpkc (approved):
LGTM now. Thanks @michalpasztamobica!

bryanpkc:
@michalpasztamobica I noticed that xflag.n and xref.n are not in the documentation index, and that there are two versions of dinit.n. Would you consider merging the following patch into the PR?

https://github.com/Huawei-PTLab/flang/commit/9c42e057ca54eaa5f4c77e8ad1c8a24af28caf95

michalpasztamobica:
Nice catch @bryanpkc , I cherry-picked and pushed your patch to this PR.

gklimowicz (approved):
Works fine on OpenPower LLVM 9.0 and 10.0.

kiranchandramohan:
Thanks @michalpasztamobica for this patch. Do you know why 

1) there are some stray lines at the top of the ILM Definitions HTML file. e.g. given below.
```
.nr IN 0 1
.de IL
.nr IN +1
.sp
.in 0
.ne 4
.nf
.ta 0.5i 4.5i
\n(IN. \$1 \$3 \$4 \$5 \$6 \$7 \$8 \$9t Type: \$2
.in 0.5i
.ad ..
```
2) some entries in ILM/ILI Definitions do not have entry number and are not formatted properly. For e.g. consider the following, the entry for RISNAN.
```
1. BOS stc1 stc2 stc3 Type: misc

Always (and only) appears as the first ILM of an ILM block. (roughly corresponds to the beginning of a source statement.)

stc1 - source statement line number (negated for insert files?).
If more than one ILM block is written for a given
statement, all but the first have 0 specified for the
line number.
For typical statements, this is actually the line
number of the terminating ‘;’.
stc2 - current source file index. Default value is 1.
stc3 - number of words for this ILM block (including the BOS).
.AT spec trm
.IL RISNAN arth lnk
.AT spec trm
...
```

michalpasztamobica:
Hi @kiranchandramohan 

Ad 1)
According to the [nroff quick reference](http://www.quut.com/berlin/ms/numbers.html) I found these are "number registers":

> autoincrement and -decrement
>.nr name value auto-increment
>The auto-increment is added (or subtracted) to the register on every access where a + (or -) follows the n, as in
>\n+Y or \n-(XX.

Then `.de` [defines a macro](http://www.quut.com/berlin/ms/troff-def.html).
Then inside the macro there's more nroff programming.
Anyway these are not "stray lines", they bring some value to the file, although I admit I don't fully comprehend what the value exactly is.

Ad 2) 
I assume you are [asking about the source nroff](https://github.com/flang-compiler/flang/blob/a6d7701ea4f4ea37b09ac83d8351623d0292c489/tools/flang2/utils/ilmtp/x86_64/ilmtp.n#L39) `.n` files? The intermediate `.rst` files are generated based on nroff using the `n2rst` tool and the end product `.html` files get this formatting from `.rst`.
By looking at the file as a whole I infer that the reason is that the file was written with an 80-character limitation rule in mind. This at least applies to all paragraphs. There are longer lines when enumerations or lists come into play, but whenever I see a paragraph the lines are always below 80 characters.

This might not make much sense and it does make the generated html look a bit ugly. I think this could be taken care of by some clever `sed` script. However - I would strongly suggest that we make a separate PR for this, as all other changes will be lost in an ocean of line breaks removals.

kiranchandramohan:
@shivaramaarao is the patch OK with you?

shivaramaarao (approved):
looks good.

kiranchandramohan:
@michalpasztamobica please squash.

michalpasztamobica:
@kiranchandramohan , I squashed, just leaving @bryanpkc 's commit separate to keep the authorship in order. I hope this is OK?

Issue #421, opened 2018-03-20
flang2 terminates with segfault


dev-zero:
While there is much point in running `flang2` manually without arguments, it should still not just segfault:

```
$ flang2
Segmentation fault (core dumped)
```

gdb backtrace:
```
(gdb) bt
#0  0x000000000057f344 in parse_arguments ()
#1  0x000000000050e2b9 in init ()
#2  0x000000000050da80 in main ()
```

sscalpone:
% /proj/pgi/flang/x86_64/dev/bin/flang1 < /dev/null
Segmentation fault

Renaud-K:
Created https://github.com/flang-compiler/flang/pull/724 to address this issue

PR #931, opened 2020-09-30
[flang2] Fix segmentation faults (#421)

bryanpkc:
Add null checks to solve several obvious crash bugs.

### DIFF
--- a/tools/flang2/flang2exe/main.cpp
+++ b/tools/flang2/flang2exe/main.cpp
@@ -379,6 +379,7 @@ main(int argc, char *argv[])
   bool findex = false;
   bool need_constructor = false;
   int accel_cnt, accel_vendor = 0;
+  int ilm_units_seen = 0;
 
   get_rutime();
   init(argc, argv);
@@ -428,9 +429,19 @@ main(int argc, char *argv[])
     if (!process_input(argv[0], &need_constructor))
       break;
 
+    // We cannot reuse gbl.func_count here since gbl.func_count must be
+    // incremented before calling process_input().
+    ilm_units_seen++;
+
   } while (!gbl.eof_flag);
 
-  cg_llvm_end();
+  if (ilm_units_seen == 0)
+    errwarn((error_code_t)6);
+
+  // Finalize IR generation only if the output file has been initialized
+  // by cg_llvm_init or assemble.
+  if (FTN_HAS_INIT())
+    cg_llvm_end();
 
   if (flg.smp) {
     ll_unlink_parfiles();
@@ -553,6 +564,9 @@ init(int argc, char *argv[])
                           * issued:
                           */
   errini();
+  if (argc <= 1)
+    errfatal((error_code_t)1);
+
   gbl.curr_file = NULL;
   sym_init();
 
@@ -940,7 +954,7 @@ finish()
   } else {
     if (gbl.objfil != NULL)
       fclose(gbl.objfil);
-    if (!flg.es)
+    if (!flg.es && FTN_HAS_INIT())
       assemble_end();
   }
   if (gbl.asmfil != NULL && gbl.asmfil != stdout)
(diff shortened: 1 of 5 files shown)


kiranchandramohan:
@bryanpkc Can we have some tests as well?

bryanpkc:
@kiranchandramohan I have added some tests.

kiranchandramohan:
@bryanpkc Are these crashes reproducible from Fortran source or only when independently running flang1/flang2?

bryanpkc:
@kiranchandramohan These failures are not normally reproducible from Fortran source. The driver usually ensures that flang2 is invoked with correct command line arguments; however a developer might want to run flang2 directly, and would not expect it to crash when invoked without any argument.

We found the second crash by accident when a file system was full and an empty .ilm file got passed to flang2.

bryanpkc:
flang2 should probably fail in the case of an empty input file, but I just found that Fortran preprocessing (`-E`) expects flang2 to succeed in that case. I'm not really sure why flang2 is required when preprocessing; I can look a bit deeper.

bryanpkc:
main.cpp processes two types of input files, an `.ilm` file and an `.stb` file. Both contains multiple sections that are read in loops, which call `upper` to process each section. The loops depends on the EOF condition to signify a successful termination of the loop; the way the functions are structured, `upper` will always return with an "empty section" condition at the end of an unempty file. I have added counters to let the loops distinguish between an empty section and a really empty input file, and make flang2 produce a warning in the latter case, instead of crashing.

tools/flang2/flang2exe/main.cpp
```diff
@@ -1006,6 +1025,10 @@ process_stb_file()
 
   } while (!gbl.eof_flag);
 
+  if (stb_units_seen == 0) {
```
kiranchandramohan, line 1028: Nit: drop trivial brace.

tools/flang2/flang2exe/main.cpp
```diff
@@ -428,9 +429,18 @@ main(int argc, char *argv[])
...
       break;
 
+    // We cannot reuse gbl.func_count here since gbl.func_count must be
+    // incremented before calling process_input().
+    ilm_units_seen++;
+
   } while (!gbl.eof_flag);
 
-  cg_llvm_end();
+  if (ilm_units_seen == 0) {
```
kiranchandramohan, line 438: Nit: Drop trivial brace.

tools/flang2/flang2exe/main.cpp
```diff
@@ -428,9 +429,18 @@ main(int argc, char *argv[])
...
+    errwarn((error_code_t)6);
+  }
+
+  if (FTN_HAS_INIT())
```
kiranchandramohan, line 442: Nit: A comment here on what FTN_HAS_INIT is and when it is set might be useful.

kiranchandramohan (approved):
LGTM.

> main.cpp processes two types of input files, an `.ilm` file and an `.stb` file. Both contains multiple sections that are read in loops, which call `upper` to process each section.

Now that you understand the flow, can you write this somewhere if you get the time?

gklimowicz (approved):
Works fine on OpenPower LLVM 9.0 and 10.0.

shivaramaarao (approved):
Looks good to me.

kiranchandramohan:
@bryanpkc is this ready to merge?

tools/flang2/flang2exe/main.cpp
bryanpkc: Done.

tools/flang2/flang2exe/main.cpp
bryanpkc: Done.

tools/flang2/flang2exe/main.cpp
bryanpkc: Done.

bryanpkc:
Yes, @kiranchandramohan, I was just addressing your comments just as your reminder came.

PR #742, opened 2019-06-25
Implement UNROLL directive

CarolineConcatto:
Author: Caroline Concatto <caroline.concatto@arm.com>
This patch implements UNROLL pragma.
However in order for it to work it depends on a previous patch #660  
You can use this directive when you want a loop to be unrolled.
 It must be placed immediately before a DO loop.
It works as:
!dir$ unroll

PR #660, opened 2019-01-21
Enable support for simd directives

kiranchandramohan:
This commit enables support for the following simd directives. These
directives can be used to control vectorisation performed by LLVM. It
works by passing vectorisation metadata to LLVM.
Note that these directives do not control any vectorisation performed
by flang itself.
1) !dir$ ivdep (ignore mem-dep and vectorise)
2) !$omp simd (requires -fopenmp, ignore mem-dep and vectorise)
3) !dir$ vector always (ignore cost and vectorise)
4) !dir$ novector

PR #942, opened 2020-10-25
Implement UNROLL(n) directive

bryanpkc:
This PR is based on #742 and cleans it up such that it no longer conflicts with #660 (which is included in this PR). A second commit is added to extend the directive to support user-specified unroll factors, and remove support for the old PGI-style syntax (which cannot be easily supported with `llvm.loop.unroll` metadata AFAICT). The new syntax is similar to what Intel Fortran supports:
```
!dir$ unroll
!dir$ unroll(6)
!dir$ unroll = 6
!dir$ nounroll
```
This PR also clarifies the usage of `XBIT(11)` and updates the test cases to make them more complete and understandable.

This PR needs to be rebased after #660 is accepted.

### DIFF
--- a/tools/shared/pragma.c
+++ b/tools/shared/pragma.c
@@ -1184,37 +1184,27 @@ do_sw(void)
       return true;
     break;
   case SW_UNROLL:
-    /* [no]unroll		-x/y 11 3
-     * [no]unroll = c	-x/y 11 1
-     * [no]unroll = n	-x/y 11 2
-     *     unroll = c:v     -y   11 3  -x    9 v
-     *     unroll = n:v     -y   11 3  -x   10 v
+    /* unroll       -x 11 0x1   -y 11 0x402
+     * unroll = n   -x 11 0x2   -y 11 0x401 -x 9 n
+     * unroll (n)   -x 11 0x2   -y 11 0x401 -x 9 n
+     * nounroll     -x 11 0x400 -y 11 0x3
      */
     typ = gtok();
-    if (typ != T_EQUAL) {
-      if (no_specified)
+    // !dir$ [no]unroll
+    if (typ == T_END) {
+      if (no_specified) { // !dir$ nounroll
         bset(DIR_OFFSET(currdir, x[11]), 0x400);
-      else
-        bset(DIR_OFFSET(currdir, x[11]), 0x3);
-    } else if (gtok() == T_IDENT) {
-      if (strcmp(ctok, "c") == 0)
-        i = 9;
-      else if (strcmp(ctok, "n") == 0)
-        i = 10;
-      else
-        return true;
-      if (no_specified)
-        bset(DIR_OFFSET(currdir, x[11]), i - 8);
-      else if (gtok() != T_COLON)
-        bclr(DIR_OFFSET(currdir, x[11]), i - 8);
-      else if (gtok() != T_INT)
-        return true;
-      else {
-        if (itok <= 0)
-          itok = 1;
-        assn(DIR_OFFSET(currdir, x[i]), (int)itok);
-        bclr(DIR_OFFSET(currdir, x[11]), 3);
+        bclr(DIR_OFFSET(currdir, x[11]), 0x3);
+      } else { // !dir$ unroll
+        bset(DIR_OFFSET(currdir, x[11]), 0x1);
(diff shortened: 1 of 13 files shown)


kiranchandramohan:
Thanks @bryanpkc for this patch.
Does this patch contain the changes you asked for in 660?
If it contains 660 and 742 would it be better to go ahead with this patch and drop 660 and 742?

bryanpkc:
@kiranchandramohan No, it doens't contain the changes I asked for in 660. I can certainly make it so if you don't think my requests were controversial.

kiranchandramohan:
Thanks, @bryanpkc. I have now made the suggested changes in 660.

Will review this patch soon, this looks great on an initial look.

kiranchandramohan:
Have submitted https://github.com/flang-compiler/flang/pull/660. You can rebase now. And then we can review this instead of https://github.com/flang-compiler/flang/pull/742.

gklimowicz (approved):
This PR passes our tests on OpenPOWER for LLVM 9 and LLVM 10.

tools/shared/pragma.c
```diff
@@ -1184,37 +1184,27 @@ do_sw(void)
...
     break;
   case SW_UNROLL:
-    /* [no]unroll		-x/y 11 3
-     * [no]unroll = c	-x/y 11 1
-     * [no]unroll = n	-x/y 11 2
-     *     unroll = c:v     -y   11 3  -x    9 v
-     *     unroll = n:v     -y   11 3  -x   10 v
+    /* unroll       -x 11 0x1   -y 11 0x402
+     * unroll = n   -x 11 0x2   -y 11 0x401 -x 9 n
+     * unroll (n)   -x 11 0x2   -y 11 0x401 -x 9 n
```
kiranchandramohan, line 1189: Is the 401, 402 setting used anywhere?

tools/shared/pragma.c
```diff
@@ -1184,37 +1184,27 @@ do_sw(void)
...
-        if (itok <= 0)
-          itok = 1;
-        assn(DIR_OFFSET(currdir, x[i]), (int)itok);
-        bclr(DIR_OFFSET(currdir, x[11]), 3);
+        bclr(DIR_OFFSET(currdir, x[11]), 0x3);
+      } else { // !dir$ unroll
+        bset(DIR_OFFSET(currdir, x[11]), 0x1);
+        bclr(DIR_OFFSET(currdir, x[11]), 0x402);
       }
+    } else if (typ == T_EQUAL || typ == T_LP) {
```
kiranchandramohan, line 1202: for the T_LP case should we check for T_RP as well to ensure well formedness?

kiranchandramohan (commented):
Looks fine to me. Tests pass. Two questions.

tools/shared/pragma.c
bryanpkc: AFAIK, this is the only use of bits 0x400, 0x2, and 0x1, in x-flag 11. #742 used 0x400 to mean `NOUNROLL`, and this PR repurposes 0x2 to mean `UNROLL(n)`. The original distinction between 0x1 and 0x2 did not seem applicable to the LLVM loop unrolling parameters.

tools/shared/pragma.c
kiranchandramohan: Some existing code in a different place does the following. Would it make sense to do something similar?

```
    if (typ == T_LP) {
      if (gtok() != T_INT)
        return true;
      i = itok;
      if (gtok() != T_RP)
        return true;
      typ = gtok();
    } 
```

bryanpkc: Yeah, we have written something like that internally (i.e. failing silently), but we are trying to see if we could make it better and produce a syntax error like what clang would do if the `T_RP` is missing, e.g.
```
clang-unroll-pragma.c:4:36: error: expected ')'
  #pragma clang loop unroll_count(8
                                   ^
1 error generated.
```
But in the interest of time, we could just do the simpler thing and get this PR merged, and plan to improve the error reporting in the future.

kiranchandramohan: I am fine with all the improvements coming in a new patch. Approving this one.

kiranchandramohan (approved):
LGTM.

shivaramaarao (approved):
LGTM

PR #945, opened 2020-11-02
libpgmath: Stop using pgstdinit.h

bryanpkc:
This PR fixes a build failure we see on AArch64 Linux with the just-built Clang (based on flang-compiler/classic-flang-llvm-project@da9edcedc4f97506042e04486b27c291c087ca25) and with `_FORTIFY_SOURCE` defined:
```
In file included form /home/bryanpkc/flang/runtime/libpgmath/lib/common/fpcvt.c:12:
/home/bryanpkc/flang/runtime/libpgmath/lib/common/pgstdinit.h:220:5: error: expected parameter declarator
int fprintf(FILE *, const char *, ...);
    ^
/usr/include/bits/stdio2.h:110:26: note: expanded from macro 'fprintf'
  __fprintf_chk (stream, __USE_FORTIFY_LEVEL - 1, __VA_ARGS__)
                         ^
```
pgstdinit.h explicitly declares many stdio functions, which can conflict with the prototypes given in the system libc. The only user of this header is fpcvt.c, which actually works fine with it (at least on Linux).


gklimowicz (approved):

runtime/libpgmath/lib/common/fpcvt.c
```diff
@@ -9,7 +9,6 @@
 #include <string.h>
 #include <ctype.h>
 #include <errno.h>
-#include "pgstdinit.h"
```
kiranchandramohan, line 12: Should this file be removed if it is not used anywhere else?

d-parks: Yes, please remove the file. We have already done so in our source tree.

bryanpkc: Done.

kiranchandramohan (approved):
LGTM. 
Nit: There is a comment left.
runtime/flang/initpar.c: * include/pgstdinit.h:#define __io_getenv(name) getenv(name)

bryanpkc:
@kiranchandramohan Good catch. I grepped the code before but somehow missed this one. Removed.

bryanpkc:
@kiranchandramohan I found that `__fort_getenv` is actually also `#define`-d in stdioInterf.h, which is also included into initpar.c. To prevent `__fort_getenv` calls from getting redirected to `getenv`, I have moved the `#undef` in my latest push, and updated the comment accordingly. I confirmed that on both X86 and AArch64, `__fort_initopt` and `__fort_getopt` call (or inline) the correct implementation both before and after this change, but I wonder if there is any reason not to use `getenv` from libc instead.

@gklimowicz Could you review this again please?

kiranchandramohan:
LGTM. Passes all tests.

shivaramaarao (approved):
LGTM. thanks

PR #946, opened 2020-11-02
Call check_member() for PD_is_contiguous

### LINKED ISSUE #936: [flang1]: Lowering error when using is_contiguous intrinsic on an array pointer in a derived type
An error is observed when using the is_contiguous intrinsic function on an array pointer in a derived type.

Reproducer:
```
module message
  type :: msg_type
    integer, pointer :: arr(:)
  end type msg_type
contains
subroutine test(msg)
    type(msg_type) :: msg
    print *, is_contiguous(msg%arr)
end subroutine
end module
```

Error:
```
Lowering Error: symbol arr$sd is an inconsistent array descriptor
F90-F-0000-Internal compiler error. Errors in Lowering       1  (test.f90: 9)
```

michalpasztamobica:
Fixes #936 .

As @mleair explained in [his comment](https://github.com/flang-compiler/flang/issues/936#issuecomment-720657812):

> I believe we missed the call to check_member() in the PD_is_contiguous case of ref_pd() in semfunc.c

This is correct. The reproducer now compiles fine and the other tests are still passing

I added `is_contiguous` test based on the issue's reproducer to prevent regression.

@mleair , @RichBarton-Arm , @kiranchandramohan , please review.

### DIFF
--- a/test/f90_correct/inc/is_contiguous.mk
+++ b/test/f90_correct/inc/is_contiguous.mk
@@ -1,0 +1,20 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+########## Make rule for test is_contiguous  ########
+
+build:
+	@echo ------------------------------------ building test $(TEST)
+	-$(CC) -c $(CFLAGS) $(SRC)/check.c -o check.$(OBJX)
+	-$(FC) -c $(FFLAGS) $(LDFLAGS) $(SRC)/$(TEST).f90 -o $(TEST).$(OBJX)
+	-$(FC) $(FFLAGS) $(LDFLAGS) $(TEST).$(OBJX) check.$(OBJX) $(LIBS) -o $(TEST).$(EXESUFFIX)
+
+run:
+	@echo ------------------------------------ executing test $(TEST)
+	./$(TEST).$(EXESUFFIX)
+
+verify: ;
+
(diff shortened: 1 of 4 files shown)


kiranchandramohan (approved):
LGTM.

gklimowicz (approved):

shivaramaarao (approved):
internal tests pass. LGTM

bryanpkc (approved):
LGTM. The only nit I have is that, IMHO, test cases that only need to confirm a successful compilation do not need to reside under `f90_correct`, nor do they need the *.sh and *.mk boilerplate.

michalpasztamobica:
@bryanpkc , unfortunately I couldn't get the `make check-all` to build the newly added f90 test without the `.mk and .sh files.
I tried reverting the fix in the code and then running the `make check-all`. I didn't get the error that this PR is fixing until I had both .mk and .sh files in. If you know a way to move the `f90` test file to another location where it gets build without the boilerplate scripts, please let me know more details.

kiranchandramohan:
I believe what Bryan is saying is that f90_correct contains execution tests, so you do not need to place the tests there but can place the test in llvm_ir_correct directory.

Alternatively, you can check the results of the execution (return value of is_contiguous) as shown in the test in the link below?
https://github.com/flang-compiler/flang/blob/9ab6fc8024bea353e0d6888ba928b4e02a4bd4ca/test/f90_correct/src/array_constructor.f90#L26

michalpasztamobica:
@kiranchandramohan , right, I made the test check the `is_contiguous` call result. It will make the test stricter. We have no other `is_contiguous` tests available.
@bryanpkc , is it OK for the test to stay in `f90_correct` now that it actually checks something?

bryanpkc:
@michalpasztamobica Sure, it's fine with me. Thanks!

PR #949, opened 2020-11-16
USE rename should check if renamed sptr is available in the scope

### LINKED ISSUE #889: Problem with rename-list on USE statement

michalpasztamobica:
Fixes #889 .

In case the sptr is not in current scope - check if there is a symbol with same name available.
When an `alias` is being `rename`d, the newly created variable on the stack should also be of `ST_ALIAS` type.

I've also added a test based on @RichBarton-Arm 's reproducer.

@RichBarton-Arm , @kiranchandramohan , would you please review and help me set up the other reviewers appropriately?

### DIFF
--- a/test/f90_correct/inc/rename00.mk
+++ b/test/f90_correct/inc/rename00.mk
@@ -1,0 +1,20 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+########## Make rule for test rename00  ########
+
+build:
+	@echo ------------------------------------ building test $(TEST)
+	-$(CC) -c $(CFLAGS) $(SRC)/check.c -o check.$(OBJX)
+	-$(FC) -c $(FFLAGS) $(LDFLAGS) $(SRC)/$(TEST).f90 -o $(TEST).$(OBJX)
+	-$(FC) $(FFLAGS) $(LDFLAGS) $(TEST).$(OBJX) check.$(OBJX) $(LIBS) -o $(TEST).$(EXESUFFIX)
+
+run:
+	@echo ------------------------------------ executing test $(TEST)
+	./$(TEST).$(EXESUFFIX)
+
+verify: ;
+
--- a/test/f90_correct/lit/rename00.sh
+++ b/test/f90_correct/lit/rename00.sh
@@ -1,0 +1,9 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Shared lit script for each tests. Run bash commands that run tests with make.
+
+# RUN: KEEP_FILES=%keep FLAGS=%flags TEST_SRC=%s MAKE_FILE_DIR=%S/.. bash %S/runmake | tee %t
+# RUN: cat %t | FileCheck %S/runmake
--- a/test/f90_correct/src/rename00.f90
+++ b/test/f90_correct/src/rename00.f90
@@ -1,0 +1,68 @@
+! Code based on the reproducer from https://github.com/flang-compiler/flang/issues/889
+
+module declare_var
+  integer, parameter, public :: var = 8
+end module declare_var
+
+module declare_var_2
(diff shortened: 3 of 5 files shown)


michalpasztamobica:
As @mleair , pointed out inthe issue discussion, this is not a correct solution. It gets rid of the compilation error, but actually the `dp` variable points to an incorrect alias. Still in progres...

michalpasztamobica:
@mleair , @kiranchandramohan , thanks for your support with this. The PR is now ready for review.
I've updated the title and description to reflect what it does now.

michalpasztamobica:
Sure, @mleair . Tests are passing and we can express the intention of the test better. Thanks for the hint!

michalpasztamobica:
@gklimowicz , @bryanpkc , @kiranchandramohan , @shivaramaarao , your review of this patch would be very welcome if you have a moment to take a look at it. Thank you in advance!

gklimowicz (approved):

bryanpkc (approved):
LGTM. The only comment I have is that the reproducer seems to contain multiple occurrences of symbol hiding, and it might be a good idea to break them down into multiple test cases, or add multiple checks in this test case to verify that the correct symbols are used at the correct spots.

michalpasztamobica:
Hi @bryanpkc , thanks for the review. I added two more checks against potentially hidden and incorrect variables. Please let me know if there's anything more you'd like to see checked?

michalpasztamobica:
@kiranchandramohan , @shivaramaarao , any comments on this PR?

kiranchandramohan (approved):
LGTM.

shivaramaarao (approved):
LGTM. thanks for the fix.

Issue #864, opened 2020-02-06
Flang doesn't warn about use of unsupported CHARACTER KIND=2


nSircombe:
Flang's default character set is ASCII, which corresponds to KIND=1.

This appears to be the only supported CHARACTER KIND in Flang (https://github.com/flang-compiler/flang/blob/cb42a171b6462b2f3acd3e1841a56a1ae08291ef/runtime/flang/miscsup_com.c).

Attempting to use other KINDs should generate an error at compile time. For example:

```
PROGRAM charkind
 IMPLICIT NONE                                                                                                                                                                                                                                
 CHARACTER(KIND=2, LEN=4):: data
END PROGRAM charkind
```

Here KIND=2 corresponds, I think, to JIS X 0213. This generates no error at compile time, where as KIND=4 (ISO 10646) does generate an error:
```
F90-S-0081-Illegal selector - KIND parameter has unknown value for data type  (charkind.f90: 5)
  0 inform,   0 warnings,   1 severes, 0 fatal for charkind
```

This can lead to build systems returning a false positive for CHARACTER(KIND=2) support.

PR #955, opened 2020-11-24
Do not issue an error when character kind 2 is used

michalpasztamobica:
Fixes https://github.com/flang-compiler/flang/issues/864.

See issue description for details of the problem.

EDIT: In the course of the discussion and with great help from the reviewers it was proven that in fact `character(kind=2)` is using flang's extension `ncharacter` to handle multibyte characters. There are tests checking this basic functionality and we also managed to store an actual Kanji character correctly. So in the end the fix is not to raise internal compiler errors when `character(kind=2)` is used.

Original outdated description:

I tried to apply a minimum patch that will issue an error, as requested. It seems flang has been to some extent deliberately prepared to support `kind = 2`, because there is a dedicated type for it as well as some documentation. I left all the prepared code in, just amending the information where necessary and modifying the critical part to issue an error and explaining the situation in a comment.

I hope this is as expected, @nSircombe?

@RichBarton-Arm , @kiranchandramohan , as usually - kind request to review the patch and set up more reviewers.

### DIFF
--- a/test/llvm_ir_correct/character_kind2.f90
+++ b/test/llvm_ir_correct/character_kind2.f90
@@ -1,0 +1,14 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+
+! RUN: %flang %s
+
+PROGRAM charlen
+END PROGRAM charlen
+FUNCTION cksum(data)
+CHARACTER(LEN=30):: cksum
+CHARACTER(KIND=2):: data(:)
+END
--- a/tools/flang1/flang1exe/transfrm.c
+++ b/tools/flang1/flang1exe/transfrm.c
@@ -2570,8 +2570,7 @@ trans_freeidx(void)
 LOGICAL
 is_bad_dtype(int dtype)
 {
-  if ((DTYG(dtype) != TY_NCHAR) && (DTYG(dtype) != TY_STRUCT) &&
-      (DTYG(dtype) != TY_UNION))
+  if ((DTYG(dtype) != TY_STRUCT) && (DTYG(dtype) != TY_UNION))
     return FALSE;
   return TRUE;
 }


michalpasztamobica:
@gklimowicz , would you have a moment to review this patch and perhaps help me set up the other reviewers? 
Sorry for bothering you, but unfortunately I don't have the privileges to set this up myself.

gklimowicz (approved):

shivaramaarao (approved):

kiranchandramohan (commented):
Can you add an error test?

KIND=2 seems to work OK for intrinsics. The test test/f90_correct/src/fe82.f90 has the following code,
```
result(73) = kind(len_trim("xxyybc    ",KIND=2))
```

michalpasztamobica:
@kiranchandramohan , thanks for your feedback.
Perhaps my issue analysis went in the wrong direction. The character kind has multiple dimensions and I might have mixed them up... @nSircombe pointed to the [miscsup.c file](https://github.com/flang-compiler/flang/blob/cb42a171b6462b2f3acd3e1841a56a1ae08291ef/runtime/flang/miscsup_com.c) for evidence that we do not support kanji characters, but he did not point to any particular line. I thought that [this snippet](https://github.com/flang-compiler/flang/blob/cb42a171b6462b2f3acd3e1841a56a1ae08291ef/runtime/flang/miscsup_com.c#L4290-L4306) is what he had on mind:
```
/** \brief Check charset
 *
 * Make sure this routine is consistent with
 * - fe90: semfunc.c:_selected_char_kind()
 * - f90:  dinit.c:_selected_char_kind()
 */
static int
_selected_char_kind(char *p, __CLEN_T len)
{
  if (__fortio_eq_str(p, len, "ASCII"))
    return 1;
  else if (__fortio_eq_str(p, len, "DEFAULT"))
    return 1;
  return -1;
}
``` 
Indeed there seems to be just one character kind supported: 1.
But if we do support kanji characters, then perhaps the right solution to the issue would be to enhance `_selected_char_kind()` to handle "JIS X 0213" or "KANJI" or something similar?
Or should we just let code stay as is and not worry about `_selected_char_kind` function?
@kiranchandramohan , @RichBarton-Arm , @nSircombe, @mleair  - I would much appreciate your guidance here.

mleair:
Kanji is supported through the ncharacter data type. Is that a viable solution for the user who reported the error?

The ncharacter data type is an extension that was added several years ago. As @kiranchandramohan mentioned above, there are some tests that use it. I could not find any documentation on it (it's that old of an extension). Those tests probably are the closest we have for documentation.

michalpasztamobica:
The user was so far not responsive on github, but maybe @RichBarton-Arm can check with them?

michalpasztamobica:
@mleair , from the reproducer and the issue description I would suppose the issue might have been found by chance and reported to github so it doesn't get lost. I doubt that the reporter is blocked by this issue. Using `ncharacter` seems the way to go.

@nSircombe, your opinion on this would be much appreciated :)

RichBarton-Arm:
The context is cmake. The 'bug' was reported to us as a problem with cmake feature detection for fortran compilers. It compiles code like:
`cat charlen.f90
PROGRAM charlen
END PROGRAM charlen
FUNCTION cksum(data)
CHARACTER(LEN=30):: cksum
CHARACTER(KIND=2):: data(:)
END`

And expects the compiler to error on it if `CHARACTER(KIND=2)` is not supported to represent multi-byte characters.

It seems like although we have some support for multi-byte characters, it is not expressed using that Fortran code but instead with an extension. So for me, it would be natural for the compiler to fail when it sees input like this, whilst still preserving support for `NCHARACTER`. That Michal's patch passes CI suggests to me that it must do this.

On the other hand, if the `NCHARACTER` extension is somehow compatible with `CHARACTER(KIND=2)` then perhaps it would be ok for flang to allow this input so long as it handles it well. Michal's example shows that clearly it does not.

As Mark says, flang does have multi-byte character support via an extension, so perhaps cmake's flang support could also be improved to check via this extension. I must admit I don't know how much cmake supports flang or even if the above code really exists in CMake as presented. We could report this to cmake as a potential extension to their flang support. I still feel as though the behaviour as reported is wrong though.

michalpasztamobica:
I failed to initialize an `NCHARACTER` variable with an actual Kanji character and print it out to see it in terminal. The tests which @kiranchandramohan mentioned are using normal ASCII characters, not real Kanji (at least I haven't found any Kanji apart from `BLANK = 'A1A1'x`, which is a blank, so I won't see it on terminal...). I am not able to say if either `NCHARACTER` or `CHARACTER(KIND=2)` work with Kanji.

mleair:
> I failed to initialize an `NCHARACTER` variable with an actual Kanji character and print it out to see it in terminal. The tests which @kiranchandramohan mentioned are using normal ASCII characters, not real Kanji (at least I haven't found any Kanji apart from `BLANK = 'A1A1'x`, which is a blank, so I won't see it on terminal...). I am not able to say if either `NCHARACTER` or `CHARACTER(KIND=2)` work with Kanji.

I think you need to specify the Kanji literal with an nc prefix. Fox example,

kanji = nc'漢'

michalpasztamobica:
I tried this and many other things, among others a unicode character as per [kanji unicode table](http://www.rikai.com/library/kanjitables/kanji_codes.unicode.shtml) `kanji = nc'.\4e20'`. This compiles and runs but does not print anything out.
But when I tried 
```
NCHARACTER kanji
kanji = nc'a'
print *,kanji
```
I also got nothing printed, even though I expected to see "a" on the terminal so maybe you cannot print out an `NCHARACTER` variable? The test examples don't attempt to do this.
@mleair , do you know if this is feasible or if there is any other way of checking if NCHARACTER works? The tests seem to focus on NCHARACTER manipulations using alphanumeric ASCII characters, so to me they don't fully prove Kanji support right now.

mleair:
> NCHARACTER kanji
> kanji = nc'a'
> print *,kanji

It's possible that ncharacter is not fully supported or there's a bug in our I/O. I am not sure without really digging into semantio.c, etc. However, this works for me:

character(kind=2) :: kanji
kanji = nc'a'
print *, kanji

and this works for me:

NCHARACTER kanji
kanji = nc'a'
print *, achar(ichar(kanji))
end

michalpasztamobica:
I tried with an actual kanji character and got some difference in behavior:
```
character(kind=2) kanji
kanji = '漢'
print *, kanji
```
Failed to compile with `F90-S-0146-Expression must be character type` in the line with the Kanji character assignment.
```
NCHARACTER kanji
kanji = nc'漢'
print *, achar(ichar(kanji))
```
compiled fine but printed `�` on terminal. I [checked in the code](https://github.com/flang-compiler/flang/blob/2c7b3a15417c851e75e33ea0dbc8d45c8e5ec064/tools/flang1/flang1exe/dtypeutl.c#L3020) that the kanji representation is using [Extended Unix Code](https://en.wikipedia.org/wiki/Extended_Unix_Code). From [here](https://www.itscj.ipsj.or.jp/iso-ir/233.pdf) I infer that '漢' should become 0x3320 while it is becoming 0xbc20, but maybe I am missing some version difference or another detail...
```
$ flang char_kind_2.f90 
$ ./a.out | tee kanji.txt
 �
$ hexdump -x kanji.txt 
0000000    bc20    000a                                                
0000003
```
Anyway - it's not unicode, so it might not print on terminal, but otherwise it seems to work.

So with `NCHARACTER` we have something compiling and running, just using a representation different from UTF8 while with `character(kind=2)` the code doesn't compile if we really use Kanji. Clearly, just changing type to DT_NCHAR is not enough to handle Kanji properly?
I will update the PR to remove the "unsupported" additions. Perhaps Kanji is supported, but not with `CHARACTER(KIND=2)`. I hope this solution is OK for everyone?

mleair:
I think you forgot the nc prefix in the character(kind=2) example. It compiles with the nc prefix, but I don't know if it prints anything.

character(kind=2) kanji
kanji = nc'漢'
print *, kanji

michalpasztamobica:
I skipped it on purpose, thinking `character(kind=2)` should store a character, but you are right @mleair , that the type is changed to `DT_NCHAR`, so it will be able to store an `nc'xxx'`. And it works:
```
$ flang char_kind_2.f90 
$ ./a.out | tee kanji.txt
  
$ xxd kanji.txt 
00000000: 2020 0a                       
```
Nothing gets printed.
The funny thing is that no matter which Kanji character I store in `CHARACTER(KIND=2)`, I get the same hexdump with `0x2020` in it.
For `NCHARACTER` the value stored in the file changes depending on the Kanji I use.

michalpasztamobica:
To sum up:
* storing a Kanji as a normal character in a `character(kind=2)`, e.g. does not compile.
* storing an `ncharacter` in a `character(kind=2)` compiles and runs but seems not to work as expected (no matter what Kanji I put I get 0x2020 stored).
* stroing an `nc` variable with a Kanji in an `ncharacter` seems to work, although I cannot explain the bytes stored, but at least they vary from one character to another.

I think the edited patch (just throw an error if someone uses `character(kind=2)`) is a good step - we are compliant with other compilers and we are not "faking" the support for `character(kind=2)` in fact silently wrapping it with an `ncharacter`, which isn't working anyway.

@mleair , @kiranchandramohan , please let me know if the above makes sense to you? :)

michalpasztamobica:
I still cannot make much sense out of it...
For this Kanji '漢':

1) UTF-8 code is 0x6F22 (see [here](http://www.rikai.com/library/kanjitables/kanji_codes.unicode.shtml))
2) EUC code ([source](https://www.itscj.ipsj.or.jp/iso-ir/233.pdf)) is binary first 7 bits: 0110 100, second 7 bits: 1000 001, which would translate to 0x6882 or 0x3441 depending if the lowest bit is 0 or 1 (the document gives bits1-7, but it's in Japanese and I can't figure out it it thinks there is some unused bit0 or unused bit8). 

I later found another document [here](http://x0213.org/joyo-kanji-code/index.en.html) confirming it is 0x3441 in EUC and 0x6F22 in Unicode

3) Flang's compiled binary prints out 0x208D (or 0x20c1 if I convert the source file to euc-jp with `iconv`)
4) A simple C program I wrote (`long kanji = '漢'; printf("%08lx", kanji);`) prints out 0x00E6BCA2
5) Python:
```
# encoding=utf_8
kanji = '漢'
print kanji
print ':'.join(x.encode('hex') for x in kanji)
```
produces the same as C and can even dump the kanji to output correctly:
```
漢
e6:bc:a2
```
I get an error if I set encoding to euc-jp in Python.

So Python and C print out the same byte contents, although it doesn't seem to be neither UTF8 nor EUC.
Fortran prints out its own thing, which is neither of the standards, but then... if C and Python print out something else as well then it must be me doing something wrong here...


I am trying to figure all of this out to decide if Kanji support exists in flang. At the moment I am out of ideas and this is running out of the original scope of the ticket, which was addressing `character(kind=2)` being not supported.

So the question is: is it OK to support `character(kind=2)` using the underlying `ncharacter`? If not, then we should merge this patch. If yes, then I need to dig further to figure out if Kanji support is working or not.

Regarding Cmake's context - that is another field to explore and I haven't looked into it yet.

mleair:
What does the generated assembly look like?

Is the byte representation that gets printed the same as the byte representation of the Kanji literal in the generated assembly?

bryanpkc:
Don't know if this helps, but:

1. I found an old Fortran standard proposal that discusses `NCHARACTER` in some details:
    https://j3-fortran.org/doc/year/88/107.Post.pdf
2. I wrote a test program in Shift-JIS encoding (EUC-JP should also work):
    ```fortran
          ncharacter kanji*4
    ! 0x82A0
          kanji = nc'あ'
          write(*,100) kanji
    100   format (A2)
          end
    ```
    This compiles and runs, but prints the Japanese character in the wrong endianness:
    ```
    $ ./a.out | xxd
    00000000: a082 0a                                  ...
    ```

michalpasztamobica:
@mleair , I am seeing this in the assembly file regardless if I use `ncharacter` or `character(kind=2)`:
```
100 .C300_MAIN_:                                                                                    
101     .short  59068                           # 0xe6bc                                            
102     .short  162                             # 0xa2     
```
If I convert the source file to EUC-JP with `iconv` I get:
```
101 .C300_MAIN_:
102     .short  46273                           # 0xb4c1 - binary 1011 0100 1100 0001, this is not 0x3441 with wrong endianess
```
If I convert to SHIFT-JIS:
```
104 .C300_MAIN_:
105     .short  35519                           # 0xbf82 - again not 0x3441 with wrong endianess
```
This is in line with the previous analysis I made.

I also checked that the numbers I am getting are not related to 0x3441 that I am expecting (not switched bits or bytes or endianess, the numbers of 1s and 0s don't match).

@bryanpkc , thanks for your input! It is very valuable as you seem to have almost gotten the Kanji stored right. 
However, I did not find the Kanji you used (あ) in the JIS X 0213 character set, neither in [plane 1](https://www.itscj.ipsj.or.jp/iso-ir/233.pdf) nor in [plane 2](https://www.itscj.ipsj.or.jp/iso-ir/229.pdf). I got these documents from [wikipedia](https://en.wikipedia.org/wiki/JIS_X_0213). Can you point to where you got it? Also, if JIS X 0213 assumes 7-bit encoding then how is a value with highest bits set possible (0x82A0)?

bryanpkc:
@michalpasztamobica The character あ is in the plane 1 table you linked, on the 4th row, at the 2nd column.

JIS X 0213 defines both 7-bit and 8-bit encodings, including [Shift-JIS](http://www.rikai.com/library/kanjitables/kanji_codes.sjis.shtml) (the encoding I used). Both the 7-bit JIS and the 8-bit Shift-JIS use a maximum of 2 bytes for Japanese characters (if I remember correctly), while EUC-JP is a variable-width encoding and may use up to 3 bytes.

I'm not sure that the Flang runtime actually cares about encoding (need to check), but based on my experiment, I suspect that `PRINT` and `WRITE` have different levels of support for the `NCHARACTER` type.

michalpasztamobica:
Thanks a lot, @bryanpkc .
Indeed I found the Kanji and got the same results. I also got it for the Kanji I was using.
```
$ iconv -f UTF-8 -t SHIFT-JIS char_kind_2.f90 -o char_kind_2_shift-jis.f90
$ flang char_kind_2_shift-jis.f90
$ ./a.out  | xxd
00000000: a082 0a
```
And the assembly file got the bytes in the right order:
```
123 .C300_MAIN_:
124     .short  33440                           # 0x82a0
```
So in the end - it seems that storing Kanji in `ncharacter`/`character(kind=2)` is working when Shift-JIS is used.
I didn't manage to get the right result when using `EUC-JP` with iconv.

I will now check the CMake aspect...

michalpasztamobica:
**Cmake**

@RichBarton-Arm , I haven't found the code in CMake, that the reporter has used to identify the Fortran compiler to be able to use `character(kind=2)` or not. It must have been some project-specific approach.

CMake requires that the `CMAKE_Fortran_COMPILER` is set to the compiler's path. Then using that compiler it runs a number of checks: OpenMP, MPI, support for modules, but none of them checks for details such as character kind support.

I have also gone through other checks it does for flang and fortran with "PGI fortran" in particular, but nothing targeting Kanji or character kind is there.

So to sum up - I do not see any action for CMake here, unless the reporter points us to where they got this `character(kind=2)` check from.

**Does it work?**

Rather yes. Both `ncharacter` and `character(kind=2)` seem to work when it comes to storing Kanji literals as explained earlier. We have some tests, indicated by @kiranchandramohan , proving that in general `ncharacter` is able to handle data correctly.
I did not manage to find a way to print out a stored Kanji character to the terminal output, but I would say this might not be an issue with `ncharacter` as such, but perhaps with `print/write` functions as @bryanpkc  suggested. 

**Documentation**

I also analyzed the documentation provided by @bryanpkc . It is a document from 1988 and much of the arguments  for using `ncharacter` instead of `character(kind=2)` have become outdated, in my opinion.
* `ncharacter` back then was supposed to become a standard in Japan and proposals for it were submitted for example to SQL. I haven't found any mention of `ncharacter` right now outside of Fortran, but maybe I am missing something...)
* authors of the document did not anticipate that there would be any hardware capable of handling multiple encodings within the next decade
* there is more arguments in the document, some of them I find hard to judge right now.
In summary - as a person rather new to Fortran I find the `ncharacter` extension less intuitive than `character(kind=X)`, so if both work I would opt for leaving the `character(kind=2)` "basing" on `ncharacter` extension as it is now. That would mean closing this PR without merge.

I would much prefer to leave the decision whether to merge this patch or not to more experienced Fortran users. If no one explicitly advocates for merging the patch and disabling `character(kind=2)` support, I will just silently close the PR and the corresponding issue report, as this sounds like the most reasonable way forward.

michalpasztamobica:
@RichBarton-Arm  has come back to me with an update from the reporter. The CMake code was in fact internal to some project and not CMake's, so that is not an issue for us to solve.

However a point was made that the reporter's reproducer (which is now provided as a new test in this patch) raises some internal compiler errors. A quick investigation showed that `is_bad_dtype()` function considers `TY_NCHAR` "bad". 

Therefore the valid approach for this issue is not only not to merge the previous patch which disabled `character(kind=2)`, but instead merge the current patch which fixes the `is_bad_dtype()` function to accept `TY_NCHAR`.

I was considering if a new PR should be created, now that the patch is pretty much contradictory to the original idea, but I decided to edit the title and description slightly and keep to keep the whole discussion in one place. I will stale the reviews that has been made so far.

bryanpkc (approved):
LGTM.

mleair:
LGTM

michalpasztamobica:
@kiranchandramohan , @shivaramaarao , would you be able to give your feedback to this PR after recent changes?

shivaramaarao (approved):

kiranchandramohan (approved):
LGTM.

PR #956, opened 2020-11-30
[test] Minor FileCheck pattern fixes

bryanpkc:
Simplify the FileCheck pattern used in test/openmp_examples/lit/runmake; most of the original pattern was never generated and hence misleading. Prevent check_compilation.py from appending an extraneous space when printing a properly newline-terminated string.

I needed the FileCheck pattern change to catch test failures that would be erroneously reported as passing. The original problem has since been fixed (by 2beb1f781af26009535e511aa7e8ac063f86b65d), but I think these changes are still good to have, especially the check_compilation.py fix.

### DIFF
--- a/test/openmp_examples/lit/runmake
+++ b/test/openmp_examples/lit/runmake
@@ -23,4 +23,4 @@ export PATH=$PATH:$(pwd)
 make -f $MAKE_FILE HOMEQA=$MAKE_FILE_DIR TEST=$test_name OPT="$FLAGS" build 2>&1
 make -f $MAKE_FILE HOMEQA=$MAKE_FILE_DIR TEST=$test_name OPT="$FLAGS" run 2>&1
 make -f $MAKE_FILE HOMEQA=$MAKE_FILE_DIR TEST=$test_name OPT="$FLAGS" verify 2>&1
-# CHECK: {{([1-9][0-9]* tests PASSED\. 0 tests failed|[[:space:]]*PASS(ED)?[[:space:]]*$)}}
+# CHECK: {{^PASS$}}
--- a/test/tools/check_compilation.py
+++ b/test/tools/check_compilation.py
@@ -103,7 +103,8 @@ def getLogErrors(log):
 				terminated = True
 			else:
 				sys.stdout.write (line)
-				sys.stdout.write (" ")
+				if not line.endswith("\n"):
+					sys.stdout.write (" ")
 	return terminated, errorList
 
 def getFileSet(source, logErrors):


gklimowicz (approved):
This passes tests on OpenPOWER for LLVM 9 and LLVM 10.

test/openmp_examples/lit/runmake
```diff
@@ -23,4 +23,4 @@ export PATH=$PATH:$(pwd)
 make -f $MAKE_FILE HOMEQA=$MAKE_FILE_DIR TEST=$test_name OPT="$FLAGS" build 2>&1
 make -f $MAKE_FILE HOMEQA=$MAKE_FILE_DIR TEST=$test_name OPT="$FLAGS" run 2>&1
 make -f $MAKE_FILE HOMEQA=$MAKE_FILE_DIR TEST=$test_name OPT="$FLAGS" verify 2>&1
-# CHECK: {{([1-9][0-9]* tests PASSED\. 0 tests failed|[[:space:]]*PASS(ED)?[[:space:]]*$)}}
+# CHECK: {{^PASS$}}
```
michalpasztamobica, line 26: Just out of curiosity and to get some better understanding of how the tests work: This line looks like a comment... Is this the way it is meant to be?
And looking at the fix you mentioned in the description: Do we now rely on test commands to include the `&& echo PASS` sentence? I looked at a few .mk files at random, but none of them had it...

kiranchandramohan: You might have got it now, but just replying incase.
See the typical usage below. In the second line there is a call to FileCheck and the input is the runmake file. This will cause FileCheck to Check for all patterns beginning with CHECK after a comment symbol.

The fix is only for the openmp_examples directory. And i guess successful tests should print PASS.
```
lit/Example_collapse.2f.sh:# RUN: KEEP_FILES=%keep FLAGS=%flags TEST_SRC=%s MAKE_FILE_DIR=%S/.. bash %S/runmake | tee %t 
lit/Example_collapse.2f.sh:# RUN: cat %t | FileCheck %S/runmake
```

bryanpkc: Sorry for the late reply. The tests themselves were supposed to print `PASS` on success. Some tests relied on the wrapper script to `echo PASS`, and FileCheck would always succeed because the command `echo PASS` itself was being written to FileCheck's input. My fix was to change the regexp and ensure that FileCheck would only match a single `PASS` at the beginning of a line.

michalpasztamobica: Thanks for the explanations, this all makes perfect sense to me now!

michalpasztamobica (approved):

kiranchandramohan (approved):
LGTM.

@shivaramaarao ?

shivaramaarao (approved):

Issue #871, opened 2020-02-18
Flang unable handle nested implied DO loops in array constructors.


nSircombe:
**Minimal Reproducer**
```
PROGRAM test
  ! Array constructors, and implicit DO loops are detailed in sec. 7.8
  ! of the Standard.
  IMPLICIT NONE
  INTEGER, PARAMETER :: n=4
  INTEGER, PARAMETER :: m=10 
  INTEGER :: i, j
  INTEGER :: array(m) = (/ (((i), j=1,i), i=1,n ) /) !(/1, 2, 2, 3, 3, 3, 4, 4, 4, 4.../)
                                                                                                                                                                                                                                              
  PRINT*, array
END PROGRAM test
```
Compiling the above example generates the following error:
```
F90-S-0066-Too few data constants in initialization statement (test.f90: 8)
  0 inform,   0 warnings,   1 severes, 0 fatal for test
```

**Other Compilers**
pgfortran 19.5 on x86 (pgfortran 19.5-0 LLVM 64-bit target on x86-64 Linux) - fails with the same error.
gfortran 9.2 on x86 - compiles and runs
ifort 19.5 on x86 - compiles and runs
gfortran 9.2 on Aarch64 - compiles and runs

**Standards compliance**
I believe the nested implied DO loop construct above is valid fortran. Paraphrasing section 7.8 of the Fortran 2018 standard:

- An array constructor is of the form: (/ _ac-spec_ /)
- _ac-spec_ (in this case) is an _ac-value-list_
- an _ac-value-list_ is just a list of _ac-values_ (obvious but, just to sure, it is stated elsewhere in the Standard)
- an _ac-value_ can be an expression OR an ac-implied-do
- an _ac-implied-do_ is of the form ( _ac-value-list_, _ac-implied-do-control_ )
- _ac-implied-do-control_ is just the obvious "i = 1, n" etc.
- an _ac-value-list_ is a list of _ac-values_ (see above)
- an ac-value can be an _ac-implied-do_ (see above)

It appears to be perfectly valid to nest implied do loops as the code example does, because the ac-value-list which is controlled by the ac-implied-do-control can be anything that you might otherwise be able to use in an array constructor including another implied do-loop.

This appears to be a Flang bug.

michalpasztamobica:
The error comes from [this line](https://github.com/flang-compiler/flang/blob/93179b56533fcf50c0687f2d59cfddaf44068be2/tools/flang1/flang1exe/dinit.c#L324).

[...]

PR #963, opened 2020-12-29
Fix errors on array initialisation with an implied do loop

michalpasztamobica:
Fixes https://github.com/flang-compiler/flang/issues/871.

Summary of the changes in this PR:
* The array is resolved after the checks are performed - use a precalculated size and skip size checks performed before the array is ready.
* flang2 also needs to use a precalculated size.
* Add a test checking that the array is correctly initialised.
* Fixed inconsequent spacing in gram.txt and improved comments.

Rationale:
I believe that the root cause for the issue is that the algorithm performs some checks on an implied-do loop before it actually gets properly resolved. I failed to find an error in earlier stages of the implied-do handling and unless it is used to initialize an array the code works fine. I do not feel competent to redesign the whole algorithm of the data initialization, but I managed to find a solution which acts the way we would expect it:
1) It initializes the array correctly in the https://github.com/flang-compiler/flang/issues/871 reproducer.
2) It reports errors if the implied-do loop has insufficient elements.
3) It does not break any tests.

I achieved this by using a correctly calculated size from an earlier stage of the algorithm and relaxing another check (`cmpat_dtype_with_size`), which also couldn't calculate the size correctly.
I realise that this is more of a work-around, a patch applied to handle a particular use case, so I am fully open to suggestions on a better solution.

### DIFF
--- a/test/f90_correct/src/array_constructor_nested_implied_do.f90
+++ b/test/f90_correct/src/array_constructor_nested_implied_do.f90
@@ -1,0 +1,52 @@
+PROGRAM stress_test_nested_implied_do
+  ! Test that a nested implied do loop can be used to initialize an array.
+  ! Array constructors, and implicit DO loops are detailed in sec. 7.8
+  ! of the Standard.
+  IMPLICIT NONE
+  INTEGER, PARAMETER :: n=4
+  INTEGER, PARAMETER :: m=10
+  INTEGER, PARAMETER :: o=2
+  INTEGER, PARAMETER :: p=3
+  INTEGER :: i, j, k, l
+
+  ! 1 level of nesting
+  INTEGER :: array1(n) = (/ (i, i=1, n) /)
+  INTEGER :: expect1(n)
+  data expect1/1, 2, 3, 4/
+
+  ! 2 levels of nesting
+  INTEGER :: array2(m) = (/ ((i, j=1, i), i=1, n ) /)
+  INTEGER :: expect2(m)
+  data expect2/1, 2, 2, 3, 3, 3, 4, 4, 4, 4/
+
+  ! 3 levels of nesting
+  INTEGER :: array3(m*o) = (/ (((i, j=1, i), i=1,n ), k=1, o) /)
+  INTEGER :: expect3(m*o)
+  data expect3/1, 2, 2, 3, 3, 3, 4, 4, 4, 4, 1, 2, 2, 3, 3, 3, 4, 4, 4, 4/
+
+  ! 4 levels of nesting
+  INTEGER :: array4(m*o*p) = (/ ((((i, j=1, i), i=1,n ), k=1, o), l=1, p) /)
+  INTEGER :: expect4(m*o*p)
+  data expect4/1, 2, 2, 3, 3, 3, 4, 4, 4, 4, 1, 2, 2, 3, 3, 3, 4, 4, 4, 4,&
+               1, 2, 2, 3, 3, 3, 4, 4, 4, 4, 1, 2, 2, 3, 3, 3, 4, 4, 4, 4,&
+               1, 2, 2, 3, 3, 3, 4, 4, 4, 4, 1, 2, 2, 3, 3, 3, 4, 4, 4, 4/
+
+  ! 3 levels of nesting and multiple items
+  INTEGER :: array5(3*m*o) = (/ (((i, i+10, i+100, j=1, i), i=1,n ), k=1, o) /)
+  INTEGER :: expect5(3*m*o)
+  data expect5/1, 10, 100,&
+               2, 20, 200, 2, 20, 200,&
+               3, 30, 300, 3, 30, 300, 3, 30, 300,&
+               4, 40, 400, 4, 40, 400, 4, 40, 400, 4, 40, 400,&
+               1, 10, 100,&
+               2, 20, 200, 2, 20, 200,&
+               3, 30, 300, 3, 30, 300, 3, 30, 300,&
+               4, 40, 400, 4, 40, 400, 4, 40, 400, 4, 40, 400/
+
+  call check(array1, expect1, n)
+  call check(array2, expect2, m)
+  call check(array3, expect3, m*o)
+  call check(array4, expect4, m*o*p)
+  call check(array5, expect5, 3*m*o)
+
+END PROGRAM stress_test_nested_implied_do
--- a/tools/flang1/flang1exe/dinit.c
+++ b/tools/flang1/flang1exe/dinit.c
@@ -358,11 +358,15 @@ dinit_data(VAR *ivl, ACL *ict, int dtype)
                 dinit_acl_val(sptr, DTYPEG(memptr), ict);
               }
             } else if (ict->id == AC_ACONST) {
-              ACL *subict;
+              ACL *subict = ict->subc;
               /* MORE most of these calls to dinit_eval should be calls to
                * get_int_cval, dinit_eval is
                * for evaluating implied so expressions only */
-              ni = dinit_eval(ADD_NUMELM(ict->dtype));
+              if (subict != 0 && subict->id == AC_IDO) {
+                ni = num_elem; // Use the previously calculated size.
+              } else {
+                ni = dinit_eval(ADD_NUMELM(ict->dtype));
+              }
               dinit_acl_val(sptr, DTYPEG(sptr), ict);
             } else {
               /* AC_AST, either a constant or a named constant */
@@ -794,11 +798,22 @@ dinit_acl_val2(int sptr, int dtype, ACL *ict, int op)
       break;
     }
   } else if (ict->id == AC_ACONST) {
-    if (!cmpat_dtype_with_size(dtype, ict->dtype)) {
-      errsev(91);
-    }
-    if (!cmpat_dtype_with_size(DDTG(dtype), DDTG(ict->dtype))) {
-      errsev(91);
+    if ((ict->subc != 0) && (ict->subc->id == AC_IDO)
+        && (ict->subc->subc != 0) && (ict->subc->subc->id == AC_IDO)) {
+      /* Perform a more relaxed check for a nested implied-do loop. */
+      if (!cmpat_dtype(dtype, ict->dtype)) {
+        errsev(91);
+      }
+      if (!cmpat_dtype(DDTG(dtype), DDTG(ict->dtype))) {
+        errsev(91);
+      }
+    } else {
+      if (!cmpat_dtype_with_size(dtype, ict->dtype)) {
+        errsev(91);
+      }
+      if (!cmpat_dtype_with_size(DDTG(dtype), DDTG(ict->dtype))) {
+        errsev(91);
+      }
     }
   } else if (ict->id == AC_IDO) {
     dinit_acl_val2(sptr, dtype, ict->subc, 0);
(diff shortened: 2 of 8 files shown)


mleair (approved):
Looks good to me.

tools/flang2/flang2exe/dinit.cpp
```diff
@@ -580,7 +580,8 @@ dinit_varref(VAR *ivl, SPTR member, CONST *ict, DTYPE dtype,
       *repeat = 0;
       (void)dinit_varref(ivl, member, ict->subc, dtype, struct_bytes_initd,
                          repeat, base_off);
-      i = *repeat = ad_val_of(AD_NUMELM(AD_DPTR(ict->dtype)));
```
kiranchandramohan, line 583: The previous definition of num_elem looks similar. What is the difference in value?
```
num_elem = ad_val_of(AD_NUMELM(AD_PTR(init_sym)));
```

kiranchandramohan (commented):
Some minor comments. Running on an internal test-suite did not show any issues. But it will be good to stress test this change.

mleair:
> Some minor comments. Running on an internal test-suite did not show any issues. But it will be good to stress test this change.

As an FYI: I ran it on a subset of our internal tests and did not see any issues either. I agree, more test cases for this change would be good.

tools/flang2/flang2exe/dinit.cpp
michalpasztamobica: `num_elem` is 10 in the earlier call and 4 (or whatever the value of `n` in the reproducer is) in the one I removed here (same as for `flang1`).

tools/flang1/flang1exe/dinit.c
```diff
@@ -794,11 +798,22 @@ dinit_acl_val2(int sptr, int dtype, ACL *ict, int op)
       break;
     }
   } else if (ict->id == AC_ACONST) {
-    if (!cmpat_dtype_with_size(dtype, ict->dtype)) {
-      errsev(91);
-    }
-    if (!cmpat_dtype_with_size(DDTG(dtype), DDTG(ict->dtype))) {
-      errsev(91);
+    if ((ict->subc != 0) && (ict->subc->id == AC_IDO)
+        && (ict->subc->subc != 0) && (ict->subc->subc->id == AC_IDO)) {
```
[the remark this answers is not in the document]
michalpasztamobica: I will test more levels, but this if will include them as well.

test/f90_correct/src/array_constructor_nested_implied_do.f90
```diff
@@ -0,0 +1,14 @@
+PROGRAM test
```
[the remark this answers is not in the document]
michalpasztamobica: Will do.
There is quite some implied do loops tests [in here](https://github.com/flang-compiler/flang/blob/master/test/f90_correct/src/eg00.f), but they don't cover the array initialization scenario.

tools/flang1/utils/prstab/gram.txt
```diff
@@ -931,8 +931,8 @@
 
 <ac item> ::= <expression> |
               <elp> <ac list> , <implied do control> ) |
-	      <expression> : <expression> <opt stride> |
-	      <elp> <ac list> , <expression> )
+              <expression> : <expression> <opt stride> |
```
SouraVX, line 934: NIT: This is not a change ? or probably unintended ?

michalpasztamobica: Thanks for your alertness, @SouraVX , but actually this change was done intentionally. I was sniffing around this piece of code at an early stage of debugging and noticed that the indentation is inconsistent, so I fixed it.

michalpasztamobica:
@kiranchandramohan , I have improved the test scenario to test different levels of nesting and multiple items. I can't think of any other stress tests, but if you have more ideas - please let me know.

michalpasztamobica:
Now I just force-pushed an empty line at the end of the test file.

pawosm-arm:
Sadly, this fix does not work for REAL type arrays, which now produce a new error (gfortran compiles this code properly):
```
F90-S-0091-Constant expression of wrong data type (librsb.f90: 14)
F90-S-0091-Constant expression of wrong data type (librsb.f90: 14)
```

The affected line 14 is:
```
  REAL(KIND=8),PARAMETER :: VA(nnz) = (/ ((1, j=1,i), i=1,n ) /) ! (/1, 1, 1/)
```

michalpasztamobica:
Thanks for pointing this out @pawosm-arm !

The type comparison function does not allow arrays to have different types:
We hit [this line](https://github.com/flang-compiler/flang/blob/8afbdfb4b16e77e40b332b0fc803ebbe4e266c4a/tools/flang1/flang1exe/dtypeutl.c#L1640):
```
  if (DTY(d1) != DTY(d2))
    return FALSE;
```
Previously we check that both variables are arrays and recurse the function with types of the array.

So the root cause is clear. I will think on the solution, starting from the most obvious, that is - allowing the copying of an `INTEGER` array to a `REAL` array.

pawosm-arm:
@michalpasztamobica will you be updating this PR with fixes for this type mismatch issue?

michalpasztamobica:
That depends if the test you've been running comes from an internal ARM's suite?
If it isn't and we don't break anyone's CI with the current patch, then I suggest to push it to master and file a separate Github issue for the your new finding.
The fixes, in my opinion, will be logically separate, although they will touch the same are of code.

My current plan is to create a new function similar to `cmpat_dtype()` and `cmpat_dtype_with_size()` that will be checking the array's type allowing casts from INT to REAL.

pawosm-arm (approved):

pawosm-arm:
> The fixes, in my opinion, will be logically separate, although they will touch the same are of code.

Then this fix should be merged and the followup should be a subject of new PR addressing the new issue #992

michalpasztamobica:
@bryanpkc , @shivaramaarao , your feedback on this PR are most welcome.

tools/flang2/flang2exe/dinit.cpp
bryanpkc: Which of the three paths above would flang take for an array constructor with an implied do loop? If it is the last block, where `num_elem = ad_val_of(AD_NUMELM(AD_PTR(init_sym)))`, what is `init_sym` set to? Also, it would be nice to add a comment in that block to indicate it is used for handling implied do loops.

michalpasztamobica: Having another look at this code, I am starting to think I might have not understood how it is meant to work correctly...

But answering your questions, @bryanpkc ...
flang2 enters the [first block](https://github.com/flang-compiler/flang/blob/master/tools/flang2/flang2exe/dinit.cpp#L484-L509) . We have `sptr` set to the `array` variable. 
`numelem` is set to 10 and `elemsize` to 4.

Then we [recursively run the same](https://github.com/flang-compiler/flang/blob/8afbdfb4b16e77e40b332b0fc803ebbe4e266c4a/tools/flang2/flang2exe/dinit.cpp#L581) function for the sub-loop and `numelem` is set to 10 and `elemsize` to 4 again and initializes the whole array correctly.

With the old code we later do
```
i = *repeat = ad_val_of(AD_NUMELM(AD_DPTR(ict->dtype)));
```
Which initialized both `i` and `*repeat` to `n` (4 in our case). But we have just initialized the whole array with the recursive call and  if we now run again for just 4 elements this is not going to be enough to match the `num_elem=10`, which the algorithm expects. I also do not understand why should we initialize all 10 elements again.

With the new code, we set both `i` and `*repeat` to 10. And later `numelem - i` becomes 0 and we exit the loop. Now that I think about it this solution might be a bit "hacky".... 

This is the code I am running:
```
  IMPLICIT NONE
  INTEGER, PARAMETER :: n=4
  INTEGER, PARAMETER :: m=10 
  INTEGER :: i, j, k
  INTEGER :: array(m) = (/ (((i), j=1,i), i=1,n ) /) !(/1, 2, 2, 3, 3, 3, 4, 4, 4, 4.../)
```

michalpasztamobica: @bryanpkc , @kiranchandramohan  would you have a look at [this snippet](https://github.com/flang-compiler/flang/blob/master/tools/flang2/flang2exe/dinit.cpp#L5145-L5157) of code?
It claims to "collapse nested array constructor". I examined the code and it seems to [glue all nested loops together into a single one](https://github.com/flang-compiler/flang/blob/72e87c583df2cb015349adac0702019da7190f6c/tools/flang2/flang2exe/dinit.cpp#L5396). 
This would explain why we get just two levels of nesting in the ict, no matter how many nested loops there are:
```
  INTEGER :: array2(m) = (/ ((i, j=1, i), i=1, n ) /) ! ict has two levels of subict
  INTEGER :: array3(m*o) = (/ (((i, j=1, i), i=1,n ), k=1, o) /)   ! ict has two levels of subict
```
Then the recursion found in [dinit_varref](https://github.com/flang-compiler/flang/blob/72e87c583df2cb015349adac0702019da7190f6c/tools/flang2/flang2exe/dinit.cpp#L581) only triggers once to handle the subict and it always performs all iterations of a "collapsed", or I would rather say "unrolled" or "flattened" nested loop.

So my solution effectively just cuts any further evaluations after `dinit_varref()` was called for `ict->subict`, which makes sense in the view of what has been done to the `ict` earlier.

I would stay with the solution I proposed and just add a comment explaining the situations. 
Do you see any flaw in the thinking above and if not - would the current solution be acceptable?

michalpasztamobica:
I have just added the comment as discussed in the [comment in the code](https://github.com/flang-compiler/flang/pull/963/files#r572059150). @bryanpkc , let me know if you also think this is the right way?

@pawosm-arm , trailing whitespaces are now removed :)

tools/flang2/flang2exe/dinit.cpp
bryanpkc: Thanks for the detailed analysis @michalpasztamobica! Is it possible to add a check here, and `assert` or throw ICE if
`ict->subc` isn't "unrolled" as expected?

michalpasztamobica: Adding the internal error/warning wasn't trivial as I saw a few runtime functions throwing it during compilation (the `norm2` function, namely). It took me a while to figure out that there's this hidden info under `u1` structure which lets me distinguish an implied-do loop from other cases. The solution I propose is just below. Let me know if this is what you had in mind?
Also - I suggest we leave it as a warning. This code must have been unused or incorrectly used for years and somehow passed all the testing and didn't raise any concerns, so I am afraid we might block someone's otherwise effective work by adding an error here.

michalpasztamobica: @bryanpkc , is the Internal warning OK as is?

michalpasztamobica:
@bryanpkc , @shivaramaarao , @kiranchandramohan , I hope I applied all the review remarks, would you have a minute to review this PR?

tools/flang2/flang2exe/dinit.cpp
michalpasztamobica: Hi @bryanpkc , I'd be very grateful for your feedback on this, so we can merge this patch and set up the next one, which resolves @pawosm-arm 's issue with casting arrays of different numeric types.

bryanpkc: Yes, LGTM.

bryanpkc (approved):
LGTM. Thanks.

michalpasztamobica:
Thanks for the review, all your valuable remarks and the final approval, @bryanpkc ! 
@SouraVX , @shivaramaarao , it is only AMD's review pending now. Would you have a moment to take a look at this PR, please?

shivaramaarao (approved):

PR #964, opened 2021-01-06
[Flang2] Support for noinline/forceinline pragmas

Chirag-Khandelwal:
This PR adds support for `!dir$ noinline` and `!dir$ forceinline` directives along with the test cases for the same.
These directives are usable right before a routine definition.

### DIFF
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -13177,10 +13177,14 @@ print_function_signature(int func_sptr, const char *fn_name, LL_ABI_Info *abi,
   if (need_debug_info(SPTR_NULL)) {
     /* 'attributes #0 = { ... }' to be emitted later */
     print_token(" #0");
-  } else if (!XBIT(183, 0x10)) {
+  } else if (!XBIT(183, 0x10) || XBIT(14, 0x8)) {
     /* Nobody sets -x 183 0x10, besides Flang. We're disabling LLVM inlining for
      * proprietary compilers. */
+    /* 2nd XBIT - Apply noinline attribute if the pragma "noinline" is given */
     print_token(" noinline");
+  } else if (XBIT(191, 0x2)) {
+    /* Apply alwaysinline attribute if the pragma "forceinline" is given */
+    print_token(" alwaysinline");
   }
 
   if (func_sptr > NOSYM) {
--- a/tools/shared/pragma.c
+++ b/tools/shared/pragma.c
@@ -412,6 +412,7 @@ p_pragma(char *pg, int pline)
 #define SW_SAVE_USED 69
 #define SW_LIBM 70
 #define SW_SIMD 71
+#define SW_FORCEINLINE 73
 
 struct c {
   char *cmd;
@@ -430,6 +431,7 @@ struct c {
 static struct c table[] = {
     {"align", SW_ALIGN, false, S_NONE, S_NONE},
     {"altcode", SW_ALTCODE, true, S_LOOP, S_LOOP | S_ROUTINE | S_GLOBAL},
+    {"forceinline", SW_FORCEINLINE, true, S_ROUTINE, S_ROUTINE | S_GLOBAL},
     {"assoc", SW_ASSOC, true, S_LOOP, S_LOOP | S_ROUTINE | S_GLOBAL},
     {"bounds", SW_BOUNDS, true, S_ROUTINE, S_ROUTINE | S_GLOBAL},
     {"c", SW_C, false, S_NONE, S_NONE},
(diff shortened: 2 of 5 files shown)


tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -13181,6 +13181,12 @@ print_function_signature(int func_sptr, const char *fn_name, LL_ABI_Info *abi,
     /* Nobody sets -x 183 0x10, besides Flang. We're disabling LLVM inlining for
      * proprietary compilers. */
     print_token(" noinline");
+  } else if (XBIT(14, 0x8)) {
```
kiranchandramohan, line 13184: Can we reuse XBIT(183, 0x10) for noinline? If not can the two else ifs be combined?

kiranchandramohan (commented):
Is the attribute not present with a debug flag (-g) compilation?

tools/shared/pragma.c
```diff
@@ -430,6 +431,7 @@ struct c {
 static struct c table[] = {
     {"align", SW_ALIGN, false, S_NONE, S_NONE},
     {"altcode", SW_ALTCODE, true, S_LOOP, S_LOOP | S_ROUTINE | S_GLOBAL},
+    {"alwaysinline", SW_ALWAYSINLINE, true, S_ROUTINE, S_ROUTINE | S_GLOBAL},
```
bryanpkc, line 434: I recommend naming the directive `INLINE` or `FORCEINLINE` instead of `ALWAYSINLINE`, since they seem to be used by other Fortran compilers (Intel, Cray). If you go with `INLINE`, you can also leverage Flang's support for the optional `NO` prefix to consolidate the processing of the directives.

tools/flang2/flang2exe/cgmain.cpp
bryanpkc: The comment for `XBIT(183, 0x10)` suggests that it can be removed if nobody (e.g. from NVIDIA/PGI) cares about non-Flang user of flang2.

Chirag-Khandelwal: I'll combine the two conditions. That okay?

tools/shared/pragma.c
Chirag-Khandelwal: Well, `NOINLINE` already exists. Only force/alwaysinline is required. I went with `alwaysinline` because that's the attribute in clang/llvm. That said, I am changing `alwaysinline` to `forceinline` for, as you said, compatibility reasons.

Chirag-Khandelwal:
> Is the attribute not present with a debug flag (-g) compilation?

Tested that and that is apparently correct. I also observed that it does show up in the LLVM IR file generated by flang2 (just before the opening braces in the function definition), but seems to be erased at the driver level.

tools/shared/pragma.c
```diff
@@ -1297,6 +1299,12 @@ do_sw(void)
      */
     bset(DIR_OFFSET(currdir, x[14]), 8);
     break;
+  case SW_FORCEINLINE:
+    if (no_specified)
+      bclr(DIR_OFFSET(currdir, x[191]), 0x2);
+    else
+      bset(DIR_OFFSET(currdir, x[191]), 0x2);
```
bryanpkc, line 1306: If I understand correctly, the pragma that appears later in the code should override a conflicting pragma that appears earlier. `SW_FORCEINLINE` should clear `XBIT(14, 8)`, and `SW_NOINLINE` should clear `XBIT(191, 2)`.

tools/shared/pragma.c
```diff
@@ -430,6 +431,7 @@ struct c {
 static struct c table[] = {
     {"align", SW_ALIGN, false, S_NONE, S_NONE},
     {"altcode", SW_ALTCODE, true, S_LOOP, S_LOOP | S_ROUTINE | S_GLOBAL},
+    {"forceinline", SW_FORCEINLINE, true, S_ROUTINE, S_ROUTINE | S_GLOBAL},
```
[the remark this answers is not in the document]
Chirag-Khandelwal: Should using `noforceinline` produce an error?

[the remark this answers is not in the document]
bryanpkc: I believe unsupported pragmas should be silently ignored, or perhaps cause a warning at most.

tools/shared/pragma.c
Chirag-Khandelwal: Fixed in 752770a

tools/shared/pragma.c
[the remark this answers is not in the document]
Chirag-Khandelwal: Added the test for `noforceinline` which verifies that the `alwaysinline` attribute isn't generated.

Chirag-Khandelwal:
Closing this PR as the fork was deleted. Created new PR for the same #1017 https://github.com/flang-compiler/flang/pull/1017

Issue #673, opened 2019-02-18
contiguous and save conflict


ajmay81:
There appears to be some inappropriate policing of save and contiguous attributes which causes compilation to fail:

    > cat test.f90
    module test
     implicit none
     integer, dimension(:), pointer, contiguous, save :: ptr
    end module
    > flang -c test.f90 
    F90-S-0134-Illegal attribute - conflict with save (test.f90: 3)
    > gfortran -c test.f90
    > ifort -c test.f90
    >

smparker:
There's a similar error for contiguous and asynchronous:

    > cat test.f03
    subroutine test
      implicit none
      integer, dimension(:), pointer, contiguous, asynchronous :: x
    end subroutine test
    > flang -c test.f03
    F90-S-0134-Illegal attribute - conflict with asynchronous (test.f03: 3)
      0 inform,   0 warnings,   1 severes, 0 fatal for test
    > gfortran -c test.f03
    > ifort -c test.f03

Not sure whether this should be a separate issue as it seems closely related.

gklimowicz:
@smparker Shane, I'll add this test case to our internal tracking. If they're separate issues, we'll handle that on our end. Thanks for the additional information.

pawosm-arm:
Funny finding I've made: if you change order of the CONTIGUOUS and SAVE attribute (namely, `SAVE, CONTIGUOUS` instead of `CONTIGUOUS, SAVE`) this code compiles!

pawosm-arm:
Finally, I've found an easy patch for this issue, at least for the CONTIGUOUS attribute:

```
diff --git a/tools/flang1/flang1exe/semant.c b/tools/flang1/flang1exe/semant.c
index 747b7a1b..2b098ee7 100644
--- a/tools/flang1/flang1exe/semant.c
+++ b/tools/flang1/flang1exe/semant.c
@@ -291,7 +291,7 @@ static struct {
        ET_B(ET_VALUE) | ET_B(ET_VOLATILE) | ET_B(ET_SHARED) |
        ET_B(ET_ASYNCHRONOUS) | ET_B(ET_PROTECTED) | ET_B(ET_PINNED) |
        ET_B(ET_TEXTURE) | ET_B(ET_DEVICE) | ET_B(ET_MANAGED) |
-       ET_B(ET_IMPL_MANAGED))},
+       ET_B(ET_CONTIGUOUS) | ET_B(ET_IMPL_MANAGED))},
     {"target",
[...]

PR #968, opened 2021-01-11
Fix a clash between CONTIGUOUS and SAVE attribute in flang (issue #673)

pawosm-arm:
This simple patch adds CONTIGUOUS attribute to the list of attributes allowed to use along with the SAVE attribute.

Fixes a problem signaled in #673.


kiranchandramohan (approved):
LGTM.

bryanpkc:
@pawosm-arm This doesn't address the ordering issue, right? I didn't look in detail, but it seems that `CONTIGUOUS` would not conflict with any other attribute if it is specified later than the other ones?

How about the conflict between `CONTIGUOUS` and `ASYNCHRONOUS`, also mentioned in #673?

pawosm-arm:
> This doesn't address the ordering issue, right?

By ordering you meant the order those attributes are given? The funny thing (I've also mentioned in the bug report) is that `SAVE, CONTIGUOUS` compiled fine, while the other way round it didn't. With this patch, both orders are accepted.

> I didn't look in detail, but it seems that CONTIGUOUS would not conflict with any other attribute if it is specified later than the other ones?

Indeed.

> How about the conflict between CONTIGUOUS and ASYNCHRONOUS, also mentioned in #673?

I've looked into this just briefly, and it didn't seem to look as obvious as with CONTIGUOUS. Also, I don't have any real life code replicating that, while simultaneous use of CONTIGUOUS and SAVE has appeared recently in cp2k project's git master.

test/llvm_ir_correct/contig.f90
```diff
@@ -0,0 +1,5 @@
+! RUN: %flang -c %s -fsyntax-only
+module contig
+  implicit none
+  integer, dimension(:), pointer, contiguous, save :: ptr
+end module
```
bryanpkc, line 5: This test case should probably be moved into `test/sema/` I guess? Consider naming it `contiguous.f90`.

pawosm-arm: I don't think such a detailed division of test cases would give any additional value to the classic flang, also in the past I've been placing more test cases like this one in this same directory. And adding test case to this directory does not require modifying or adding any other files.

I've deliberately avoided naming this test case 'conriguous' to prevent name clash with one of the f2008 keywords.

bryanpkc (approved):
This patch looks okay to me, but it would probably be useful to open an issue to remind ourselves of the inconsistent handling of `CONTIGUOUS` in the presence of other attributes..

PR #940, opened 2020-10-24
[DebugInfo] Flang should generate debug location for limited instructions in prolog

alokkrsharma:
Due to this wrong debug_line section entry is created with wrong instruction
marked as 'prologue_end'. More details in committed test cases.
This causes debugger not skipping prologue when inserting break-point with
function/subroutine name.
To make breakpoint work as expected, some debugger (GDB) changes were
also required which are already committed as 
commit c2fd7faea8f2c3a267f276ceb6a95f9f537ea7c1
Author: Alok Kumar Sharma <AlokKumar.Sharma@amd.com>
Date:   Thu Aug 20 10:35:27 2020 +0530

PR #969, opened 2021-01-14
Revert "[DebugInfo] Flang should generate debug location for limited instructions in prolog"

kiranchandramohan:
Reverts flang-compiler/flang#940

The test fails in post-commit CI on AArch64. See last comment ing  flang-compiler/flang#940 for reason. 
(Note: The test passed when merged into our tool's codebase. Note sure what is the difference.)


alokkrsharma:
I think we can get rid of test case (dwarfdump_prolog.f90) alone as explained in https://github.com/flang-compiler/flang/pull/940#issuecomment-760138075

kiranchandramohan:
We can restrict to x86 for now. Closing this revert.

PR #970, opened 2021-01-14
Restricting dwarfdump_prolog.f90 test to x864 for now

kiranchandramohan:


kiranchandramohan:
@alokkrsharma I think this test fails on some x86 builds too. So let us remove this test as u suggested.

PR #971, opened 2021-01-14
Remove dwarfdump_prolog.f90 test since it is dependent on codegen

kiranchandramohan:
Remove dwarfdump_prolog.f90 test since it is dependent on codegen
and fails on some architectures and versions of clang.


kiranchandramohan:
Overriding usual process here since without this the post-commit CI fails.
1) I have tested on AArch64 machines.
2) The CI has tested on X86_64 machines.
3) Power is not a requirement.
4) The original author has suggested this in https://github.com/flang-compiler/flang/pull/940

Issue #960, opened 2020-12-16
Infinite loop in find_funcptr_name()


pawosm-arm:
https://github.com/flang-compiler/flang/blob/93179b56533fcf50c0687f2d59cfddaf44068be2/tools/flang2/flang2exe/llassem.cpp#L5503
Shouldn't it be `goto Continue;` instead of just `continue;`? I've encountered a case where `(np - sptrnm != len)` and `(gblsym == FPTR_HASHLK(gblsym))`, the former condition is never checked due to this `continue` and therefore it runs into an infinite loop. It would be checked if it was replaced by `goto Continue`.

pawosm-arm:
@mleair considering this code originates from the initial commit, I think only you may know the answer...

mleair:
@pawosm-arm You might be right. Can you provide a test case that goes into an infinite loop?

pawosm-arm:
Hi @mleair, unfortunately, I don't have a test case or a minimal reproducer. Trouble is, this bug did not manifest itself for decades until I tried to compile latest git master of cp2k (a code newer than the latest cp2k official release) where the problem revealed itself on a significantly large file which imported some number of modules, therefore it had a potential to put a strain on any name searching function. With a problem of that size, it is hard to track what was the unfortunate sequence of searches which resulted in entering this infinite loop.

mleair:
@pawosm-arm  I am not too familar with this code in the compiler. It's obviously doing some sort of search on a hash value. Maybe there is  a bug in the hash function? Or maybe the bug is really somewhere else in the compiler and your change masks the real bug. I suggest you find the problematic code in cp2k. How many source files are there? Maybe you can find the problematic source file using an interative process where you compile all but one file with this change. Then test the application. If the test passes, then move onto the next file until you get a failure.  Once you find the file, you could instrument your compiler to apply the change to certain functions. Or if there are not very many functions in the problematic source file, you probably can use a debugger.

pawosm-arm:
@mleair I've tried to investigate this a little more, still no luck. Responding to your questions:

> How many source files are there?

A lot. Really lot. And they are huge. With crazy dependencies: the largest of the files import hundreds of modules which also import tens of modules and this is a recursive process. All those places define thousands of names. Bottom line: this can really expose any flaw in any name finding routine.

> Maybe you can find the problematic source file

I've prepared a following change in `flang2`:

```
[...]

PR #974, opened 2021-01-20
Fix hash collision handling routine that didn't work due to a fatal mistake (issue #960).

pawosm-arm:
Since the presence of this bug cannot be reliably tested by usual
source code test cases, I've ensured that any malfunction of this algorithm
should result in a fatal ICE. Having this error detection check alone (without
fixing the bug itself) causes certain number of currently used test cases to
fail, this fact alone reveals how well this bug was hidden for years.

Fixes a problem signaled in #960.

### DIFF
--- a/tools/flang2/flang2exe/llassem.cpp
+++ b/tools/flang2/flang2exe/llassem.cpp
@@ -5500,11 +5500,11 @@ find_funcptr_name(SPTR sptr)
         goto Continue;
     } while (*sp);
     if (np - sptrnm != len)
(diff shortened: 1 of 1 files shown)


michalpasztamobica (approved):
I read through the discussion of the issue in https://github.com/flang-compiler/flang/issues/960.
The patch looks good to me. @pawosm-arm mentioned that the bug was actually visible in the current test base after turning on the internal error, so there is no need to add any new tests.

PR #977, opened 2021-01-25
Update apt data before installing sphinx

michalpasztamobica:
Github's default ubuntu image contents might have changed and we need to get the latest ubuntu database to install `sphinx`.


kiranchandramohan:
Overriding the consensus approval process since this patch is needed to keep CI passing. FYI @shivaramaarao @bryanpkc

PR #978, opened 2021-01-25
Fix the readme to point to the correct flang-dev list

RichBarton-Arm:
The readme still points to the old flang-dev list. The link times out for me now. Point it to the new flang-dev list page.


bryanpkc (approved):
LGTM. Since this is so simple, I'll just go ahead and merge.

PR #988, opened 2021-01-31
cmake: disable object depends for Ninja generator

xoviat:


kaadam:
LGTM, but it is just informal. Since it is a general issue, not related to Windows directly.

bryanpkc:
Can you elaborate on the problem this patch is trying to solve?

xoviat:
Ninja doesn't allow multiple rules to generate a target, whereas the other generators do. Ninja also doesn't require these lines. Therefore, this is required to use Ninja.

xoviat:
see https://github.com/ninja-build/ninja/issues/821

kaadam:
As far as I see this issue was already reported here: https://gitlab.kitware.com/cmake/cmake/-/issues/17599

xoviat:
I believe I reported that issue (deleted my account since then, which is why it says "ghost issue"). But testing on the latest cmake/ninja still shows the issue, which is why this fix is needed.

kaadam (approved):
It is okay to me. I think it is useful if we can use ninja during the build especially on Windows.

kiranchandramohan:
ping @bryanpkc

bryanpkc (approved):
LGTM.

kiranchandramohan (approved):
LGTM.

PR #980, opened 2021-01-28
patch flang runtime library

xoviat:
patch the runtime library so that it works on windows.

PR #991, opened 2021-02-02
Clean up Windows related macros

kaadam:
Start using compiler defined macros such as _WIN32, _WIN64.
WINNT macro could be also renamed to _WIN32.

### DIFF
--- a/lib/scutil/pgnewfil.c
+++ b/lib/scutil/pgnewfil.c
@@ -78,7 +78,7 @@ addn(char *p, unsigned long val, int n)
 static LONG pgrand = 0;
 static unsigned long pid = 0;
 
-#if defined(USETEMPNAM) || defined(HOST_WIN) || defined(WIN64)
+#if defined(USETEMPNAM) || defined(HOST_WIN) || defined(_WIN64)
 static int next = 0; /* counter of files created */
 
 /*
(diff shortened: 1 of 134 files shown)


xoviat:
Can #980 be merged first? That's the largest patch that I'd like to see merged soon.

kaadam:
> Can #980 be merged first? That's the largest patch that I'd like to see merged soon.

Yes, could be. After I will rebase on top of your change.

kiranchandramohan:
Will @kaadam and @xoviat be attending the call today? I previously forwarded the invite to you.

kaadam:
> Will @kaadam and @xoviat be attending the call today? I previously forwarded the invite to you.
Yes, I will. Thanks for the invitation.

bryanpkc:
Can you explain the difference between `_WIN32` and `_WIN64` in the commit message, or provide a link to relevant documentation?

kaadam:
@bryanpkc 
_WIN32 Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, x86, or x64. Otherwise, undefined.
_WIN64 Defined as 1 when the compilation target is 64-bit ARM or x64. Otherwise, undefined.

Documentation link: https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160

kiranchandramohan:
@kaadam is this patch still needed?

kaadam:
@kiranchandramohan I close this PR because of these. #1163 #1165 #1167 #1168 They are updated and replace all of user defined old WINNT or WIN64 macros with the compiler predefined macro _WIN64.

PR #1000, opened 2021-02-09
Add ccache support to GitHub Actions

janisozaur:
By default action is set up to:
1. Use 500MB size for cache
2. Compress the cache
3. Show the cache stats post-build to ensure it's used efficiently

### DIFF
--- a/.github/workflows/build_flang.yml
+++ b/.github/workflows/build_flang.yml
@@ -40,6 +40,8 @@ jobs:
 
       - name: ccache
         uses: hendrikmuhs/ccache-action@v1
+        with:
+          key: ${{ matrix.cc }}-${{ matrix.version }}
 
       - if: matrix.cc == 'gcc' && matrix.version == '10'
         run: |
--- a/.github/workflows/build_flang.yml
+++ b/.github/workflows/build_flang.yml
@@ -183,7 +183,9 @@ jobs:
             -DCMAKE_INSTALL_PREFIX=/usr/local \
             -DCMAKE_BUILD_TYPE=Release \
             -DCMAKE_C_COMPILER=/usr/bin/${{ matrix.cc }} \
-            -DCMAKE_CXX_COMPILER=/usr/bin/${{ matrix.cpp }}"
+            -DCMAKE_CXX_COMPILER=/usr/bin/${{ matrix.cpp }} \
+            -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache"
           cd runtime/libpgmath
           mkdir -p build && cd build
           cmake $CMAKE_OPTIONS ..
@@ -202,6 +204,8 @@ jobs:
             -DCMAKE_Fortran_COMPILER_ID=Flang \
             -DFLANG_INCLUDE_DOCS=ON \
             -DFLANG_LLVM_EXTENSIONS=ON \
+            -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             ..
           make -j$(nproc)
           sudo make install
--- a/.github/workflows/build_flang.yml
+++ b/.github/workflows/build_flang.yml
@@ -38,6 +38,9 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so the job can access it
       - uses: actions/checkout@v2
 
+      - name: ccache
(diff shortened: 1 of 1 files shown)


janisozaur:
All the jobs report 0 hits, this needs a bit more inspection

.github/workflows/build_flang.yml
```diff
@@ -38,6 +38,9 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so the job can access it
       - uses: actions/checkout@v2
 
+      - name: ccache
+        uses: hendrikmuhs/ccache-action@v1
```
kiranchandramohan, line 42: Where is this defined?

kiranchandramohan (commented):
Congratulations for making PR 1000, a milestone for Flang. Hope you will be able to complete this PR.

BTW, is this hobby work?

.github/workflows/build_flang.yml
janisozaur: https://github.com/hendrikmuhs/ccache-action

janisozaur: And here on the gh actions marketplace: https://github.com/marketplace/actions/ccache-for-gh-actions

janisozaur:
> BTW, is this hobby work?

Yes, I just happen to know @michalpasztamobica 

Previously I haven't noticed I was editing only one of the jobs in the matrix. It should be good now.

janisozaur:
Caches now populate correctly and in the case of GCC 9, it even got used already:
```
cache directory                     /home/runner/work/flang/flang/.ccache
primary config                      /home/runner/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
stats updated                       Tue Feb  9 14:19:15 2021
cache hit (direct)                     0
cache hit (preprocessed)              12
cache miss                          1406
cache hit rate                      0.85 %
cleanups performed                     0
files in cache                      2821
cache size                          17.7 MB
max cache size                     500.0 MB
```

As it was only for a minor part of the build, you can see it in the line:
```
cache hit (preprocessed)              12
```

.github/workflows/build_flang.yml
junghans: Is there a reason to use this custom cacching action over the Github-provided one (https://github.com/actions/cache)?

janisozaur: Yes - the GitHub cache action only provides a way of caching, the use of you actually have to implement yourself. The ccache action used here provides that and builds on top of that - installs ccache itself, configures some options (the ones I mentioned in the opening comment), ensures ccache is available in the `$PATH`, creates a log entry with stats.

junghans: Thanks!

janisozaur:
ping

.github/workflows/build_flang.yml
kiranchandramohan: Only concern is using an action in a private individual's repo.

WDYT, @bryanpkc @shivaramaarao

bryanpkc: The TypeScript code looks pretty straightforward. Does the presence of the action in the GitHub Marketplace ensure its continued availability? Will it disappear suddenly and break our CI?

janisozaur: I doubt presence on the actions marketplace guarantees anything. If that's a concern for you, you may consider forking the project into `flang-compiler` organisation. Personally though I wouldn't recommend that, this PR is quite small and self-contained. Should the action ever be removed, fallback to previous version is easy enough.

Alternatively, ccache installation and caching can be rewritten in bash so that you don't need to rely on external actions, but I'm not keen on doing that when current solution seems sufficient.

bryanpkc: @janisozaur I agree with your view. I don't have a concern with this use.

bryanpkc:
@janisozaur Would you consider creating a similar PR to implement ccache for https://github.com/flang-compiler/classic-flang-llvm-project? IMO the LLVM build will benefit much more from caching than the Flang build, since it takes way longer.

janisozaur:
Sure!

janisozaur:
Just had a look at that repo - it seems so much larger than this one, it may exceed the (default) cache capacity available. GitHub offers up to 5GB (https://github.com/actions/cache#cache-limits) so it should be possible to bump it a bit higher. This may need some manual fiddling with options.

michalpasztamobica (approved):

janisozaur:
https://github.com/flang-compiler/classic-flang-llvm-project/pull/17

bryanpkc (approved):
LGTM.

PR #1001, opened 2021-02-09
Dummy PR to test ccache setup

janisozaur:
This PR intends to test ccache implemented in #1000. As #1000 is not merged just yet, I don't actually know if caches will be available to this job or not.


janisozaur:
The caches seem unavailable until the caching is merged into the main branch of the repository.

Issue #992, opened 2021-02-02
Array initialisation with an implied do loop: allow the copying of an INTEGER array to a REAL array


pawosm-arm:
This is the followup for the solution #963 for the #871 issue. The solution presented in #963 exposed a new problem when facing REAL-typed arrays, e.g.:

```
PROGRAM test
  ! Array constructors, and implicit DO loops are detailed in sec. 7.8
  ! of the Standard.
  IMPLICIT NONE
  INTEGER, PARAMETER :: n=4
  INTEGER, PARAMETER :: m=10 
  INTEGER :: i, j
  REAL(KIND=8),PARAMETER :: array(m) = (/ ((1, j=1,i), i=1,n ) /) ! (/1, 1, 1/)
                                                                                                                                                                                                                                            
  PRINT*, array
END PROGRAM test
```
...causes following error to be thrown by compiler:
```
F90-S-0091-Constant expression of wrong data type
```

michalpasztamobica:
I noticed that the current approach is to [allow copying between INT arrays even if they are of different kind](https://github.com/flang-compiler/flang/blob/8afbdfb4b16e77e40b332b0fc803ebbe4e266c4a/tools/flang1/flang1exe/dtypeutl.c#L1885) without checking if they will fit into each other or not:
```
    if (DT_ISINT(d1) && DT_ISINT(d2))
      return TRUE;
```
I guess we should only allow copying if the source array can fit into target array.
So for example this should work (4-byte INT array can hold data from a 2-byte INT array):
```
INTEGER(KIND=4) :: array_i4_i2(3) = (/ ((1_2, j16=1_2, i16), i16=1_2, 3_2 ) /)
```
And this should not (1-byte INT array may not be able to hold data from a 2-byte INT array if the numbers are too large):
```
INTEGER(KIND=1) :: array_i1_i2(3) = (/ ((1_2, j16=1_2, i16), i16=1_2, 3_2 ) /)
```
Same holds for real (REAL4 won't hold REAL8)
Is this right?

pawosm-arm:
> So for example this should work (4-byte INT array can hold data from a 2-byte INT array):
```
INTEGER(KIND=4) :: array_i4_i2(3) = (/ ((1_2, j16=1_2, i16), i16=1_2, 3_2 ) /)
```

Are you sure about `KIND=4`? It fails to compile due to type mismatch. Yet after I changed it to `KIND=2`, it compiled fine!

bryanpkc:
[...]

PR #1004, opened 2021-03-02
Nested implied do loop fix for real numbers

michalpasztamobica:
Fixes https://github.com/flang-compiler/flang/issues/992.

A new `cmpat_dtype_*` function is added that does a few preliminary checks to allow casting of arrays wherever this is possible.
Also a few test cases are added presenting the possibility of initializing arrays with different types.

### DIFF
--- a/test/f90_correct/src/array_constructor_nested_implied_do.f90
+++ b/test/f90_correct/src/array_constructor_nested_implied_do.f90
@@ -34,14 +34,90 @@ PROGRAM stress_test_nested_implied_do
   ! 3 levels of nesting and multiple items
   INTEGER :: array5(3*m*o) = (/ (((i, i+10, i+100, j=1, i), i=1,n ), k=1, o) /)
   INTEGER :: expect5(3*m*o)
-  data expect5/1, 10, 100,&
-               2, 20, 200, 2, 20, 200,&
-               3, 30, 300, 3, 30, 300, 3, 30, 300,&
-               4, 40, 400, 4, 40, 400, 4, 40, 400, 4, 40, 400,&
-               1, 10, 100,&
-               2, 20, 200, 2, 20, 200,&
-               3, 30, 300, 3, 30, 300, 3, 30, 300,&
-               4, 40, 400, 4, 40, 400, 4, 40, 400, 4, 40, 400/
+  data expect5/1, 11, 101,&
+               2, 12, 102, 2, 12, 102,&
+               3, 13, 103, 3, 13, 103, 3, 13, 103,&
+               4, 14, 104, 4, 14, 104, 4, 14, 104, 4, 14, 104,&
+               1, 11, 101,&
+               2, 12, 102, 2, 12, 102,&
+               3, 13, 103, 3, 13, 103, 3, 13, 103,&
+               4, 14, 104, 4, 14, 104, 4, 14, 104, 4, 14, 104/
+
+  INTEGER(KIND=1) :: i8, j8   ! TY_BINT
+  INTEGER(KIND=2) :: i16, j16 ! TY_SINT
+  INTEGER(KIND=4) :: i32, j32 ! TY_INT
+  INTEGER(KIND=8) :: i64, j64 ! TY_INT8
+
+  ! Different integers fitting into each other
+  ! Array contents are intentionally oversized to show that a warning rather than an error is thrown
+
+  INTEGER(KIND=1) :: array_i1_i1(6) = (/ ((256_1, j8=1_1, i8), i8=1_1, 3_1 ) /)
+  !{warning "PGF90-W-0128-Integer constant truncated to fit data type: 1" 61}
+  !{warning "PGF90-W-0128-Integer constant truncated to fit data type: 1" 61}
+  !{warning "PGF90-W-0128-Integer constant truncated to fit data type: 1" 61}
+  !{warning "PGF90-W-0128-Integer constant truncated to fit data type: 1" 61}
+  !{warning "PGF90-W-0128-Integer constant truncated to fit data type: 1" 61}
+  !{warning "PGF90-W-0128-Integer constant truncated to fit data type: 1" 61}
+  INTEGER(KIND=1) :: array_i1_i2(6) = (/ ((257_2, j16=1_2, i16), i16=1_2, 3_2 ) /)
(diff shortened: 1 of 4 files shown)


pawosm-arm (approved):
I've checked that it fixes the issue I've encountered.

test/f90_correct/src/array_constructor_nested_implied_do.f90
```diff
@@ -43,6 +43,46 @@ PROGRAM stress_test_nested_implied_do
                3, 30, 300, 3, 30, 300, 3, 30, 300,&
                4, 40, 400, 4, 40, 400, 4, 40, 400, 4, 40, 400/
 
+  INTEGER(KIND=1) :: i8, j8   ! TY_BINT
+  INTEGER(KIND=2) :: i16, j16 ! TY_SINT
+  INTEGER(KIND=4) :: i32, j32 ! TY_INT
+  INTEGER(KIND=8) :: i64, j64 ! TY_INT8
+
+  ! Different integers fitting into each other
+  ! Array contents are intentionally oversized to show that a warning rather than an error is thrown
```
bryanpkc, line 52: Consider matching the warnings with check_compilation.py.

tools/flang1/flang1exe/dtypeutl.c
```diff
@@ -2115,6 +2115,50 @@ same_dtype(DTYPE d1, DTYPE d2)
...
+     * _TY_INT attribute.
+     */
+    if (DT_ISLOG(d1) && DT_ISLOG(d2))
+      return TRUE;
+    if (DT_ISINT(d1) && DT_ISINT(d2))
+      return TRUE;
+    return FALSE;
+  }
+  if (DTY(d1) == TY_CHAR || DTY(d1) == TY_NCHAR)
+    return TRUE;
```
bryanpkc, line 2141: Does this mean that any type can be successfully cast to `CHARACTER` type (with truncation)?

michalpasztamobica: To me this means that "all character types are compatible with each other". In the lines above we have already established that the types are different and that at least one of them is not logical and not an integer.
I took this line and the lines above from `cmpat_dtype()` [function body](https://github.com/flang-compiler/flang/blob/442d797075f4b5cd461593a2e216d0f49cb1eba5/tools/flang1/flang1exe/dtypeutl.c#L1889) and built on top of that similarly to what `cmpat_dtype_with_size()` has done.
Also see [the comment](https://github.com/flang-compiler/flang/blob/442d797075f4b5cd461593a2e216d0f49cb1eba5/tools/flang1/flang1exe/dtypeutl.c#L1930) which `cmpat_dtype_with_size()` has added before this check.

test/f90_correct/src/array_constructor_nested_implied_do.f90
michalpasztamobica: Sure, I will check what the script is doing and consider how to use it for these checks. Thanks for the hint!

michalpasztamobica: BTW, please check the [comment I added in the issue report](https://github.com/flang-compiler/flang/issues/992#issuecomment-780572479). `gfortran` is throwing errors here (or nothing if you use `-fno-range-check`)

tools/flang1/flang1exe/dtypeutl.c
bryanpkc: Thanks for the detailed explanation; I was not reading carefully enough.

tools/flang1/flang1exe/dtypeutl.c
```diff
@@ -2115,6 +2115,50 @@ same_dtype(DTYPE d1, DTYPE d2)
...
+
+  /* Perform checks for array casting possibilities */
+  if (DTY(d1) == TY_ARRAY && DTY(d2) == TY_ARRAY) {
```
bryanpkc, line 2144: Since we know `DTY(d1) == DTY(d2)`, why check both of them here?

test/f90_correct/src/array_constructor_nested_implied_do.f90
michalpasztamobica: I added the `check_compilation.py` usage and I manually checked that it passes fine.
But if I "break" the expected messages or put them to a wrong line and manually check that the test failed, I can still see the test suite passing. 
If I "break" the test so it does not compile or run - I can see the failures alright.

For example I "broke" the message printed for line 103, by removing most of its body. I saw the output of the check_compilation.py to be:

```
  0 inform,  36 warnings,   0 severes, 0 fatal for stress_test_nested_implied_do
('Unexpected:', 'warning', '/home/mipas/flang/flang/test/f90_correct/lit/../src/array_constructor_nested_implied_do.f90', '103', 'Integer constant truncated to fit data type: 1')
('Missed:', 'warning', '/home/mipas/flang/flang/test/f90_correct/lit/../src/array_constructor_nested_implied_do.f90', '103', 'Integer const')
FAIL
```

Yet, the suite as a whole did not report this...
Any ideas what I might be doing wrong?

tools/flang1/flang1exe/dtypeutl.c
michalpasztamobica: True, I removed the second check and added a comment explaining that we are already in a place where we know that the types are identical.

test/f90_correct/src/array_constructor_nested_implied_do.f90
michalpasztamobica: Nevermind, must be something with my local setup. CI has just shown me some failures, and when I manually checked the logs I saw them too, even though the suite allegedly passed all tests.

michalpasztamobica: @bryanpkc , the tests are passing now with the additional checks added.

bryanpkc: Thanks.

bryanpkc (approved):
LGTM.

michalpasztamobica:
@mleair , @shivaramaarao , @SouraVX , would you find some time to review this PR on behalf of NVIDIA and AMD, please?

shivaramaarao (approved):
Looks fine for me.

Issue #1008, opened 2021-03-10
Runtime crash due to an incorrect string length in a string array pointer


bryanpkc:
This test program crashes with a segmentation fault when compiled with the latest Flang on X86:
```fortran
program example
      implicit none
      character(:), pointer :: ptr(:)
      character(3), target :: array(1) = ['foo']

      ptr(2:2) => array
      ptr(2) = 'bar'

      print *, array
end
```
When compiled with gfortran, the program prints "bar" as expected.

The backtrace of the crash is as follows:
```
Program received signal SIGSEGV, Segmentation fault.
tcache_get (tc_idx=<optimized out>) at malloc.c:2937
2937    malloc.c: No such file or directory.
(gdb) bt
#0  tcache_get (tc_idx=<optimized out>) at malloc.c:2937
#1  __GI___libc_malloc (bytes=1024) at malloc.c:3051
#2  0x00007ffff766fe84 in __GI__IO_file_doallocate (fp=0x7ffff77d76a0 <_IO_2_1_stdout_>) at filedoalloc.c:101
#3  0x00007ffff7680050 in __GI__IO_doallocbuf (fp=fp@entry=0x7ffff77d76a0 <_IO_2_1_stdout_>) at libioP.h:948
#4  0x00007ffff767f0b0 in _IO_new_file_overflow (f=0x7ffff77d76a0 <_IO_2_1_stdout_>, ch=-1) at fileops.c:745
#5  0x00007ffff767d835 in _IO_new_file_xsputn (n=1, data=<optimized out>, f=<optimized out>) at libioP.h:948
#6  _IO_new_file_xsputn (f=0x7ffff77d76a0 <_IO_2_1_stdout_>, data=<optimized out>, n=1) at fileops.c:1197
#7  0x00007ffff7671541 in __GI__IO_fwrite (buf=0x7ffff7ef385e, size=1, count=1, fp=0x7ffff77d76a0 <_IO_2_1_stdout_>) at libioP.h:948
#8  0x00007ffff7b94c0d in __io_fwrite (ptr=0x7ffff7ef385e " ", size=1, nitems=1, stream=0x7ffff77d76a0 <_IO_2_1_stdout_>) at /home/bryanpkc/src/llvm/flang/runtime/flangrti/iostdinit.c:253
[...]

PR #1009, opened 2021-03-11
[flang1] Set type length of string array pointer

bryanpkc:
This patch fixes a bug in pointer association that would set the type length of a `CHARACTER(:), POINTER` variable to an incorrect value, which would lead to a crash in the runtime during I/O.

This is a first attempt to fix issue #1008. We are not certain whether `emit_alnd_secd` is the best spot to fix this problem, nor how many other scenarios (apart from the rank-1 simply contiguous target case we have analyzed) we are not handling correctly. Advice from the community would be most welcome.

Based on a patch originally by Peixin Qiao.

### DIFF
--- a/test/f90_correct/inc/string_array_pointer_1.mk
+++ b/test/f90_correct/inc/string_array_pointer_1.mk
@@ -1,0 +1,18 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+$(TEST): run
+
+build: $(SRC)/$(TEST).f90
+	-$(RM) $(TEST).$(EXESUFFIX) core *.d *.mod FOR*.DAT FTN* ftn* fort.*
+	@echo ------------------------------------ building test $@
+	-$(FC) $(FFLAGS) $(LDFLAGS) $(SRC)/$(TEST).f90 -o $(TEST).$(EXESUFFIX)
+
+run:
+	@echo ------------------------------------ executing test $(TEST)
+	$(TEST).$(EXESUFFIX)
+
+verify: ;
--- a/test/f90_correct/inc/string_array_pointer_2.mk
+++ b/test/f90_correct/inc/string_array_pointer_2.mk
@@ -1,0 +1,18 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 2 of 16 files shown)


test/f90_correct/src/string_array_pointer.f90
```diff
@@ -0,0 +1,19 @@
...
+
+! Check that the type length of a string array pointer is set correctly.
+
+program string_array_pointer_test
+  implicit none
+  character(:), pointer :: ptr(:)
+  character(3), target :: array(1) = ['foo']
+  ptr(2:2) => array
+  ptr(2) = 'bar'
+  if (len(ptr) /= 3) stop 1
```
pawosm-arm, line 15: I feel slightly confused here. Is this the len value we care about in the #1008? IMHO we should check for length of the pointed string, namely:
```
  if (len(ptr(2)) /= 3) stop 1
```

bryanpkc: The code is correct and equivalent to what you suggest.

The Fortran 2018 standard states (16.9.112, "LEN (STRING [, KIND])"):
> **Result Value.** The result has a value equal to the number of characters in STRING if it is scalar or in an element of STRING if it is an array.

pawosm-arm (approved):

kiranchandramohan:
ping @SouraVX @shivaramaarao

bryanpkc:
@pawosm-arm @mleair @kiranchandramohan @SouraVX @shivaramaarao  Please review again, as I have pushed an updated patch with more test cases.

I have also added another commit to fix a related problem. The `flags` argument was not being passed to the correct parameter of the `RTE_template` runtime functions (when F90 descriptors are enabled, which is the default). As a result, the descriptor created by `RTE_template` would end up with a bogus type length and a `flags` of zero. I couldn't come up with a regression test case, but you can see the effect by recompiling `flang/runtime/flang/rdst.c` with `-DDEBUG` and setting the environment variable `PGHPF_TEST=64` to enable `__fort_show_flags`:
```
$ ./flagstest.before > /dev/null
flags=0x0
$ ./flagstest.after > /dev/null
flags=0xa, SEQUENCE, SAVE
```
I used the following test program:
```fortran
program flagstest
real, save, dimension(2) :: arr = 1
integer:: length
length = len(get_string(arr))
print *, length
contains
 function get_string(marg) result(res_string)
   real, dimension(:), intent(in) :: marg
   character(len=size(marg)) :: res_string
 end function get_string
end program flagstest
```

pawosm-arm (approved):

SouraVX (approved):
LGTM, Thanks :)

bryanpkc:
Thanks for the reviews.

Issue #1005, opened 2021-03-02
The callsite of f90_mmul_* flang runtime functions wrongly generated


pawosm-arm:
@mleair this bug is extremely hard to track and causes random errors and is very fragile to whatever was encountered on the stack. Also I still don't have a good minimal reproducer that would actually crash. Let me try to describe the problem anyway.

Consider the following program:
```
MODULE kinds
  INTEGER, PARAMETER :: dp = KIND(0.0D0)
END MODULE

program matmu
  use kinds
  integer, parameter :: m = 10, n = 20
  REAL(KIND = dp), DIMENSION(1:n, 1:n, 0:n)  :: hmat, umat

  hmat = 1._dp
  umat = -1._dp
  print *, hmat, umat
  call do_matmul(hmat, umat, 3, 3, 1)

contains

  subroutine do_matmul(hmat, umat, m, n, l)
    use kinds
    REAL(KIND = dp), DIMENSION(:, :, 0:), INTENT(IN)     :: hmat, umat
    integer, intent(in) :: m, n, l
    REAL(KIND = dp), ALLOCATABLE, DIMENSION(:, :)        :: a, tmp1, tmp2

    allocate(a(n, n))
    a(1:m, 1:m) = MATMUL(TRANSPOSE(umat(1:n, 1:m, l)), MATMUL(hmat(1:n, 1:n, l), umat(1:n, 1:m, l)))
    print *, a(1, 1)
    deallocate(a)
  end subroutine do_matmul

end program matmu
```

When compiled and executed, it prints what we expect and completes succesfully.
Yet when examined carefully in gdb, it turns out, something went wrong unnoticed under the hood:
```
(gdb) b f90_mmul_real8
Breakpoint 1 at 0x400af4
(gdb) r
...
Breakpoint 1, f90_mmul_real8 (ta=0, tb=0, mra=3, ncb=3, kab=3, alpha=0x401db0 <.C292_matmu_do_matmul>, a=0x4597c0, lda=3, b=0x459830, ldb=281474976690608, beta=0x401da8 <.C291_matmu_do_matmul>, c=0x45a4e0, ldc=3) at runtime/flang/mmreal8.c:58
58        int small_size = SMALL_ROWSA * SMALL_ROWSB * SMALL_COLSB;
(gdb)
```

Look at the `ldb` value and compare it to what `lda` and `ldc` have. It is definitely wrong, however in most cases it does not hurt. Until it hurts like that:
```
[...]

PR #966, opened 2021-01-09
Fixes to address cp2k compilation and runtime issues

shivaramaarao:
This patch addresses the compilation and runtime issues of CP2K. Following additional flags are required for successful cp2k compilation and run.
1. -Hx,2,0x400000  - disable pointer optimizations
2. -Hx,47,0x1000000  - to disable calls to streamlined/dgemm-like matmul run-time routines

PR #1011, opened 2021-03-15
Fix for the issue #1005 : Do not call streamlined/dgemm-like matmul runtime routines for data layouts that could not be proven to be contiguous

pawosm-arm:
Returning FALSE from mmul_arg() function prevents further preparations for making a call to the dgemm-alike f90_mmul_* flang runtime functions. Instead, it should fall back to making a call to one of the f90_matmul_* runtime functions.

### DIFF
--- a/test/f90_correct/inc/mmul_misc3.mk
+++ b/test/f90_correct/inc/mmul_misc3.mk
@@ -1,0 +1,27 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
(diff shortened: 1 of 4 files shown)


shivaramaarao (approved):
tested and LGTM

jing64:
I saw in https://github.com/flang-compiler/flang/pull/966 that we need `-Hx,47,0x1000000` to avoid a run-time issue caused by streamlined/dgemm-like matmul run-time routines. Is this patch addressing the same problem?

pawosm-arm:
Yes, using -Hx,47,0x1000000 could have been used as a temporary workaround, yet this is a total solution that prevents from using streamlined/dgemm-like matmul run-time routines even in situations where it is safe. This patch disables generating the calls to the streamlined/dgemm-like matmul run-time routines exactly in the places where generated code is blatantly wrong.

jing64:
Thanks for the clarification!

PR #1025, opened 2021-04-08
runtime: register atfork handler to re-initialize internal flangrti locks at fork

pawosm-arm:
The llcrit.c code of flangrti runtime library creates internal locks
which are initialized on the first encounter of any Fortran procedure
that contains at least one OpenMP directive. The flang compiler emits
a call to one of the two functions modified by this patch at the
beginning of every Fortran procedure containing at least one OpenMP
directive. When executed for the first time, these functions
initialize internal locks and modify corresponding static global
variable confirming the initialization has actually occured. On any
subsequent execution, these functions make use of those locks.

Unfortunately, the values of those global indicator variables are not
being reset at fork, hence the locks are not being initialized in the
child processes. Any attempt of using those internal locks results in
a runtime crash of the child process.

This patch registers atfork handler executed as the first thing by any
of the children process.

A test case depicting the problem is also provided with this patch.

### DIFF
--- a/runtime/flangrti/llcrit.c
+++ b/runtime/flangrti/llcrit.c
@@ -5,9 +5,12 @@
  *
  */
 
+#include <stddef.h>
 #include <stdlib.h>
 #include <stdint.h>
 #include <string.h>
+#include <pthread.h>
+#include <stdioInterf.h>
 #include "komp.h"
 
 /* This routine makes a simple omp library call to force lazy initialization of
@@ -63,6 +66,19 @@ static kmp_critical_name nest_sem;
 static omp_nest_lock_t nest_lock;
 
 static int is_init_nest = 0;
+static int is_init_nest_red = 0;
+static int is_atfork_registered = 0;
+
+static void __llcrit_atfork(void)
+{
+  is_init_nest = 0;
+  is_init_nest_red = 0;
+  /* The atfork handlers are inherited by the sub-processes,
+   * see https://elias.rhi.hi.is/libc/Threads-and-Fork.html
+   */
+  if (!is_atfork_registered)
+    fprintf(__io_stderr(), "The atfork not registered when it should be!\n");
+}
 
 void
 _mp_bcs_nest(void)
@@ -70,6 +86,12 @@ _mp_bcs_nest(void)
   if (!is_init_nest) {
     _mp_p(&nest_sem);
     if (!is_init_nest) {
+      if (!is_atfork_registered) {
+        if (pthread_atfork(NULL, NULL, __llcrit_atfork))
+          fprintf(__io_stderr(), "Could not register atfork handler!\n");
+        else
+          is_atfork_registered = 1;
+      }
       omp_init_nest_lock(&nest_lock);
       is_init_nest = 1;
     }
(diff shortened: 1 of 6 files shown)


SouraVX:
Changes, look legit. But when I tried the test case(part of this PR) without code changes:
```
           12
            0
            7
           15
            8
            2
            3
            1
           10
            4
           11
            6
            9
           13
           14
            5
FAIL
```
>Unfortunately, the values of those global indicator variables are not
being reset at fork, hence the locks are not being initialized in the
child processes. Any attempt of using those internal locks results in
a runtime crash of the child process.

Based on your comment, I was expecting a CRASH :). Am I missing something, or could you provide a repro for crash.

SouraVX:
Steps followed while building:
```
clang -c -I/home/flang/test/mp_correct/lit/../src  -fopenmp  /home/flang/test/mp_correct/lit/../src/fork.c
clang -c -I/home/flang/test/mp_correct/lit/../src  -fopenmp  /home/flang/test/mp_correct/lit/../src/check.c -o check.o
flang -c -I/home/flang/test/mp_correct/lit/../src  -fopenmp  /home/flang/test/mp_correct/lit/../src/fork_omp.F90
flang   -fopenmp fork_omp.o fork.o check.o -o a.out
```

pawosm-arm:
If you see `FAIL`, then there was a crash and it was detected by waitpid(). If you really want to post-mortem-debug it, enable core dumps in your system (with `ulimit -c unlimited`), in effect, whenever you see `FAIL` message a core file should appear, which then you can open in a debugger.
When a child process stops or terminates, `SIGCHLD` is sent to the parent process. The default response to the signal is to ignore it. Therefore if you want this crash to be more visible, you'd have to un-ignore SIGCHLD signal. I think checking the status set by the `waitpit()` function and drop a failure message seems neater.

pawosm-arm:
@SouraVX are you OK with this patch?

kiranchandramohan (approved):
LGTM. Please wait for an approval from @SouraVX or @shivaramaarao.

SouraVX:
I noted following behavior:
**Before patch (Signal received by the process):**
```
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_KILLED, si_pid=16277, si_uid=1001, si_status=SIGSEGV, si_utime=1, si_stime=0} 
```
**After patch (Signal received by the process):**
```
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=16696, si_uid=1001, si_status=0, si_utime=16, si_stime=2} ---
```
If that's the intended thing, I'm happy :)
Some tests are still running, is this Okay to wait for 1 more day ? 
Thanks a lot!

pawosm-arm:
Sure it can wait. Thanks for looking at it @SouraVX

SouraVX:
BTW @bryanpkc did you get a chance to have a look ?

SouraVX (approved):
LGTM. Thanks!
Please wait for approval from @bryanpkc  also.

kiranchandramohan:
@SouraVX just want to highlight that technically waiting for @bryanpkc is not required as per the Pull Request process. The PR process only requires that two reviewers test and approve the PR on Aarch64 and AMD. If that is not everyone's understanding then we can discuss the PR processing again.
However it is a good practise to wait for @bryanpkc.

SouraVX:
> @SouraVX just want to highlight that technically waiting for @bryanpkc is not required as per the Pull Request process. The PR process only requires that two reviewers test and approve the PR on Aarch64 and X86_64. If that is not everyone's understanding then we can discuss the PR processing again.
> However it is a good practise to wait for @bryanpkc (or all organizations involved).

Thanks for clarification. No issues from our side :)

runtime/flangrti/llcrit.c
```diff
@@ -63,13 +66,32 @@ static kmp_critical_name nest_sem;
 static omp_nest_lock_t nest_lock;
 
 static int is_init_nest = 0;
+static int is_init_nest_red = 0;
+static int is_atfork_registered = 0;
+
+static void __llcrit_atfork(void)
```
bryanpkc, line 72: Nit: Use PGI coding style?

bryanpkc (approved):
LGTM. Built fine in our code base.

PR #1026, opened 2021-04-08
[DebugInfo] Test case update for #895

SouraVX:
Thanks! @pawosm-arm for the heads up :)


pawosm-arm (approved):
This test indeed verifies the bugfix introduced by PR #895.

PR #1033, opened 2021-04-16
[DebugInfo] Support for `-gpubnames` option in `flang`

SouraVX:
This option is used to generate extra debug information.
i.e
DWARFv4
`flang -gdwarf-4 -gpubnames test.c` will generate `.debug_pubnames` and
`.debug_pubtypes` in the executable.
DWARFv5
`flang -gdwarf-5 -gpubnames test.c` will generate `.debug_names` in
the executable.

Prior to this change these sections were getting generated by default.

These sections were extra and may or maynot be used by the debugger. Hence
generation of these sections are bind to this `-gpubnames` option.
This behavior is analogous to `clang`, `gcc`, gfortran.

PR #1034, opened 2021-04-16
Github Action use the prebuilt clang to build flang

michalpasztamobica:
Previously the native compiler was used, but this resulted in some tests failing for X86 and AArch64.

### DIFF
--- a/.github/workflows/build_flang.yml
+++ b/.github/workflows/build_flang.yml
@@ -9,9 +9,11 @@ on:
 jobs:       
   build_flang:
     runs-on: ubuntu-20.04
+    env:
+      install_prefix: /usr/local
     strategy:
       matrix:
-        target: [X86] # , AArch64, PowerPC]
+        target: [X86]
         cc: [clang]
         cpp: [clang++]
         version: [9, 10, 11]
@@ -42,25 +44,6 @@ jobs:
         with:
           key: ${{ matrix.cc }}-${{ matrix.version }}
 
-      - if: matrix.cc == 'gcc' && matrix.version == '10'
-        run: |
-          sudo add-apt-repository ppa:ubuntu-toolchain-r/ppa
-          sudo apt install gcc-10 g++-10
-          sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 10
-          sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 20
-          sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 10
-          sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-10 20
-      
-      - if: matrix.cc == 'clang'
-        run: |
-          sudo add-apt-repository ppa:ubuntu-toolchain-r/ppa
-          sudo touch /etc/apt/sources.list.d/llvm.list
-          echo 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-11 main' | sudo tee -a /etc/apt/sources.list.d/llvm.list
-          echo 'deb-src http://apt.llvm.org/focal/ llvm-toolchain-focal-11 main'  | sudo tee -a /etc/apt/sources.list.d/llvm.list
-          wget -q -O - http://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
-          sudo apt update
-          sudo apt install -f -y llvm-${{ matrix.version }} clang-${{ matrix.version}}
-
       - name: Check tools
         run: |
           sudo apt update
@@ -68,7 +51,6 @@ jobs:
           git --version
           cmake --version
           make --version
-          ${{ matrix.cc }} --version
 
       # Download artifacts for the classic-flang-llvm-project-related builds (all toolchains)
       - name: Download artifacts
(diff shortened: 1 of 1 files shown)


SouraVX (approved):
LGTM, Thanks for spotting and fixing it :)
Moving All PR's that has dependency on both `driver` and `flang` should benefit from this.

Not to mention, if such a situation occur again as was with https://github.com/flang-compiler/classic-flang-llvm-project/pull/32 and https://github.com/flang-compiler/flang/pull/1033 i.e, developer mentioning that PR's building/validating fine locally but failing in CI, then CI infra should be scrutinized/revisited again.
Does above point make sense to everyone ?

.github/workflows/build_flang.yml
```diff
@@ -104,8 +104,8 @@ jobs:
           CMAKE_OPTIONS="-DLLVM_TARGETS_TO_BUILD=${{ matrix.target }} \
             -DCMAKE_INSTALL_PREFIX=/usr/local \
             -DCMAKE_BUILD_TYPE=Release \
-            -DCMAKE_C_COMPILER=/usr/bin/${{ matrix.cc }} \
-            -DCMAKE_CXX_COMPILER=/usr/bin/${{ matrix.cpp }} \
+            -DCMAKE_C_COMPILER=/usr/local/bin/clang \
+            -DCMAKE_CXX_COMPILER=/usr/local/bin/clang++ \
```
kiranchandramohan, line 108: Does this mean that we will not be testing with gcc anymore?

michalpasztamobica: Yes and no.
Yes - we will not build `flang` with a native Linux compiler (gcc-10, clang-9, 10 or 11)
No - we are still using different compilers to build the llvm/clang in `classic-flang-llvm-project`. We download it as an artifact. So effectively - we still need one build per compiler in the `flang` project.

kiranchandramohan: So, are these following steps required?

```
          sudo apt install gcc-10 g++-10
          sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 10
          sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 20
          sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 10
          sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-10 20
 ```

Comment (michalpasztamobica) [reply .github/workflows/build_flang.yml]: Good point, @kiranchandramohan , I updated the PR removing these steps (and also the clang installation)
@pawosm-arm I used the opportunity to also use an env variable to store `/usr/local` prefix and not copy it all over the script. Thanks for the hint!

**Comment by michalpasztamobica** [comment]
> Not to mention, if such a situation occur again as was with [flang-compiler/classic-flang-llvm-project#32](https://github.com/flang-compiler/classic-flang-llvm-project/pull/32) and #1033 i.e, developer mentioning that PR's building/validating fine locally but failing in CI, then CI infra should be scrutinized/revisited again.
> Does above point make sense to everyone ?

@SouraVX , answering your question - I fully agree. The CI's script is public and always open for a revisit :). 
Sorry that this bug slowed down your PR, it silently lived with us for 6 months without manifesting itself. 
I hope CI will be fully reliable now, but if you see any further suspicious places, please let us know.

**Comment by kiranchandramohan** [comment]
Whater are these versions in the matrix? How is it different from the llvm_branch? I am OK with just having llvm-11 and llvm-12. I believe @bryanpkc has moved to 12. What about @shivaramaarao / @SouraVX ?

```
        version: [9, 10, 11]
        llvm_branch: [release_100, release_11x, release_12x]
        include:
          - target: X86
            cc: gcc
            cpp: g++
            version: 10
            llvm_branch: release_100
          - target: X86
            cc: gcc
            cpp: g++
            version: 10
            llvm_branch: release_11x
          - target: X86
            cc: gcc
            cpp: g++
            version: 10
            llvm_branch: release_12x
  ```

michalpasztamobica:
@kiranchandramohan , the difference in how matrices look like in `classic-flang-llvm-project` and in here is caused by the fact that for `classic-flang-llvm-project` we have branches and each branch builds 5 configs (using llvm-9,10,11 and gcc-10 to bootstrap). Then for `flang` we only have `master` branch, so we pass the the branch as a parameter, but we should still have the same number of configs

I now noticed that we dropped gcc-9 (along with the support for release_90), but we did not drop it from `classic-flang-llvm-project`. We have a redundant config to remove in there. I'll take care of this.

bryanpkc:
@kiranchandramohan I would like to keep building/testing flang with release_100. Thanks.

bryanpkc (approved):
LGTM. Thanks @michalpasztamobica.

SouraVX:
At the moment we also build & test llvm_10, so would like to keep it. Thanks!

kiranchandramohan (approved):
LGTM. Let me know if this is ready to merge.

SouraVX (approved):
LG, Thanks!

michalpasztamobica:
@kiranchandramohan , please go ahead with merging.

kiranchandramohan:
@michalpasztamobica The post commit CI failed.
https://github.com/flang-compiler/flang/actions/runs/762633691

michalpasztamobica:
@kiranchandramohan , the failure is at artifacts upload at some TLS action, we did not touch this part in this PR... Would you please try to trigger a rebuild of the job? I will try this out on my fork as well.

michalpasztamobica:
Looks like the artifacts upload worked fine after retriggering: https://github.com/flang-compiler/flang/runs/2378515711
It must have been some glitch in Github's connectivity...

PR #1035, opened 2021-04-21
Drop building with llvm-9 and adjust the number of artifacts from classic-flang-llvm-project

michalpasztamobica:
The first change results from https://github.com/flang-compiler/classic-flang-llvm-project/pull/33.
The second change adjusts the expected number of artifacts to 3. We used to have every branch built with llvm-9,10,11 and gcc-10,11 (total 5), but now we will only build llvm-10,11 and gcc-10.

I considered dropping this check entirely, but then I thought we should really have some means of skipping any kind of incorrect builds from the `classic-flang-llvm-project` repository, so that a CI failure there does not instantly break the CI here. Right now we allow 10 broken runs of llvm CI before this repo is affected. The easiest way to recognize a broken build is that it doesn't have enough artifacts.


pawosm-arm (approved):
ok

bryanpkc (approved):
LGTM.

michalpasztamobica:
@kiranchandramohan , would you be able to re-run these jobs now, that the dependent PR is merged and a build with 3 artifacts is ready in `classic-flang-llvm-project`?

kiranchandramohan:
Rerunning.

michalpasztamobica:
Sorry, I realised I need to apply the same changes to other branches of `classic-flang-llvm-project` to get all branches to build correctly:
https://github.com/flang-compiler/classic-flang-llvm-project/pull/34
https://github.com/flang-compiler/classic-flang-llvm-project/pull/35
I don't think we need another round of reviews, this is the same commit cherry-picked to other branches.

michalpasztamobica:
@SouraVX , @shivaramaarao , @mleair , would you have a moment to give your reviews, please?

SouraVX (approved):

PR #1038, opened 2021-04-22
Fix error of outputting derived type.

yaoliu-hnc:
The test case is as following :
```fortran
program test
  type my_type
     integer :: i
     integer :: j
     integer :: k
  end type my_type
  character(80) :: str1, str2
  type(my_type), dimension(3) :: t

  t(1) = my_type(11, 12, 13)
  t(2) = my_type(21, 22, 33)
  t(3) = my_type(31, 32, 33)

  print *, t(func())

contains
  integer function func()
    integer, save :: i = 1
    func = i
    i = i + 1
  end function
end
```
The correct result should be:
 `         11          12          13`
But flang print incorrect result:
 `         11          22          33`
Flang calls incorrectly `func()` three times and prints the values of `t(1)%i`, `t(2)%j` and `t(3)%k`.

### DIFF
--- a/test/f90_correct/src/io28.f90
+++ b/test/f90_correct/src/io28.f90
@@ -1,0 +1,33 @@
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Test for I/O take derived type array reference.
+! The function should be called only one time if a function
+! reference appears in the derived type array reference.
+
+program test
+  type my_type
+     integer :: i
+     integer :: j
+  end type my_type
+  character(80) :: str1(2), str2(2)
+  type(my_type) :: t(2)
+
+  t(1) = my_type(11, 12)
+  t(2) = my_type(21, 22)
+
+  write(str1(1), *), t(1)
+  write(str1(2), *), t(2)
+  write(str2(1), *), t(func())
+  write(str2(2), *), t(func())
+  if (any(str1 .ne. str2)) STOP 1
+  write(*, *) 'PASS'
+
+contains
+  integer function func()
+    integer, save :: i = 1
+    func = i
+    i = i + 1
+  end function
+end
--- a/tools/flang1/flang1exe/semantio.c
+++ b/tools/flang1/flang1exe/semantio.c
@@ -2895,7 +2895,12 @@ semantio(int rednum, SST *top)
             CNULL);
     }
     if (DTYG(SST_DTYPEG(RHS(1))) == TY_DERIVED &&
-        A_TYPEG(SST_ASTG(RHS(1))) == A_FUNC) {
+        (A_TYPEG(SST_ASTG(RHS(1))) == A_FUNC ||
+        /* Allocate a temporary ast to store the value of the derived type array
+         * reference whose subscript is a function reference, otherwise the
+         * function would be incorrectly called in each component I/O. */
+        (A_TYPEG(SST_ASTG(RHS(1))) == A_SUBSCR &&
+         A_CALLFGG(SST_ASTG(RHS(1)))))) {
       ast = sem_tempify(RHS(1));
       (void)add_stmt(ast);
       SST_IDP(RHS(1), S_IDENT);
--- a/test/f90_correct/inc/io28.mk
+++ b/test/f90_correct/inc/io28.mk
@@ -1,0 +1,22 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+$(TEST): run
+
+
+build:  $(SRC)/$(TEST).f90
+	-$(RM) $(TEST).$(EXESUFFIX) core *.d *.mod FOR*.DAT FTN* ftn* fort.*
+	@echo ------------------------------------ building test $@
+	-$(FC) -c $(FFLAGS) $(LDFLAGS) $(SRC)/$(TEST).f90 -o $(TEST).$(OBJX)
+	-$(FC) $(FFLAGS) $(LDFLAGS) $(TEST).$(OBJX) $(LIBS) -o $(TEST).$(EXESUFFIX)
+
+
+run:
+	@echo ------------------------------------ executing test $(TEST)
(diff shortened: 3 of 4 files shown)


tools/flang1/flang1exe/semantio.c
```diff
@@ -2895,7 +2895,10 @@ semantio(int rednum, SST *top)
             CNULL);
     }
     if (DTYG(SST_DTYPEG(RHS(1))) == TY_DERIVED &&
-        A_TYPEG(SST_ASTG(RHS(1))) == A_FUNC) {
+        (A_TYPEG(SST_ASTG(RHS(1))) == A_FUNC ||
+        /* Fix bug: if a function reference appears in the array reference. */
```
bryanpkc, line 2899: Could you rewrite this comment in a way that explains what the whole conditional block is doing, and why the two different types of checks are needed?

test/f90_correct/src/io28.f90
```diff
@@ -0,0 +1,38 @@
...
+! test for I/O take derived types
+
+program test
+  type my_type
+     integer :: i
+     integer :: j
+     integer :: k
+  end type my_type
+  character(80) :: str1, str2
+  type(my_type), dimension(3) :: t
```
bryanpkc, line 14: Why do you need three elements in this array? Wouldn't just two suffice to show the problem?

test/f90_correct/src/io28.f90
```diff
@@ -0,0 +1,38 @@
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! test for I/O take derived types
```
bryanpkc, line 5: Please elaborate on what this test case is checking.

bryanpkc (changes requested):
Please reword the commit message to explain what bug you are fixing, and what part of the code is being modified.

test/f90_correct/src/io28.f90
yaoliu-hnc: Flang would call incorrectly `func()` tree time and write the value of `t(1)%i`, `t(2)%j` and `t(3)%k` to `str2`, so I have declared three elements in this array. But it need only one element if the issue is fixed.

yaoliu-hnc: > Why do you need three elements in this array? Wouldn't just two suffice to show the problem?

Flang would call incorrectly func() tree time and write the value of t(1)%i, t(2)%j and t(3)%k to str2, so I have declared three elements in this array. But it need only one element if the issue is fixed.

tools/flang1/flang1exe/semantio.c
yaoliu-hnc: > Could you rewrite this comment in a way that explains what the whole conditional block is doing, and why the two different types of checks are needed?

`/* Allocate a temporary ast to store the value of the derived type array
         * reference whose subscript is a function reference, otherwise the
         * function would be incorrectly called in each component I/O. */`
 The comment has been rewritten in the new commit.

test/f90_correct/src/io28.f90
yaoliu-hnc: > Please elaborate on what this test case is checking.

The comment has been rewritten in the new commit.

test/f90_correct/src/io28.f90
bryanpkc: OK thanks.

shanksjoe: Can you check the simplified test case as the following?
```
program test
  type my_type
     integer :: i
     integer :: j
  end type my_type
  character(80) :: str1(2), str2(2)
  type(my_type) :: t(2)

  t(1) = my_type(11, 12)
  t(2) = my_type(21, 22)

  write(str1(1), *), t(1)
  write(str1(2), *), t(2)
  write(str2(1), *), t(func())
  write(str2(2), *), t(func())
  if (any(str1 .ne. str2)) STOP 1
  write(*, *) 'PASS'

contains
  integer function func()
    integer, save :: i = 1
    func = i
    i = i + 1
  end function
end
```

shanksjoe:
Please squash two commits into one since the two commits only fix one problem. In addition, please explain the detailed bug info in commit message.

test/f90_correct/src/io28.f90
yaoliu-hnc: OK. Thank you.

yaoliu-hnc:
@bryanpkc The commit has been modified. Please have a look.

bryanpkc (approved):
LGTM.

pawosm-arm (approved):
Using parentheses more generously would make it easier to digest. Otherwise, it looks good.

test/f90_correct/src/io28.f90
```diff
@@ -0,0 +1,33 @@
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Test for I/O take derived type array reference.
+! The function should be called only one time if a function
+! reference appears in the derived type array reference.
+
+program test
```
SouraVX, line 9: Is it possible to make reduce/shorten this test case ?

yaoliu-hnc: These test cases have been simplified once. I don't think they need to be reduced.

SouraVX: @yaoliu-hnc for clarifying.
CI still fails to build the PR, any reasons ?

SouraVX:
@yaoliu-hnc despite CI failures, PR works well when I tried it locally :)
Some tests are running I'll approve this soon.

yaoliu-hnc:
> @yaoliu-hnc despite CI failures, PR works well when I tried it locally :)
> Some tests are running I'll approve this soon.

We modified the makefile of the testcase. Could you approve this for another CI check?
Thanks.

SouraVX:
> > @yaoliu-hnc despite CI failures, PR works well when I tried it locally :)
> > Some tests are running I'll approve this soon.
> 
> We modified the makefile of the testcase. Could you approve this for another CI check?
> Thanks.
 
@kiranchandramohan this is for you :)
```
1 workflow awaiting approval
First-time contributors need a maintainer to approve running workflows.
```
We could modify this rule ? WDYT ?

SouraVX (approved):
LG, Thanks!

PR #1039, opened 2021-04-22
Fix power operation optimization when the left operand is -1 and the right is a negative integer.

yaoliu-hnc:
The test case is following:
```fortran
program test
  integer :: i1, i2, res(2)
  i1 = -1
  i2 = -8
  res(1) = i1 ** i2
  res(2) = i1 ** (-7)
  print *, res
end
```
the result with `-O0` and `-O1` is correct:
 `           1            -1`
the result with `-O2`, `-O3`, and `-Os` is incorrect:
 `           0            0`

### DIFF
--- a/test/f90_correct/src/apower.f90
+++ b/test/f90_correct/src/apower.f90
@@ -1,0 +1,30 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! This test case is test for power operator (int precision).
+! The result of power using base -1 depends on the parity of the exponent.
+
+program test
+  integer :: i1, i2, res(6)
+  integer(kind = 8) :: i2_8
+  integer :: eres(6) = [1, 1, -1, -1, 0, 0]
+
+  i2 = -8
+  i1 = -1
+  res(1) = i1 ** i2
+  res(2) = i1 ** (-8)
+
+  i2 = -7
+  i1 = -1
+  res(3) = i1 ** i2
+  res(4) = i1 ** -7
+
+  i2 = -7
+  i1 = -2
+  res(5) = i1 ** i2
+  res(6) = i1 ** -7
+
+  call check(res, eres, 6)
+end
--- a/tools/flang1/flang1exe/ast.c
+++ b/tools/flang1/flang1exe/ast.c
@@ -918,8 +918,12 @@ mk_binop(int optype, int lop, int rop, DTYPE dtype)
           INT v;
           v1 = CONVAL2G(A_SPTRG(lop));
           v2 = CONVAL2G(A_SPTRG(rop));
-          if (v2 < 0)
+          if (v2 < 0) {
+            if (v1 == -1)
+              /* if v2 is odd number, the result will be -1 */
+              return (v2 & 1) ? mk_cval((INT)-1, DT_INT) : astb.i1;
             return astb.i0;
+          }
           v = v1;
           while (--v2 > 0)
             v *= v1;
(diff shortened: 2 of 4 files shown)


bryanpkc:
The changes seem correct to me, but could you explain why the problem only happens with `-O2`, `-O3`, and `-Os`? `mk_binop` is called all the time, I thought.

It would be nice to have a description about what you are trying to fix in the pull request.

yaoliu-hnc:
> The changes seem correct to me, but could you explain why the problem only happens with `-O2`, `-O3`, and `-Os`? `mk_binop` is called all the time, I thought.

Yes，`mk_binop` is called all the time, but the code changed in the PR would only be called in propagation constant optimization from const_prop () with `-O2`, `-O3`, and `-Os`, which causes the issue.

test/f90_correct/src/apower.f90
```diff
@@ -0,0 +1,58 @@
...
+
+  i1 = 2
+  res(1) = i1 ** i2
+  res(2) = i1 ** (-8)
+  i1 = -2
+  res(3) = i1 ** i2
+  res(4) = i1 ** (-8)
+  i1 = -1
+  res(5) = i1 ** i2
+  res(6) = i1 ** (-8)
```
kiranchandramohan, line 22: When I try with optimisations, only these two fail. Everything else matches. 
Should there be better tests?

tools/flang1/flang1exe/ast.c
```diff
@@ -918,8 +918,13 @@ mk_binop(int optype, int lop, int rop, DTYPE dtype)
           INT v;
           v1 = CONVAL2G(A_SPTRG(lop));
           v2 = CONVAL2G(A_SPTRG(rop));
-          if (v2 < 0)
+          if (v2 < 0) {
+            if (v1 == -1 && (v2 & 1))
```
kiranchandramohan, line 922: Can you add a comment here saying if v2 is odd then the result will be -1. Can you also add a test with an odd number which is not -1?

kiranchandramohan (changes requested):
Can you update the title or description to specify that this is for integers only?

Should/Will this work with the i8 (default-integer-8) or with fixed widths?

kiranchandramohan:
> Yes，`mk_binop` is called all the time, but the code changed in the PR would only be called in propagation constant optimization from const_prop () with `-O2`, `-O3`, and `-Os`, which causes the issue.

Should the fix be in const_prop?

bryanpkc:
> Should the fix be in const_prop?

It seems that `const_prop` triggers the condition for the given test cases, but there could be another potential code path which passes an integer `OP_XTOI` to this function. IMO the optimization itself should be made correct for all inputs.

test/f90_correct/src/apower.f90
yaoliu-hnc: The testcase has been modified to be more readable and clear.

tools/flang1/flang1exe/ast.c
yaoliu-hnc: The comment has been added in the new commit. Testcase of odd base with exponent which is not -1 has also been added.

yaoliu-hnc:
> Can you update the title or description to specify that this is for integers only?
> 
> Should/Will this work with the i8 (default-integer-8) or with fixed widths?

The description has been updated. It works fine with i8 as expected. 
In mk_binop, int8 and int4 are handled differently. The test case that triggered the bug is handled by calling __mth_i_kpowk in the runtime library under the -fdefault-integery-8 option. This fix does not affect int8.
And the test for i8 is also included in our testcase.

yaoliu-hnc:
> > Yes，`mk_binop` is called all the time, but the code changed in the PR would only be called in propagation constant optimization from const_prop () with `-O2`, `-O3`, and `-Os`, which causes the issue.
> 
> Should the fix be in const_prop?

Under O2/O3/OS optimization, const_prop overwrites the AST node and converts the exponent or base to a constant. Calling mk_binop during the overwriting process leads to error, so we fix it in mk_binop.

yaoliu-hnc:
@kiranchandramohan The commit has been modified. Please have a look.

pawosm-arm (approved):

bryanpkc (approved):

kiranchandramohan:
ping @shivaramaarao @SouraVX

shivaramaarao (approved):
LGTM

PR #1044, opened 2021-05-18
Add ARM64 CI

michalpasztamobica:
As discussed in the Wednesday call and as a follow-up of https://github.com/flang-compiler/classic-flang-llvm-project/pull/39 and  https://github.com/flang-compiler/classic-flang-llvm-project/pull/41 we want to use the docker image created and pushed to ghcr.io in `classic-flang-llvm-project` to safely build and test PRs on Aarch64 machines provided by Arm.

Docker and Github Actions scripts are running on a non-root user account and within the docker instance we also provide unprivileged accounts, to ensure as much safety as possible (Github explicitly warns against someone trying to build malicious code in a PR to public repos).

Two servers are provided, which should easily suffice for the needs of Flang project. One build&test takes about 3 minutes.

### DIFF
--- a/.github/workflows/build_flang.yml
+++ b/.github/workflows/build_flang.yml
@@ -3,8 +3,12 @@ name: Flang build & test
 on:
   push:
     branches: [ master ]
+    paths-ignore:
+      - '**/.github/workflows/build_flang_arm64.yml'
   pull_request:
     branches: [ master ]
+    paths-ignore:
+      - '**/.github/workflows/build_flang_arm64.yml'
 
 jobs:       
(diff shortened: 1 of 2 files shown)


michalpasztamobica:
CI has failed, because the `release_11x` docker images do not exist until we merge the other PR in `classic-flang-llvm-project`, however I can see that for `release_12x` the docker image download started fine.

I am setting up the reviewers already in order to 1) give you the full picture of where this is going and 2) speed up the review process.

pawosm-arm (approved):

PR #1045, opened 2021-05-20
win: fix the build on Windows

kaadam:
Pthread is not available on Windows so that is necessary to be guarded.
Therefore we need to add a workaround for those functions which rely on it.
Currently this PR aims to define atfork routine as stub function until
we figure out how it can be fixed.

### DIFF
--- a/runtime/flangrti/llcrit.c
+++ b/runtime/flangrti/llcrit.c
@@ -9,7 +9,13 @@
 #include <stdlib.h>
 #include <stdint.h>
 #include <string.h>
+#if !defined(_WIN32)
 #include <pthread.h>
+#else
+// FIXME: fork() is not available on Windows. So atfork routine can be a simple stub.
+// until it is fixed properly on Windows.
+#define pthread_atfork(F1,F2,F3) 0
+#endif
 #include <stdioInterf.h>
 #include "komp.h"
 


runtime/flangrti/llcrit.c
```diff
@@ -9,7 +9,13 @@
 #include <stdlib.h>
 #include <stdint.h>
 #include <string.h>
+#if !defined(_WIN32)
```
pawosm-arm, line 12: pthread is indeed not in windows when you install it, but it doesn't mean there is no 3rd party implementations available, I've been using some of them in my past jobs to build some of the open source projects, with success.
Nevertheless, if Windows port can live without this header, then fine, but then, do we still need to refer 32-bit windows in our ifdefs? AFAIR there are new preprocessor definitions to indicate we're building for Windows...

kaadam: Yes, you're right, pthread 'is not available' doesn't mean that there is no other third-party implementation ( https://sourceware.org/pthreads-win32 ). It only means Windows itself doesn't natively support pthreads. For example pthread_atfork is not implemented in pthread-win32 library,  and it do not correspond well to the Microsoft windows API. Maybe we can use 'CreateProcess' Windows function to way around this later.

Regarding to the predefined macro I haven't seen any new Microsoft specific predefined macros.  In the documentation _WIN32 is generally used for detecting: "Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, x86, or x64. Otherwise, undefined." 
See: https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160#microsoft-specific-predefined-macros

What about that: 
1) We defer it to CMake and add some user defined macros based on how CMake detects the OS and Arch. 
2) We add something like this (build_config.h) to Flang
#if defined(_WIN32)
...
#define OS_WIN 1
...
#endif
I prefer to defer these things to the compiler, but it's a matter of taste. Shall I open an issue for the further discussion?

pawosm-arm: As I mentioned in some other review, I'm not working with Windows for years, I don't know what current recommendations are, apparently, the 64-bit specific macros I knew about are in Clang on Windows, but not in MSVC. I'd rather see an opinion of the others (potentially, more eager Windows users) than start a new discussion. For this review, I'll follow the approval decision of the majority.

pawosm-arm (approved):

kaadam:
@kiranchandramohan @xoviat Could you take a look at this? Thanks.

xoviat:
Personally, I don't like this. Windows doesn't support forking (AFAIK) which means that there shouldn't be a 'fork' handler registered on windows. It's not a case where the winapi can be wrapped, because there isn't a windows equivalent. Where is this function used?

bryanpkc:
Instead of defining a macro to override `pthread_atfork`, I would prefer a more explicit `#ifndef _WIN32` around its uses in `_mp_bcs_nest` and `_mp_bcs_nest_red`.

xoviat:
See #1084

kiranchandramohan:
@kaadam What is the next step here? Are u going to update or close this?

xoviat:
@kiranchandramohan #1084 supersedes this.

kiranchandramohan:
Can you close this once you approve 1084?

kaadam:
@kiranchandramohan @xoviat I close this PR because of #1084.

PR #1049, opened 2021-05-25
Fix the bug that NEAREST returns wrong value for +/- infinity.

yaoliu-hnc:
When the 1st parameter is positive or negative infinity, the intrinsic function `NEAREST` wrongly returns the same value while it should return the positive or negative number of the largest magnitude accordingly when the 2nd parameter has the different sign.

This patch fixes this bug by adding logic to handle the aforementioned situations for all single-, double-precision fp numbers.

### DIFF
--- a/runtime/flang/miscsup_com.c
+++ b/runtime/flang/miscsup_com.c
@@ -4544,6 +4544,10 @@ ENTF90(NEARESTX, nearestx)(__REAL4_T f, __LOG_T sign)
       ++x.i;
     else
       --x.i;
+  } else if (!(sign & GET_DIST_MASK_LOG) && (x.i == 0x7F800000)) {
+    x.i = 0x7F7FFFFF;
+  } else if ((sign & GET_DIST_MASK_LOG) && (x.i == 0xFF800000)) {
+    x.i = 0xFF7FFFFF;
   }
   return x.f;
 }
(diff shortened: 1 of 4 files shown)


runtime/flang/miscsup_com.c
```diff
@@ -4544,6 +4544,11 @@ ENTF90(NEARESTX, nearestx)(__REAL4_T f, __LOG_T sign)
       ++x.i;
     else
       --x.i;
+  } else if ((x.i & 0x7FFFFFFF) == 0x7F800000) {
+    if (~sign && (x.i & 0xFFFFFFFF) == 0x7F800000)
```
bryanpkc, line 4548: Is there any reason for testing `(x.i & 0x7FFFFFFF) == 0x7F800000` just before testing `(x.i & 0xFFFFFFFF) == 0x7F800000`? It does not seem necessary to group the two conditions in one `else if` block.

bryanpkc: Also, is it correct to use `~sign` instead of `!(sign & GET_DIST_MASK_LOG)`?

runtime/flang/miscsup_com.c
```diff
@@ -4563,13 +4568,16 @@ ENTF90(NEARESTDX, nearestdx)(__REAL8_T d, __LOG_T sign)
...
-    if ((x.ll >> 52 & 0x7FF) != 0x7FF) { /* not nan or inf */
+  } else if ((x.ll >> 52 & 0x7FF) != 0x7FF) { /* not nan or inf */
       if ((x.d < 0) ^ (sign & GET_DIST_MASK_LOG))
         ++x.ll;
       else
         --x.ll;
-    }
+  } else {
+    if (~sign && (x.ll & ~0x0) == 0x7FF0000000000000)
+      x.ll = 0x7FEFFFFFFFFFFFFF;
```
bryanpkc, line 4578: Here as well, consider breaking this `else` block into two `else if` blocks.

qiaozhang-hnc: I've modified the code as suggested. Thanks!

yaoliu-hnc: I've modified the code as suggested. Thanks!

bryanpkc (approved):
LGTM.

pawosm-arm (approved):

shivaramaarao (approved):
LGTM

kiranchandramohan:
@bryanpkc I think this is ready for merging.

PR #1050, opened 2021-05-26
Fix improper handling of errors caused by invalid declaration of result variable

yaoliu-hnc:
This patch fixes the following bugs:
- When declaring the result variable with the `PARAMETER` attribute, flang doesn't raise an error, causing segmentation fault downstream.
- When initializing the result variable at the time of declaration, flang reports a wrong error.
- When the result variable collides with the name of a derived type, flang doesn't raise an error, causing ICE downstream.


bryanpkc (approved):
LGTM

pawosm-arm (approved):

shivaramaarao (approved):
LGTM

PR #1051, opened 2021-05-27
Add reporting for two grammar errors pertaining to PROTECTED attribute.

qiaozhang-hnc:
A syntactical error should be reported when a dummy argument is specified with `INTENT(out)` or `INTENT(inout)` while the actual argument has the attribute `PROTECTED`.

A syntactical error should be reported when modifying a `PROTECTED` pointer variable.

This patch catches and reports 2 new grammar errors related to the `PROTECTED` parameter (and extends the buffer for the error message):
1.A use-associated object with the `PROTECTED` attribute cannot be an actual argument when the dummy argument
2.A use-associated object with the `PROTECTED` attribute cannot be assigned

### DIFF
--- a/test/f90_correct/inc/protected.mk
+++ b/test/f90_correct/inc/protected.mk
@@ -1,0 +1,15 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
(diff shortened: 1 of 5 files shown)


bryanpkc:
@qiaozhang-hnc  Could you please rebase this branch?

qiaozhang-hnc:
> @qiaozhang-hnc Could you please rebase this branch?

Rebased.

bryanpkc (approved):
LGTM

pawosm-arm (approved):

PR #1052, opened 2021-05-27
Fix the bug where the name of ENTRY is used as an argument

qiaozhang-hnc:
Flang doesn’t catch and handle the syntax error that the name of an entry point is used as an argument when a result clause is present, ultimately causing a segmentation fault in flang2. This patch fixes the issue by catching and reporting the syntax error described above. It then restores the right state.

### DIFF
--- a/tools/flang1/flang1exe/semant3.c
+++ b/tools/flang1/flang1exe/semant3.c
@@ -542,6 +542,13 @@ semant3(int rednum, SST *top)
       if (CLASSG(sptr) && !MONOMORPHICG(sptr) && !ALLOCATTRG(sptr)) {
         error(1217, ERR_Severe, gbl.lineno, SYMNAME(sptr), CNULL);
       }
+      if (STYPEG(sptr) == ST_ENTRY && FVALG(sptr)) {
+        char buffer[256];
+        (void)snprintf(buffer, sizeof(buffer), "- should use %s",
+                       SYMNAME(FVALG(sptr)));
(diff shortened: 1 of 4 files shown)


pawosm-arm (approved):

pawosm-arm:
@reviewers what about this one?

bryanpkc (approved):

bryanpkc:
We are using this in our downstream compiler already. @shivaramaarao Could you review and approve?

tools/flang1/flang1exe/semant3.c
```diff
@@ -541,6 +541,10 @@ semant3(int rednum, SST *top)
       if (CLASSG(sptr) && !MONOMORPHICG(sptr) && !ALLOCATTRG(sptr)) {
         error(1217, ERR_Severe, gbl.lineno, SYMNAME(sptr), CNULL);
       }
+      if (STYPEG(sptr) == ST_ENTRY && FVALG(sptr)) {
+        error(72, 3, gbl.lineno, SYMNAME(sptr), "- should use result-name");
```
shivaramaarao, line 545: can we add actual variable name instead of result-name here?. 

 if not possible,  can we remove  "should use result-name"

error message will be 
PGF90-S-0072-Assignment operation illegal to ent

qiaozhang-hnc: This patch has been updated and we can report errors with the actual variable name now.

shivaramaarao (approved):

PR #1053, opened 2021-05-27
Fix the bug that the GETARG(n,bf) returns a wrong result for int1/int2

qiaozhang-hnc:
The problem occurs because the argument `n` is treated as a C int pointer and is directly dereferenced regardless of the kind of the integer. This patch fixes the error by fetching the scalar integer value with respect to both the address and the size.


qiaozhang-hnc:
Could anyone give me the approval to run workflows? @pawosm-arm @bryanpkc @kiranchandramohan (and other maintainers I don't @ here). Thanks!

tools/shared/rtlRtns.h
```diff
@@ -125,6 +125,7 @@ typedef enum {
   RTE_fracd,
   RTE_fracdx,
   RTE_fracx,
+  RTE_getarga,
```
bryanpkc, line 128: What's the significance of the F90 group versus the FTN group of runtime function symbols? Putting this symbol here names it as `f90_getarga_i8` instead of `fort_getarga_i8`. I am not certain if there is any practical difference; I am just curious since `GETARG` is a Fortran 77 intrinsic and does not appear in later revisions of the language standard.

bryanpkc: @mleair Perhaps you might know the answer to my question?

tools/flang1/flang1exe/semfunc.c
```diff
@@ -11474,6 +11474,36 @@ ref_pd_subr(SST *stktop, ITEM *list)
...
+        goto call_e74_arg;
+      }
+      (void)mkarg(sp, &dum);
+      XFR_ARGAST(1);
+      dtype2 = SST_DTYPEG(sp);
+      if (DTY(dtype2) != TY_CHAR) {
+        E74_ARG(pdsym, 1, NULL);
+        goto call_e74_arg;
+      }
+    }
```
bryanpkc, line 11504: The processing of the first two arguments is identical to that in `case PD_get_command_argument`. Perhaps some code reuse is possible.

qiaozhang-hnc: Do you mean to refactor the similar blocks out as a function?

qiaozhang-hnc: Also, I'm not quite sure what is the exact requirement for the 1st arg of `GETARG`. In gfortran, it is illegal for the 1st argument to be of an integer kind wider than the default kind, while in ifort, it is ok. Do you have any suggestions?

tools/shared/rtlRtns.h
mleair: Mostly this is historical as we used to have F77, HPF, and F90 runtime routines that each had their own prefix of pgftn, pghpf, and pgf90 respectively. I do not believe it matters much anymore what the prefix is. When we open sourced PGF90, we renamed the prefixes to vendor neutral names.

runtime/flang/getarg.c
```diff
@@ -27,13 +27,14 @@ __INT_T
 ENTRY(IARGC, iargc)() { return __io_get_argc() - 1; }
 
 void
-ENTRY(GETARGA, getarga)(__INT_T *n, DCHAR(arg) DCLEN64(arg))
+ENTF90(GETARGA, getarga)(__INT_T *n, DCHAR(arg), __INT_T *int_kind DCLEN64(arg))
```
mleair, line 30: Do you really need to change the prefix?

Also, changing the prefix and/or arguments of an existing runtime routine will break backward compatibility. Why not create a new runtime routine (e.g., GETARGA2) that takes care of this case? That way, you will not potentially break someone's existing code.

Or if you prefer, rename this routine to something like GETARGA2 and keep ENTRY(GETARGA) but use it as a wrapper to GETARGA2().

mleair, line 30: Disregard my previous comment. However, what is your reason the for the switch from ENTRY to ENTRYF90?

qiaozhang-hnc: To pass the additional argument `int_kind` to the runtime, this patch creates a predeclared subroutine for `GETARG`, and as a result, the prefix `ENTRY` cannot be used anymore. `ENTF90` is used here to mimic the code for `GET_COMMAND_ARGUMENT`.

mleair (commented):
LGTM

tools/flang1/flang1exe/semfunc.c
bryanpkc: `GETARG` was not a standard Fortran intrinsic; vendors would define it differently. For its first argument, it was mostly declared to be the default `integer` (`integer` in [HP Fortran](http://binf.gmu.edu/jafri/HP_Fortran_Programmers_Reference.pdf), `integer*4` in [Sun Studio](https://docs.oracle.com/cd/E19205-01/819-5259/aetdh/index.html), `integer*2` in [Microsoft Fortran](http://computer-programming-forum.com/49-fortran/c2241492a04c9fe4.htm), etc.). I think gfortran accepts all `integer` types narrower than the default `integer` simply to support more legacy programs. The Intel behaviour is probably an extension, and I am not sure that it is all that valuable.

bryanpkc: About code reuse, yes, I thought you could factor out the blocks into a function. You could use labels and `goto` which is fairly common in the code base, but I don't recommend that practice any more. Leaving it as is fine as well.

bryanpkc: Never mind, I think factoring the code is probably not worthwhile.

bryanpkc (approved):
LGTM.

kiranchandramohan:
@qiaozhang-hnc Can you rebase?

tools/flang1/flang1exe/func.c
```diff
@@ -2819,9 +2819,13 @@ rewrite_func_ast(int func_ast, int func_args, int lhs)
     }
     is_icall = FALSE;
     goto ret_call;
+  case I_GETARG:
   case I_GET_COMMAND:
   case I_GET_COMMAND_ARGUMENT:
-    if (optype == I_GET_COMMAND) {
+    if (optype == I_GETARG) {
+      rtlRtn = RTE_getarga;
```
kiranchandramohan, line 2826: Does this change mean that `getarg` won't be called anymore?

tools/flang1/flang1exe/semfunc.c
```diff
@@ -11474,6 +11474,37 @@ ref_pd_subr(SST *stktop, ITEM *list)
     REFP(pdsym, 1);
     break;
 
+  case PD_getarg:
+    if (count != 2) {
+      E74_CNT(pdsym, count, 2, 2);
```
kiranchandramohan, line 11479: Nit:Can we have a test for this and other semantic checks?

kiranchandramohan (changes requested):
An ILM/ILI or even an llvm IR test to demonstrate the changed call to `@getarga` would be good.

qiaozhang-hnc:
> @qiaozhang-hnc Can you rebase?

OK, already rebased.

tools/flang1/flang1exe/func.c
qiaozhang-hnc: Yes, we hope to call `getarga` instead of `getarg`

qiaozhang-hnc: > Does this change mean that `getarg` won't be called anymore?

Yes, we hope to call `getarga` instead of `getarg`

qiaozhang-hnc:
@kiranchandramohan LLVM IR test and semantic checks have been added.

runtime/flang/getarg.c
```diff
@@ -52,9 +53,9 @@ ENTRY(GETARGA, getarga)(__INT_T *n, DCHAR(arg) DCLEN64(arg))
 
 /* 32 bit CLEN version */
 void
-ENTRY(GETARG, getarg)(__INT_T *n, DCHAR(arg) DCLEN(arg))
+ENTF90(GETARG, getarg)(__INT_T *n, DCHAR(arg), __INT_T *int_kind DCLEN(arg))
 {
-  ENTRY(GETARGA, getarga)(n, CADR(arg), (__CLEN_T)CLEN(arg));
+  ENTF90(GETARGA, getarga)(n, CADR(arg), int_kind, (__CLEN_T)CLEN(arg));
 }
```
kiranchandramohan, line 59: Is this function needed anymore? Can it be removed, if there are no calls to it?

kiranchandramohan (commented):
> @kiranchandramohan LLVM IR test and semantic checks have been added.

Thanks, just one more question.

runtime/flang/getarg.c
qiaozhang-hnc: As 1f4b8f14   committed by @mleair shows :

> Intrinsic routines that used to take a 32-bit CHARACTER length have been updated to the __CLEN_T type. In order to keep backwards compatibility we need to keep the old interface, and create a new routine with a new interface. The convention chosen is to append an "a" to the new routine.

Should we remove the old interface `GETARG`?

bryanpkc: My two cents: 1f4b8f1 added wrappers for a large number of runtime functions. The compiler should generate calls to the new versions of the functions when compiling new programs. I believe that the backward compatibility had to be maintained for the benefit of old programs that needed to work a new version of the runtime library without recompilation. The backward compatibility may be valuable for certain vendors/customers and we should probably discuss more carefully about how to deprecate the old symbols.

kiranchandramohan: OK. Fine with me.

kiranchandramohan (approved):
LGTM.

shivaramaarao:
> 



> @shivaramaarao Did you want to review this as well?

LGTM from my side as well. For some reason, I am not  seeing the option to approve this review. This can be merged.

shivaramaarao:
> 



> @shivaramaarao Did you want to review this as well?

LGTM from my side as well. For some reason, I am not  seeing the option to approve this review. This can be merged.

bryanpkc:
Thanks!

PR #1054, opened 2021-05-27
Revert "Revert "Add ARM64 CI""

michalpasztamobica:
Build and test Flang for Aarch64/ARM64 target using pre-build docker images with appropriate llvm and clang from classic-flang-llvm-project.

I wasn't sure how to best avoid confusion, as I am in fact resubmitting this after https://github.com/flang-compiler/flang/pull/1044 was accidentally merged and reverted, but I thought it is best to just be honest and straightforward - this is a revert of a revert.
If reviewers think this should be set up in a different way, let me know.
This reverts commit 0a35fd61899924f6dd397a5a86e14915fe3fecf4, which was reverting an accidentally merged commit.


pawosm-arm (approved):
This time I'm going to click a button in the right area.

bryanpkc:
Question: If the original patch was **accidentally** merged, and then reverted **for a good reason**, shouldn't we now commit an amended version of the original patch, instead of just reverting the revert?

michalpasztamobica:
@bryanpkc , the reason it was reverted was that it should have waited for related PRs in `classic-flang-llvm-project` - there was nothing else wrong with the patch at that time.

However now that commit was actually used to run the CI I did find a bug in the way commit was being checked out on PRs (my fault - I should have tested it on my branch, while I only focused on commits landing on master branch). Now that I think of it - I have amended the reverted commit, but maybe I should keep these extra changes in a separate commit, so we have a "revert of the revert" commit + "Additional fixes" commit...

.github/workflows/build_flang_arm64.yml
```diff
@@ -36,18 +35,20 @@ jobs:
           cmake --version
           make --version
 
-      - name: Extract SHA
-        run: echo "::set-output name=sha::${GITHUB_SHA}"
-        id: extract
+      - name: Manual checkout to build in user's home dir (push)
+        if: github.ref == 'refs/heads/main'
```
bryanpkc, line 39: There is no `main` branch in https://github.com/flang-compiler/flang/. Is this intentional or a bug?

bryanpkc:
@michalpasztamobica I think it's better to just push one fixed commit, rather than to revert a revert of a known bad commit. The commit message can just use the same title as before and explain that this is a fixed version of the original commit.

michalpasztamobica:
@bryanpkc , your proposition to squash the commits sounds fine. Please, let me know if the current set up is what you expected.

I also fixed the bug for detecting the PR versus push to master, thanks for catching this. 
Unfortunately both ARM64 servers are connected to `flang-compiler` now, but [I now tried this method of telling PR from a push](https://github.com/michalpasztamobica/flang/runs/2713169168?check_suite_focus=true) to `master` on my own fork, so I am pretty sure that it works.

.github/workflows/build_flang_arm64.yml
```diff
@@ -0,0 +1,96 @@
...
+      image: ghcr.io/${{ github.repository_owner}}/ubuntu20-flang-${{ matrix.llvm_branch }}:latest
+      credentials:
+        username: github
+    strategy:
+      matrix:
+        target: [AArch64]
+        cc: [gcc]
+        cpp: [g++]
+        version: [10]
+        llvm_branch: [release_11x, release_12x]
```
bryanpkc, line 29: Perhaps this was discussed already in #1044 or even earlier but are we leaving out `release_100` intentionally?

michalpasztamobica: Yes, I only mentioned this in the technical call on the 21st of April ([see meeting minutes](https://docs.google.com/document/d/1-OuiKx4d7O6eLEJDBDKSRnSiUO2rgRR-c2Ga4AkrzOI/edit#)):

> Due to limited resources the builds would primarily run for release_11x and release_12x with gcc-10 with the possibility of further expansion if more runners are added or if the load proves to be low enough

We have very fast 32-core servers, but just two of them at the moment. Before we add more than two jobs we wanted to make sure that ARM64 builds do not become a bottle neck for PRs. I remember Huawei is most interested in even-number llvm releases, so please align with @kiranchandramohan  and @RichBarton-Arm if adding release_100 is crucial for you. It's just a matter of risking that sometimes we'll have to queue jobs before the runners are available.

bryanpkc: Oh right, I remember that discussion. Thanks.

bryanpkc (approved):

michalpasztamobica:
@kiranchandramohan , @SouraVX if you have any remarks to this PR I will be very happy to hear your feedback. If not, I would be grateful for merging this in, so I can move forward with the build scripts we discussed two weeks ago in the Wednesday call. Thank you in advance!

SouraVX:
[copied] from an resolved conversation.
We're still building for release_100 for x86 target here https://github.com/flang-compiler/flang/blob/master/.github/workflows/build_flang.yml#L20
Just wanted to understand your thoughts, are we planning to remove release_100 for x86 also ?

SouraVX:
> Hi @SouraVX , I don't think we will be removing `release_100` from X86 builds any time soon. I would rather say - if we make sure that the current set up is not slowing down PR CI checks we may want to add `release_100` to ARM64 builds :)

Thank you @michalpasztamobica for clarification!

PR #1055, opened 2021-05-31
Fix 2 bugs of PACK when the mask arg is a scalar of false

qiaozhang-hnc:
1. When the mask is a scalar false and the vector arg is absent, `PACK` wrongly returns an array of the same size as the array arg with all elements being the zero value, while it should return an empty array. This happens because the size of the result is calculated wrongly in the particular situation and the patch fixes it by giving the correct size for the result.
2. When the mask is a scalar false but the vector arg is given, `PACK` returns an array of the same size as the vector but all elements being zero, while the elements should have the same value and in the same order as the vector. This bug is caused in the runtime lib code for the intrinsic function `PACK` where it doesn't collect vector elements and returns too early when the the mask is a false scalar value. This patch therefore fixes the bug by simply remove the check-then-return block of code.

### DIFF
--- a/runtime/flang/pack.c
+++ b/runtime/flang/pack.c
@@ -70,8 +70,6 @@ void ENTFTN(PACK, pack)(void *rb,         /* result base */
   if (ISSCALAR(mask)) {
     mlen = GET_DIST_SIZE_OF(TYPEKIND(mask));
     mval = I8(__fort_varying_log)(mb, &mlen);
-    if (!mval)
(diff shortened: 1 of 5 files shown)


pawosm-arm (approved):

bryanpkc (approved):
LGTM.

shivaramaarao (approved):

bryanpkc (approved):

Issue #1056, opened 2021-05-31
[CP2K] undefined reference observed after #695 on x86_64 target


SouraVX:
Observed after #695
cp2k-7.1: error "undefined symbol: __mth_i_kidnnt" while building CP2K PSMP

log:
```
/home/cp2k/LIBINT_INSTALL/lib/libint2.a -lflang -lpgmath -lstdc++
 ld.lld: error: undefined symbol: __mth_i_kidnnt
 >>> referenced by hfx_compression_methods.F:379
```

SouraVX:
@bryanpkc @kiranchandramohan , did you guys also faced similar/related issue ?

pawosm-arm:
This must be x86_64 specific. I'm building CP2K with removed runtime duplicated files on AArch64 and I'm not experiencing that. As there is no plan here to build CP2K on x86_64, I'd rather revert that commit than fix it.

pawosm-arm:
@SouraVX I presume you have the ability to build CP2K on some x86_64 system. We can take cooperative approach here, you may try following change to see if it helps with building it on the architecture I'm not building it for:
```
diff --git a/runtime/libpgmath/lib/common/CMakeLists.txt b/runtime/libpgmath/lib/common/CMakeLists.txt
index f0e1b341..c953eec0 100644
--- a/runtime/libpgmath/lib/common/CMakeLists.txt
+++ b/runtime/libpgmath/lib/common/CMakeLists.txt
@@ -72,7 +72,8 @@ if(${LIBPGMATH_SYSTEM_PROCESSOR} MATCHES "x86_64" AND NOT ${LIBPGMATH_WITH_GENER

     set(SRCS
       mth_yintrinsics.c
-      mth_zintrinsics.c)
+      mth_zintrinsics.c
+      kidnnt.c)
   elseif(${LIBPGMATH_SYSTEM_NAME} MATCHES "Darwin|Windows")
     add_subdirectory("sincosf")
     add_subdirectory("tanf")
@@ -112,7 +113,8 @@ if(${LIBPGMATH_SYSTEM_PROCESSOR} MATCHES "x86_64" AND NOT ${LIBPGMATH_WITH_GENER

     set(SRCS
       mth_yintrinsics.c
-      mth_zintrinsics.c)
+      mth_zintrinsics.c
+      kidnnt.c)
   endif()
 elseif(${LIBPGMATH_SYSTEM_PROCESSOR} MATCHES "ppc64le")
[...]

PR #695, opened 2019-03-26
runtime: remove duplicate files

pawosm-arm:
Having those duplicates results in unwanted and sometimes
surprising consequences.

PR #1061, opened 2021-06-04
Fix x86_64 linker issues caused by PR #695

pawosm-arm:
This patch fixes link-time issues described in bug report #1056 on
x86_64 machines. The commit in PR #695 has removed duplicated files,
kidnnt.c among them. Unfortunately, due to a sudden omission, the
original kidnnt.c file was not listed for being built for x86_64
targets.


SouraVX (approved):
Thanks for the patch :)

PR #1066, opened 2021-06-06
Fix `switch`- and `case`-related warnings; NFCI

bryanpkc:
Add `FLANG_FALLTHROUGH` in affected case statements to silence
`-Wimplicit-fallthrough` warnings. The definition of the macro is lifted
from LLVM's `include/llvm/Support/Compiler.h`. Also add or remove default
`case` statements to silence `-Wswitch` or `-Wcovered-switch-default` warnings. NFCI.

PR #1067, opened 2021-06-06
[flang2] Fix fallthrough in `mk_cmp` (dinit.cpp)

bryanpkc:
The original fallthrough behaviour was probably not intentional.

This patch was part of #1066 but since it is no longer obviously NFCI, we have put the change in a separate patch.

### DIFF
--- a/tools/flang2/flang2exe/dinit.cpp
+++ b/tools/flang2/flang2exe/dinit.cpp
@@ -4618,9 +4618,9 @@ mk_cmp(CONST *c, int op, INT l_conval, INT r_conval, DTYPE rdtype, DTYPE dt)
     break;
   case OP_XTOI:
   case OP_XTOK:
-    c->u1.conval = init_fold_const(get_ast_op(op), l_conval, r_conval, rdtype);   
   default:
     c->u1.conval = init_fold_const(get_ast_op(op), l_conval, r_conval, dt);
+    break;
   }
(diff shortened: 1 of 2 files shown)


tools/flang2/flang2exe/dinit.cpp
```diff
@@ -4618,7 +4618,8 @@ mk_cmp(CONST *c, int op, INT l_conval, INT r_conval, DTYPE rdtype, DTYPE dt)
     break;
   case OP_XTOI:
   case OP_XTOK:
-    c->u1.conval = init_fold_const(get_ast_op(op), l_conval, r_conval, rdtype);
```
pawosm-arm, line 4621: how could this possibly work without `break`? I hope you've tested it thoroughly

pawosm-arm (approved):
some test case showing that nothing got broken would be helpful, but otherwise this change seems necessary enough to have it approved.

tools/flang2/flang2exe/dinit.cpp
bryanpkc: The original code actually behaved correctly. I have rewritten the patch and added a test case.

pawosm-arm (changes requested):

shivaramaarao (approved):

pawosm-arm (approved):

PR #1068, opened 2021-06-06
[flang1] Fix malformed `switch` statement in dinit.c

bryanpkc:
This patch fixes an obviously malformed `switch` statement, which evaluates
intrinsic calls with constant arguments in data initializers. After the
fix, the default case (which was never exercised) would cause spurious
warnings, so the warning is turned into an ICE in the applicable case.

### DIFF
--- a/tools/flang1/flang1exe/dinit.c
+++ b/tools/flang1/flang1exe/dinit.c
@@ -775,23 +775,25 @@ dinit_acl_val2(int sptr, int dtype, ACL *ict, int op)
           case AC_I_selected_real_kind:
             process_real_kind(sptr, subict, op);
             break;
+          case AC_I_selected_char_kind:
+            ast = subict->u1.ast;
+            if (A_TYPEG(ast) == A_CNST) {
+              int dty;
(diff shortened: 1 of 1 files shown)


pawosm-arm (approved):
Interesting. Did you encounter any observable effect of this change?

bryanpkc:
@pawosm-arm Merely fixing the malformed `switch` without removing the ICE in the `default` case would cause `test/f90_correct/src/ei28.f90` to fail. That test case contains this code:
```fortran
!Following will be done in back end dinit
logical :: int7 = (ichar("a") == 97)
logical :: int8 = (ichar("a") >= 97)
logical :: int9 = (ichar("a") > 97)
logical :: int10 = (ichar("a") <= 97)
logical :: int11 = (ichar("a") < 97)
logical :: int12 = (ichar("a") .ne. 97)
```
Each of these lines calls the `ichar` intrinsic in a data initializer, which is intentionally not handled in flang1's `dinit.c`.

shivaramaarao (approved):

PR #1070, opened 2021-06-06
[flang2] Remove unused functions from mwd.cpp; NFC

bryanpkc:
`tools/flang2/flang2exe/mwd.cpp` is causing Clang 12 to complain about unused functions. The file contains mostly undocumented `static` functions that are apparently used for debugging (a comment at the top of the file calls them "mw's dump routines").

This patch simply deletes the functions that were causing Clang 12 to complain. Even though most of the routines are guarded with `#if DEBUG`, their `static`-ness and lack of use will still cause problems in a debug build. In #1071, similar debug functions in other files are declared non-`static` to avoid the warnings; we could do the same for the ones in `mwd.cpp`, depending on how useful the community thinks these routines are. If they aren't being used by anyone, perhaps deleting them isn't a problem.

### DIFF
--- a/tools/flang2/flang2exe/mwd.cpp
+++ b/tools/flang2/flang2exe/mwd.cpp
@@ -192,20 +192,6 @@ putINT(const char *s, INT d)
 } /* putINT */
 
 static void
-putisz(const char *s, ISZ_T d)
-{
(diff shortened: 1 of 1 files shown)


pawosm-arm (approved):
Amazing how much of the code is not needed!

shivaramaarao (approved):
looks fine to me.

PR #1071, opened 2021-06-06
[flang1][flang2] Guard unused code with macros; NFCI

bryanpkc:
Work around `-Wunused-function` warnings by guarding unused code with `FLANG_*_UNUSED` macros. Functions that are unused because they are only called by conditionally compiled code are guarded with the same macros as their callers. Functions intended for debugging (e.g. calling within gdb) are guarded with `#if DEBUG` and declared non-static. Trivial unused functions are deleted completely.

I considered the alternative of simply deleting the unused functions, but I was wary of the work it would involve to rewrite or re-enable them in the future, should they be needed again. Adding the `FLANG_*_UNUSED` macros, as a deprecation measure, gives people a chance to think about how the functions could be re-included in the build, while keeping it easy to review future re-enablement patches (they would just be simple deletion of the macros).

### DIFF
--- a/runtime/flang/allo.c
+++ b/runtime/flang/allo.c
@@ -102,6 +102,7 @@ static long num_hdrs = NUM_HDRS;
 
 #define ALLHDR()
 
+#ifdef FLANG_ALLO_UNUSED
 /** \brief
  * Allocate ALLO_HDR list */
 static void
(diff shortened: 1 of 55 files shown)


pawosm-arm (approved):
What target architectures did you test it on?

bryanpkc:
> What target architectures did you test it on?

Linux aarch64 and Linux x86_64.

bryanpkc:
ping @shivaramaarao @SouraVX

shivaramaarao (approved):
LGTM

PR #1074, opened 2021-06-06
[accpp] Prevent `delete()` from writing to its argument

bryanpkc:
In `tools/flang1/flang1exe/accpp.c`, the function `delete` writes a terminating `NULL` to its `char *` argument when the argument is longer than `MAXIDLEN`, only for passing it to `strcmp`. Using `strncmp` instead allows the parameter to be declared `const char *`, which silences compiler warnings and also eliminates an unnecessary surprising behaviour.


pawosm-arm (approved):
I'm getting blind not being able to differ strncpy from strncmp. Turned out this commit is OK.

shivaramaarao (approved):

PR #1076, opened 2021-06-06
Fix `-Wsign-compare` warnings

bryanpkc:
Many `-Wsign-compare` warnings are caused by the fact that the `STG_MEMBERS` macro defines a number of `unsigned int` fields, which are often used in comparisons with signed variables. The problem is pervasive and requires further investigation in the future. For now, we reduce the errors to warnings by passing `-Wno-error=sign-compare` to Clang. Other `-Wsign-compare` warnings that are not related to `STG_MEMBERS` are fixed in separate commits.

Also, make `-DWITH_WERROR=on` enable `-Werror` not only for C compilations but also for C++ compilations.


runtime/flang/bcopys.c
```diff
@@ -26,7 +26,7 @@ __fort_bcopysl(char *to, char *fr, size_t cnt, size_t tostr, size_t frstr,
   size_t i, j;
   unsigned long n;
 #if !defined(WIN64)
-  long k;
+  size_t k;
 #else
   long long k;
```
pawosm-arm, line 31: I don't like leaving it like that. Later on, `k` is used for comparison with `len` which is of unsigned type `size_t`. Imo, `k` should be always of type `size_t` regardless the OS targeted.

tools/flang1/flang1exe/dump.c
```diff
@@ -693,7 +693,7 @@ putoptype(char *s, int optype)
 void
 dastli(int astli)
 {
-  int a;
+  unsigned int a;
```
pawosm-arm, line 696: would it be easier to do `static_cast` to silence a warning in the `for` loop condition below? Once you change signedness of this variable, this changes should be reflected further, namely, `%d` should be replaced with `%u` in `sprintf` (not sure if it was done wrong for the results of those ASTLI* macors already).

tools/flang1/flang1exe/exterf.c
```diff
@@ -1306,7 +1306,8 @@ export_ivl_ict(int lineno, VAR *ivl, ACL *ict, int dostructures)
 static void
 export_component_init(int cleanup)
 {
-  int dtype, flag;
+  unsigned int dtype;
+  int flag;
```
pawosm-arm, line 1310: it's not often to see signed flags... is it really ok?

pawosm-arm (changes requested):
The sheer extent of this patch is terrifying :) but it's much needed nevertheless.

Some small comments I have left.

tools/flang1/flang1exe/dump.c
bryanpkc: Actually this change was wrong. The type of `ASTLI_NEXT` is `int` and isn't the cause of the warning. The warning is caused by the comparison with `astb.astli.stg_avail` which is declared unsigned by the `STG_MEMBERS` macro. As I mentioned previously, solving the `STG_MEMBERS` problem will have to wait for another time. I will undo this hunk.

runtime/flang/bcopys.c
bryanpkc: You're right. `k` should always be of type `size_t`.

tools/flang1/flang1exe/exterf.c
bryanpkc: This code is a bit weird. `flag` is bitwise AND-ed and OR-ed with `ict->ci_exprt` in the function, but the `ci_exprt` field is declared as a single bit in a bit field. I am not sure whether initializing `flag` to 2 in this function is intentional or not. 

I will undo this hunk, since `dtype` should not be unsigned. The warning is coming from a comparison with `stb.dt.stg_avail` which is declared unsigned by the `STG_MEMBERS` macro.

bryanpkc:
@pawosm-arm Could you take a look again?

pawosm-arm (approved):

shivaramaarao (approved):
LGTM

bryanpkc:
Thanks for the review guys!

PR #1082, opened 2021-06-07
Add build scripts and switch CI to use them

michalpasztamobica:
As discussed in the Wednesday call - build scripts will make it easier to reproduce CI runs for users.
They were tailored with current CI setup in mind and the CI scripts were adjusted to use them.
I was partially inspired by @bryanpkc 's scripts available [here](https://github.com/Huawei-PTLab/classic-flang-llvm-project/wiki). Thanks, Bryan!
A PR to `classic-flang-llvm-project` will follow with similar build scripts introduced there.

### DIFF
--- a/.github/workflows/build_flang.yml
+++ b/.github/workflows/build_flang.yml
@@ -10,7 +10,7 @@ on:
     paths-ignore:
       - '**/.github/workflows/build_flang_arm64.yml'
 
-jobs:       
+jobs:
   build_flang:
     runs-on: ubuntu-20.04
     env:
@@ -62,7 +62,7 @@ jobs:
           cd ../..
           curl -sL https://api.github.com/repos/flang-compiler/classic-flang-llvm-project/actions/workflows/pre-compile_llvm.yml/runs --output runs_llvm.json
           wget --output-document artifacts_llvm `jq -r '.workflow_runs[0].artifacts_url?' runs_llvm.json`
-          
+
           i="0"
           # Keep checking older builds to find the right branch and correct number of artifacts
           while { [ `jq -r '.total_count?' artifacts_llvm` != "3" ] || \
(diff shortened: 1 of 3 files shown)


SouraVX (approved):
Looks great thank @michalpasztamobica ! , lots code refactored to common place/script + clear delineation between administration and execution :)

pawosm-arm (changes requested):
@michalpasztamobica we're using scripts like that internally, however all of them are armed with error trapping as such:

```
abort()
{
    echo >&2 '
    ***************
    *** ABORTED ***
    ***************
    '
    echo "An error occurred. Exiting..." >&2
    exit 1
}

trap 'abort' 0

set -e
set -o pipefail

...your code here...

trap : 0
```

michalpasztamobica:
Thanks for the hints @pawosm-arm . Can you help me understand how exactly using `trap` would improve the flow? I have never used `trap` before, but from what I understand and quickly tested locally this would basically trigger on return from shell - is this just to improve the visibility of an error in the logs? If so - I would personally leave the script as is - CMake and Make are pretty explicit on error printing and they will land at the end of the step anyway - I don't see the need to introduce such error message printing. Is it OK?

Regarding `set -e` I considered using it instead of suffixing every command with ` || exit 1`. It does have a few exceptions and I have heard voices saying it is considered dangerous. Personally I am in favor of using it instead of ` || exit 1`, I just didn't want to be controversial :D . I will change the script to use `set -e` and let's see if anyone sees the danger of using it in our script.

Regarding `set -o pipeline` - we don't have any pipelines in the script, I see no need of introducing this setting.

pawosm-arm:
@michalpasztamobica the whole point of using trap and `set -e` was for freeing you from remembering to add `|| exit 1` in dozens of places. It's a matter of taste, for me, the script full of `|| exit 1` seems polluted. Also, in our scripts we use pipelines a lot, in my snippet I've tried to present one-fit-for-all template. It's up to you to decide which is the best.

pawosm-arm (approved):

michalpasztamobica:
I just force-pushed an update that uses `set -e` instead of `|| exit 1`s. Thanks again for the hints, @pawosm-arm !

The `classic-flang-llvm-project` part PR is already there: https://github.com/flang-compiler/classic-flang-llvm-project/pull/51. I am just having some issues with CI (missing `\`s, incorrect paths, etc.). That PR is temporarily using my fork, so the flang build scripts are available. But I will switch it to `flang-compiler` as soon as I can see the CI working.

bryanpkc:
LGTM. FWIW in our internal build script, we use `|| exit 1`, but `set -e` is nice as well. We also `set -x` just before every command that is important, e.g.
```sh
if [ $clean -eq 1 ]; then
  (set -x; rm -rf "$build_prefix") || exit 1
fi
if [ $configure -eq 1 -o ! -f "$build_prefix/CMakeCache.txt" ]; then
  mkdir -p "$build_prefix"
  pushd "$build_prefix"
  (set -x; \
    $cmake -Wno-dev -G "$generator" \
           -DCMAKE_BUILD_TYPE=$buildtype \
           -DCMAKE_INSTALL_PREFIX=$install_prefix \
           .
           .
           ../llvm) \
    || exit 1
else
  pushd "$build_prefix"
fi
(set -x; $ninja $install) || exit 1
```
As Paul says the code is quite polluted but the script now prints the important commands it is executing, which helps us debug our CI jobs.

We also use `trap`, not as replacement for `set -e`, but as a sort of `finally` statement to kill off subprocesses before terminating a CI job.

bryanpkc:
Does the script have to reside in `.github/workflows/`? It could potentially be in a more user-visible location, and used by regular developers as well?

michalpasztamobica:
@bryanpkc - regarding the location - I also though about putting the script in `/scripts` but I did not want to introduce a new directory in the main source tree. If that is OK with everyone, I can move the scripts.

bryanpkc:
We could put it in the project root, unless you expect more scripts in the future.

michalpasztamobica:
Not at the moment. The location in the root directory sounds good - it will let us call the script without specifying complicated paths.

michalpasztamobica:
@kiranchandramohan , I think this PR got all required approvals. If you have a moment, would you please give it a look and merge if everything looks OK? Thank you!

PR #1083, opened 2021-06-09
[DebugInfo] Flang should include compiler options in DW_AT_producer

alokkrsharma:
Currently Flang contains only compiler information for DWARF attribute DW_AT_producer.
``````````````
 _llvm-dwarfdump test.o | grep DW_AT_producer_
              DW_AT_producer    (**" F90 Flang - 1.5 2017-05-01"**)
Other compilers include compiler options as well.
_llvm-dwarfdump test.gf.o | grep DW_AT_producer_
DW_AT_producer    (**"GNU Fortran2008 7.5.0 -mtune=generic -march=x86-64 -gdwarf-5 -fintrinsic-modules-path /usr/lib/gcc/x86_64-linux-gnu/7/finclude"**)
``````````````
Current Patch includes compiler options also for Flang
This is done by adding a field "flags", rest is taken care by LLVM back-end.
!2 = distinct !DICompileUnit(language: DW_LANG_Fortran90, file: !3, producer: " F90 Flang - 1.5 2017-05-01", isOptimized: false, **flags: "'+flang -gdwarf-5 producer.f90 -S -emit-llvm'"**, runtimeVersion: 0, emissionKind: FullDebug, enums: !4, retainedTypes: !4, globals: !4, imports: !4)

### DIFF
--- a/test/debug_info/producer.f90
+++ b/test/debug_info/producer.f90
@@ -1,0 +1,7 @@
+!RUN: %flang -gdwarf-4 -S -emit-llvm %s -o - | FileCheck %s
+
+!CHECK: !DICompileUnit(language: DW_LANG_Fortran90
+!CHECK-SAME: flags: "'+flang -gdwarf-4 -S -emit-llvm 
+
+program main
+end program main
--- a/tools/flang2/flang2exe/lldebug.cpp
+++ b/tools/flang2/flang2exe/lldebug.cpp
@@ -1798,7 +1798,7 @@ lldbg_emit_compile_unit(LL_DebugInfo *db)
     lang_tag = DW_LANG_Fortran90;
     db->comp_unit_mdnode = lldbg_create_compile_unit_mdnode(
         db, lang_tag, get_filename(1), get_currentdir(), db->producer, 1,
-        flg.opt >= 1/*isOptimized Flag*/, "",
+        flg.opt >= 1/*isOptimized Flag*/, flg.cmdline,
         0, &db->llvm_dbg_enum, &db->llvm_dbg_retained, &db->llvm_dbg_sp,
         &db->llvm_dbg_gv, &db->llvm_dbg_imported);
   }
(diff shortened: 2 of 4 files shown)


SouraVX (approved):

alokkrsharma:
Hi @kiranchandramohan and @bryanpkc , can you please review this ?

bryanpkc (approved):
LGTM.

kiranchandramohan:
The dwarf standard only requires that this attribute should identify the compiler. It does not demand that the compiler options be included. I see that it says that the actual contents will be specific to each compiler. I checked gfortran and it includes the compiler options but clang does not. I have a few questions,
1) Shouldn't we be consistent with Clang?
2) Why does gfortran generate the compiler options also for this attribute? Is it used by the debugger or other tools? 
 
"A DW_AT_producer attribute whose value is a null-terminated string
containing information about the compiler that produced the compilation
unit.
The actual contents of the string will be specific to each producer, but should begin
with the name of the compiler vendor or some other identifying character sequence
that will avoid confusion with other producer values."

alokkrsharma:
> 
> 
> The dwarf standard only requires that this attribute should identify the compiler. It does not demand that the compiler options be included. I see that it says that the actual contents will be specific to each compiler. I checked gfortran and it includes the compiler options but clang does not. I have a few questions,
> 
>     1. Shouldn't we be consistent with Clang?
> 
>     2. Why does gfortran generate the compiler options also for this attribute? Is it used by the debugger or other tools?
> 
> 
> "A DW_AT_producer attribute whose value is a null-terminated string
> containing information about the compiler that produced the compilation
> unit.
> The actual contents of the string will be specific to each producer, but should begin
> with the name of the compiler vendor or some other identifying character sequence
> that will avoid confusion with other producer values."

There can be cases when it is important to identify which optimization level the binary is generated. There can be multiple binaries generated with different options and different versions. Sometime this information can be even more crucial in cases of crashes which can be limited to compiler options.

In our case (flang/clang), compiler back-end LLVM already has the support to include compiler options. That suggests the front-ends are going to use that sooner or later. The Flang needs lesser changes and hence the Flang is leading. The Clang is next in plan.

alokkrsharma:
Ping @kiranchandramohan !

kiranchandramohan:
The only concern I have is if this feature is not supported in Clang it might be difficult to support in llvm Flang. This might be a problem in future for you.

Since there is not issue in Classic Flang, I am approving it.

kiranchandramohan (approved):
LGTM.

PR #1089, opened 2021-06-15
Prevent illegal argument accessing during procedure exporting.

qiaozhang-hnc:
When exporting a procedure, if `PARAMCT` is not positive (e.g., when a name of a function is used as the procedure interface in a procedure declaration statement before the function is defined and later the function declartion statement has a `RESULT` clause
and zero dummy arguments, `PARAMCT` will be 0), then it should be illegal to access the beginning of the argument list of the function. zero or negative  Failing to prevent this kind of illegal access will cause a segmentation fault.

This patch fixes the bug by checking whether `PARAMCT` is greater than 0 before accessing the argument list.

### DIFF
--- a/test/f90_correct/src/proc_ptr.f90
+++ b/test/f90_correct/src/proc_ptr.f90
@@ -1,0 +1,30 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Test the fix for segment fault about procedure pointer when all the following
+! three conditions are satisfied:
+! 1) module procedue used as proc-interface
+! 2) keyword RESULT appears
+! 3) none dummy argument
+
+module m
+  procedure(g), pointer :: f
+contains
+  function g() result(z)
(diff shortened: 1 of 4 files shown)


test/f90_correct/src/proc_ptr.f90
```diff
@@ -0,0 +1,30 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Test the fix for segment fault about procedure pointer under the following
+! conditions:
+! 1) module procedue used as proc-interface
+! 2) keyword RESULT appears
+! 3) none dummy argument
```
PeixinQiao, line 10: To my understanding, your code fix is for the condition when all the three conditions are satisfied, not one of them. If my understanding is correct, the comments may need to be changed by replacing conditions with one condition.

qiaozhang-hnc: You are right. I rephrased the description as suggested. Thanks!

pawosm-arm (approved):

shivaramaarao (approved):

bryanpkc (approved):

PR #1091, opened 2021-06-16
Add more checkings for derived types in COMMON statements

qiaozhang-hnc:
This patch fixes two problems:
1. When two or more `COMMON` blocks with the same name have objects with default initialization, flang will execute each 
    initialization more than once. The first pass will carry out normally, whereas the others will be rejected in the `write_comm` 
    function, leaving the `DSRT` pointer un-nullified, ultimately causing a segmentation fault.
2. If a common-block-object is of a derived type, the type shall have the BIND attribute or the SEQUENCE attribute and it shall 
    have no default initialization. A common-block-object shall not be a derived-type object with an ultimate component that is 
    allocatable. flang fails to enforce the above restrictions on derived types.

This patch fixes the 2 problems above by setting `DSRT` to `NULL` at the right time and capturing/reporting the aforementioned
syntactical error.

### DIFF
--- a/test/f90_correct/inc/common_02.mk
+++ b/test/f90_correct/inc/common_02.mk
@@ -1,0 +1,15 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+build: $(SRC)/$(TEST).f90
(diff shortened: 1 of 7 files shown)


bryanpkc:
@qiaozhang-hnc Could you rebase this branch please?

qiaozhang-hnc:
> @qiaozhang-hnc Could you rebase this branch please?

Rebased.

pawosm-arm (approved):

PR #1092, opened 2021-06-16
[scutil] Fix a couple of minor compilation warnings; NFCI

bryanpkc:
- Do not use `STDC FENV_ACCESS` pragma on AArch64 which is not supported by GCC and Clang.
- Use unsigned wrap function for `DBLUINT64`.

### DIFF
--- a/lib/scutil/host-fp-folding.c
+++ b/lib/scutil/host-fp-folding.c
@@ -59,7 +59,9 @@ fold_sanity_check(void)
  *  C99 feature: alert the compiler that the following code cares about
  *  the dynamic floating-point environment.
  */
+#ifndef __aarch64__
 #pragma STDC FENV_ACCESS ON
+#endif
 
 /*
  *  Configure "denormal inputs are zero" and "flush denormal results to zero"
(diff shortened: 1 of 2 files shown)


lib/scutil/host-fp-folding.c
```diff
@@ -55,12 +55,6 @@ fold_sanity_check(void)
   assert(sizeof(float128_t) == 16);
 }
 
-/*
- *  C99 feature: alert the compiler that the following code cares about
- *  the dynamic floating-point environment.
- */
-#pragma STDC FENV_ACCESS ON
```
shivaramaarao, line 62: I understand that clang supports this pragma.  Is the pragma not important for the functionality of the code below?

bryanpkc: LLVM 12 ignores this pragma when running without strict FP support. See https://reviews.llvm.org/D90316 for information.

bryanpkc: It seems that `HasStrictFP` is `true` for X86 and PPC, but not for AArch64.

bryanpkc: Fixed.

pawosm-arm (approved):

lib/scutil/host-fp-folding.c
bryanpkc: @shivaramaarao Please take a look at this again?

bryanpkc:
ping @shivaramaarao @SouraVX

bryanpkc:
ping @shivaramaarao

bryanpkc:
ping @shivaramaarao

shivaramaarao (approved):

PR #1093, opened 2021-06-17
Fix an ICE issue caused by const propagation

zhaochuanfeng:
When an ast rewrite is performed during constant propagation optimization, we make the dtype of new ast be the same as the original.
Consider following ast tree:
```
 +-- ast:31  atype:7:convert  shape:1  dtype:58:array deferred
     nobounds (z_b_0:z_b_1) of double precision
   +-- ast:29  atype:1:ident shape:0  dtype:10:double precision
        sptr:680=weigh_cyl
```
If ast 29 is replaced by ast y, ast 31 will be rewrited to new ast x, and the dtype of new ast x should be as some as ast 31.

### DIFF
--- a/test/f90_correct/src/const_prop.f90
+++ b/test/f90_correct/src/const_prop.f90
@@ -1,0 +1,20 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+
+! test constant propagation optimization
+
+program test
+real(8) :: weight
+real(8), allocatable, dimension(:) :: a
+
+allocate(a(8))
+
+weight = 1._8
+a = weight + a
+
+write(*, *) 'PASS'
+
+end program test
--- a/tools/flang1/flang1exe/ast.c
+++ b/tools/flang1/flang1exe/ast.c
@@ -4333,7 +4333,8 @@ ast_rewrite(int ast)
       if (rank != rank1) {
         if (rank == 0)
           rank = rank1;
-        dtype = get_array_dtype(rank, DDTG(A_DTYPEG(ast)));
+        /* Make the dtype of new ast the same as the old one. */
+        dtype = dtype_with_shape(A_DTYPEG(ast), A_SHAPEG(ast));
       }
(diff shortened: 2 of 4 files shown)


test/f90_correct/src/const_prop.f90
```diff
@@ -0,0 +1,20 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+
+! test constant propagation optimization
+
+program test
+real(8) :: weigh
```
bryanpkc, line 10: `weight`?

tools/flang1/flang1exe/ast.c
```diff
@@ -4333,7 +4333,8 @@ ast_rewrite(int ast)
       if (rank != rank1) {
         if (rank == 0)
           rank = rank1;
-        dtype = get_array_dtype(rank, DDTG(A_DTYPEG(ast)));
+        /* make the dtype of new ast the some as the old one. */
```
bryanpkc, line 4336:
```suggestion
        /* Make the dtype of new ast the same as the old one. */
```

test/f90_correct/src/const_prop.f90
zhaochuanfeng: Fixed, thanks~

tools/flang1/flang1exe/ast.c
zhaochuanfeng: Fixed, thanks~

bryanpkc:
Was this test case part of an important Fortran application?

bryanpkc:
I compiled the test case with `-O2` (without your patch), but it doesn't produce an ICE, and the generated program runs correctly as well. Could you elaborate in the PR description what the AST in this example currently gets rewritten into, and how the problem can be triggered?

zhaochuanfeng:
> I compiled the test case with `-O2` (without your patch, on X86), but it doesn't produce an ICE, and the generated program runs correctly as well. Could you elaborate in the PR description what the AST in this example currently gets rewritten into, and how the problem can be triggered?

zhaochuanfeng:
The problem cannot be reproduced in the community version. Therefore, I closed this PR.

PR #976, opened 2021-01-25
Generate loop start, end locations in llvm.loop metadata

kiranchandramohan:
This patch from @peterwaller-arm adds the loop start, end locations in llvm.loop metadata. 
The patch also reorganizes the loop metadata generation to not generate duplicates, fixes some bugs.
A few existing testcases are modified since the llvm.loop metadata is always present.

PR #1094, opened 2021-06-21
Generate loop start, end locations in llvm.loop metadata, using new metadata

michalpasztamobica:
This PR is a continuation of https://github.com/flang-compiler/flang/pull/976
I rebased that PR on top of recent metadata changes from @Chirag-Khandelwal and adjusted it to use new metadata.
I humbly ask for a detailed review as this is the first time I touched the llvm IR generation.
In particular, please pay attention to the adjustments in `ivdep.f90` and `omp_simd.f90` - These tests were failing already in #976 . To my understanding the changes in the code generation were intentional and the tests needed to be adjusted, but I am not sure if I am right.


pawosm-arm (approved):
It looks fine, although "Replace llvm's mem.parallel_loop_access with group.access" commit message does not say why it is needed to do so.

pawosm-arm (changes requested):
it fails to build. Incomplete commit?

michalpasztamobica:
I somehow lost one line during rebase... But now it's back in - everything builds and passes tests :-) 
@pawosm-arm I also amended the commit message giving a link to the changes in llvm which triggered this.

pawosm-arm (approved):

michalpasztamobica:
@Chirag-Khandelwal , I cannot add you to the reviewers list, but your confirmation that this makes sense will be very valuable :)

Chirag-Khandelwal:
It looks great to me! :) @michalpasztamobica

michalpasztamobica:
@bryanpkc , @SouraVX , @peterwaller-arm , @kiranchandramohan , your feedback on this will be much appreciated. Thank you in advance.

liuxiao0909c:
Hi, @michalpasztamobica , After the commit https://github.com/flang-compiler/flang/pull/1094/commits/efa0d02e4a44b3896260b0287bcc212a55bdce69 ,  both `write_no_depcheck_metadata` and `write_llaccgroup_metadata`(added in commit https://github.com/flang-compiler/flang/commit/f434474f1f071be10e82689c5249408f234517b5 ) will generate `!llvm.access.group` metadata,  is there any possibility of a merger?

michalpasztamobica:
Hi @liuxiao0909c . Thanks for the review. 
Indeed, the functions were similar - I removed the older `write_no_depcheck_metadata` and parametrized `write_llaccgroup_metadata` to be more flexible. I hope this is OK now?

michalpasztamobica:
In yesterday's call @bryanpkc rightly pointed out that the new metadata might be breaking compatibility with older llvm versions.
However - I noticed that `llvm.access.group` is already [available in llvm-10](https://releases.llvm.org/10.0.0/docs/LangRef.html#llvm-access-group-metadata), so perhaps the situation is not that bad? @bryanpkc , would you be able to fully assess the risk for llvm-10 and let us know if this path is OK for you? I believe only Huawei is currently interested in llvm-10 support?

I don't think it makes sense to introduce release branches like we do in llvm-project-classic-flang, but if necessary we could add a CMake flag, to address code differences for different llvm versions?

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -3350,8 +3337,10 @@ write_instructions(LL_Module *module)
 
         assert(p->next == NULL, "write_instructions(), bad next ptr", 0,
                ERR_Fatal);
-        write_no_depcheck_metadata(module, instrs);
-        write_llaccgroup_metadata(module, instrs);
+        write_llaccgroup_metadata(module, instrs,
+			LDST_HAS_METADATA, instrs->misc_metadata);
```
bryanpkc, line 3341: Please use spaces instead of tabs for indentation, and left-align multi-line parameter lists.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -2920,28 +2920,15 @@ write_memory_order_and_alignment(INSTR_LIST *instrs)
 }
 
 INLINE static void
-write_no_depcheck_metadata(LL_Module *module, INSTR_LIST *insn)
+write_llaccgroup_metadata(LL_Module *module, INSTR_LIST *insn,
+		LL_InstrListFlags flag_check, unsigned int val)
```
bryanpkc, line 2924: Please use spaces instead of tabs for indentation, and left-align multi-line parameter lists.

michalpasztamobica: Thank you for catching this! Fixed now.

test/directives/unroll.f90
```diff
@@ -74,20 +72,19 @@ subroutine func3(m, a, b)
   ! CHECK:              br i1 {{.*}}, label %[[BB3]]
   ! CHECK-O0-SAME:      !llvm.loop [[MD_LOOP3:![0-9]+]]
   ! CHECK-O1-SAME:      !llvm.loop [[MD_LOOP3:![0-9]+]]
-  ! CHECK-DISABLED-NOT: !llvm.loop !{{[0-9]+}}
 end subroutine func3
 
-! CHECK-O0: [[MD_LOOP1]] = distinct !{[[MD_LOOP1]], [[MD_ENABLE:![0-9]+]]}
-! CHECK-O0: [[MD_ENABLE]] = !{!"llvm.loop.unroll.enable"}
+! CHECK-O0: [[MD_ENABLE:![0-9]+]] = !{!"llvm.loop.unroll.enable"}
+! CHECK-O0: [[MD_LOOP1]] = distinct !{[[MD_LOOP1]], {{.*}}, {{.*}}, [[MD_ENABLE]]}
```
bryanpkc, line 78: Can these lines use the FileCheck directive `CHECK-O0-DAG`? In my downstream build, the test outputs these lines in a different order:
```
!15 = !{!15, !12, !14, !16}
!16 = !{!"llvm.loop.unroll.enable"}
```

test/f90_correct/bih-head.f90
```diff
@@ -0,0 +1,94 @@
+!
```
kiranchandramohan, line 1: Can we move this test to a new directory (test/ili_correct) since this tests the ILI output? I am OK if it is done in a subsequent patch.

kiranchandramohan (commented):
@michalpasztamobica did you mention that there is a loop that does not vectorize after these changes? Could you point me to that test?

test/f90_correct/bih-head.f90
bryanpkc: +1. I was going to write the exact same comment.

bryanpkc:
> However - I noticed that llvm.access.group is already available in llvm-10, so perhaps the situation is not that bad? @bryanpkc , would you be able to fully assess the risk for llvm-10 and let us know if this path is OK for you?

@michalpasztamobica Thanks for linking the LangRef for LLVM 10. I have merged these patches into my downstream LLVM 10-based build, but the test cases `ivdep.f90`, `omp_simd.f90`, and `vector_directive3.f90` are broken:
```
.../llvm/lib/Analysis/LoopInfo.cpp:576: llvm::Loop::isAnnotatedParallel() const::<lambda(llvm::MDNode*)>: Assertion `isValidAsAccessGroup(AccGroup) && "List item must be an access group"' failed.
```
I will re-try a vanilla LLVM 10 build and see if it behaves the same way.

test/directives/unroll.f90
michalpasztamobica: Done, I hope I got it right? :)

michalpasztamobica:
@kiranchandramohan , I moved the test and squashed that change into Peter's original commit.
Regarding the vectorization, in `ivdep.f90` and `omp_simd.f90` I had to remove the `VF` regex to get the tests passing, like so:
```
! CHECK:  {{.*}} add nsw <[[VF:[0-9]+]] x i32>{{.*}}
```
to
```
! CHECK:  {{.*}} add nsw i32{{.*}}
```

test/directives/unroll.f90
bryanpkc: Yes, thanks.

kiranchandramohan:
Do you get a warning like the one below?

```
DILocation not allowed within this metadata node
!30 = distinct !{!30, !22, !23, !16, !31}
!16 = !DILocation(line: 10, column: 1, scope: !5)
warning: ignoring invalid debug info in /tmp/ivdep-383a70.ll
```

bryanpkc:
Nit: Would you mind rewording the commit currently titled "Adjust Chirag's changes to Peter's"? Please summarize what is changed in the commit title, and credit original authors in the message body.

kiranchandramohan:
For llvm.loop and llvm.access.group, the metadata generated seems to be the same (!33). That does not look OK to me. Is this expected?
```
%53 = load i32, i32* %.dY0001_313, align 4, !llvm.access.group !33, !tbaa !42, !dbg !34
store i32 %54, i32* %.dY0001_313, align 4, !llvm.access.group !33, !tbaa !42, !dbg !34
br i1 %55, label %L.LB1_311, label %L.LB1_312, !llvm.loop !33, !dbg !34
```

michalpasztamobica:
@bryanpkc  - sure, I will reword it, thanks for pointing this out.

@kiranchandramohan , the warnings do not show up when running `make check-all`, but I could now see them when manually compiling `ivdep.f90`. I tried to figure out which change could be an issue, but I couldn't make it out from the code. 
I then tried the original patches from https://github.com/kiranchandramohan/flang/tree/loop-line and got the same results: warning from the compiler and same value for `llvm.access.group` and `llvm.loop`. @peterwaller-arm , can you help us here?

kiranchandramohan:
The warnings are probably coming because the llvm.access.group contains location metadata which it is not permitted to contain. 

We have to first find out why the llvm.access.group metadata is pointing to the same metadata as the llvm.loop, that should not happen. Since llvm.access.group is being introduced in the latest patch for ivdep, we have to correct the latest patch. It is probably an incorrect interaction between the code that we added and the code that chirag added.

kiranchandramohan:
We cannot directly replace llvm.mem.parallel_loop_access with llvm.access.group as implemented in https://github.com/flang-compiler/flang/pull/1094/commits/26dc6c7da5bfb3ab2584504219f5fefa4ec6404e

The parameter of llvm.mem.parallel_loop_access was a loop id. But the parameter of llvm.access.group is a group id that is independent of the loop. The loop then should have the llvm.loop.parallel_accesses metadata which refers to the access groups. I see that the parallel_accesses metadata is generated but it is not referring to the access groups.

What we have now is something like the following for ivdep.f90,
```
        store i32 %76, i32* %.dY0001_313, align 4, !llvm.access.group !35, !dbg !34
        %77 = load i32, i32* %.dY0001_313, align 4, !llvm.access.group !35, !dbg !34
        ...
        br i1 %78, label %L.LB1_311, label %L.LB1_312, !llvm.loop !35, !dbg !34
        
!27 = distinct !{  }
!28 = !DILocation(line: 10, column: 1, scope: !25)
!30 = !{ !"llvm.loop.vectorize.enable", i1 1 }
!31 = !{ !"llvm.loop.parallel_accesses", !27 }
!34 = !DILocation(line: 12, column: 1, scope: !25)
!35 = !{ !35, !30, !31, !28, !34 }
``` 

What we should ideally have is the following,
```
        store i32 %76, i32* %.dY0001_313, align 4, !llvm.access.group !27, !dbg !34 <----- Note the change here
        %77 = load i32, i32* %.dY0001_313, align 4, !llvm.access.group !27, !dbg !34 <----- and here. Refers to a group id not loop id now.
        ...
        br i1 %78, label %L.LB1_311, label %L.LB1_312, !llvm.loop !35, !dbg !34

!27 = distinct !{  }        
!28 = !DILocation(line: 10, column: 1, scope: !25)
!30 = !{ !"llvm.loop.vectorize.enable", i1 1 }
!31 = !{ !"llvm.loop.parallel_accesses", !27 } 
!34 = !DILocation(line: 12, column: 1, scope: !25)
!35 = !{ !35, !28, !34, !30, !31 } <----- Note the re-ordering here. Location nodes should come before property nodes.
```

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1309,50 +1331,59 @@ finish_routine(void)
   }
 }
 
+/**
+   \brief Return current loop id and mark it as vectorizable.
+
+   No dependency checking currently means to mark a loop id metadata with
+   !"llvm.loop.vectorize.enable", and to mark memory instructions with
+   !llvm.mem.parallel_loop_access referring to the loop id. This function
```
kiranchandramohan, line 1339: Nit: This comment has to be updated. We do not generate llvm.mem.parallel_loop_access now.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -2868,28 +2920,15 @@ write_memory_order_and_alignment(INSTR_LIST *instrs)
...
-  }
-}
-
-INLINE static void
-write_llaccgroup_metadata(LL_Module *module, INSTR_LIST *insn)
+write_llaccgroup_metadata(LL_Module *module, INSTR_LIST *insn,
+                          LL_InstrListFlags flag_check, unsigned int val)
 {
-  if (insn->flags & LDST_HAS_ACCESSGRP_METADATA) {
+  if (insn->flags & flag_check) {
```
kiranchandramohan, line 2926: insn->flags seems to be not set when flag_check is LDST_HAS_ACCESSGRP_METADATA.

This ultimately happens because _mark_rw_access_grp_ is only called when 191, 0x4 is set (only if the directive is _vector always_). Could you call _mark_rw_access_grp_ for all of the cases where the access group metadata has to be generated??

```
    if (XBIT(191, 0x4)) {
      fix_nodepchk_flag(bih);
      mark_rw_access_grp(bih);
    }
```

kiranchandramohan (changes requested):

michalpasztamobica:
I have just pushed changes suggested off-line by @kiranchandramohan (I would have never found this fix. Thanks, Kiran!) I squashed the changes to the "Replace llvm's mem.parallel_loop_access with group.access" commit and mentioned the authorship in the commit's message. I hope this is OK?

I confirm that now the order of generated messages, is as desired and tests are still passing. The `VF` is back, so I also brought it back into the two tests.

I will look at the remaining issue and check whether `fix_nodepchk_flag` is needed. But that sounds optional, so I wanted to push what you have fixed, to unblock the PR.

@bryanpkc , the commit message has also been reworded.

kiranchandramohan:
Thanks @michalpasztamobica, we need to test this well. Will also benefit from other reviewers having a closer look.

michalpasztamobica:
@kiranchandramohan , I removed the body of the `fix_nodepchk_flag()` and that did not fail any test and did not influence the `.ll` output when building `ivdep.f90`. Any suggestions what else to check? Should I submit this as a separate patch for further verification with any available downstream tests?

kiranchandramohan:
If the nodepcheck flow is not used anymore I would recommend creating a fix_access_group_flag function.

michalpasztamobica:
Hi @kiranchandramohan ,
I am trying to figure out in detail how to approach the problem. Apologies, but I am still not very good with compiler-internals and the closer to LLVM, the bigger the gap in my knowledge.

Can you elaborate more on what the `fix_access_group_flag` would do? I assume the current `fix_nodepcheck_flang` function is allowing us to skip some particular dependency check between branches and that for some reason it seems not to make a difference any more now that access groups have been introduced. But we have switched from `llvm.mem.parallel_loop_accesses`, so I don't fully see the connection and why we even need to "fix" the `access_group` flag.

Sorry, if these are silly questions I am counting on your patience with me :)

tools/flang2/flang2exe/cgmain.cpp
michalpasztamobica: This has been addressed by adding the flag in the appropraite section.

kiranchandramohan:
I would like to understand what fix_nodepcheck_flag does before completely removing it from the flow. Alternatively, introduce a fix_access_group_flag function in place of fix_nodepcheck_flag.

Chirag-Khandelwal (commented):
This looks good to me (quite clean too!). I'll let others take a look.
Verified the tests as well, and they run just fine for me :)

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -982,51 +1004,51 @@ cons_loop_parallel_accesses_metadata(void)
...
-  loopVect = ll_get_md_node(cpu_llvm_module, LL_PlainMDNode, lvcomp, 2);
-  ll_extend_md_node(cpu_llvm_module, rv, rv);
-  ll_extend_md_node(cpu_llvm_module, rv, loopVect);
-  cpu_llvm_module->loop_md = rv;
-  return rv;
+  if (LL_MDREF_IS_NULL(cached_vectorize_disable_metadata)) {
+    lvcomp[0] = ll_get_md_string(cpu_llvm_module, "llvm.loop.vectorize.enable");
+    lvcomp[1] = ll_get_md_i1(0);
+    cached_vectorize_disable_metadata = ll_get_md_node(cpu_llvm_module,
+      LL_PlainMDNode, lvcomp, 2);
```
PeixinQiao, line 1018: Nit: Missing two spaces. Please use the following style.
```
    cached_vectorize_disable_metadata = ll_get_md_node(cpu_llvm_module,
        LL_PlainMDNode, lvcomp, 2);
```

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -982,51 +1004,51 @@ cons_loop_parallel_accesses_metadata(void)
...
-
   if (LL_MDREF_IS_NULL(cached_unroll_disable_metadata)) {
-   rv = ll_create_flexible_md_node(cpu_llvm_module);
    lvcomp[0] = ll_get_md_string(cpu_llvm_module, "llvm.loop.unroll.disable");
-   loopUnroll= ll_get_md_node(cpu_llvm_module, LL_PlainMDNode, lvcomp, 1);
-   ll_extend_md_node(cpu_llvm_module, rv, rv);
-   ll_extend_md_node(cpu_llvm_module, rv, loopUnroll);
-   cached_unroll_disable_metadata=rv;
+   cached_unroll_disable_metadata = ll_get_md_node(cpu_llvm_module,
+     LL_PlainMDNode, lvcomp, 1);
```
PeixinQiao, line 1033: Nit: Missing two spaces. The same as above.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -982,51 +1004,51 @@ cons_loop_parallel_accesses_metadata(void)
...
   LL_MDRef lvcomp[2];
-
-  lvcomp[0] = ll_get_md_string(cpu_llvm_module, "llvm.loop.vectorize.enable");
-  lvcomp[1] = ll_get_md_i1(1);
-  return ll_get_md_node(cpu_llvm_module, LL_PlainMDNode, lvcomp, 2);
+  if (LL_MDREF_IS_NULL(cached_vectorize_enable_metadata)) {
+    lvcomp[0] = ll_get_md_string(cpu_llvm_module, "llvm.loop.vectorize.enable");
+    lvcomp[1] = ll_get_md_i1(1);
+    cached_vectorize_enable_metadata = ll_get_md_node(cpu_llvm_module,
+      LL_PlainMDNode, lvcomp, 2);
```
PeixinQiao, line 1049: Nit: Missing two spaces. The same as above.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1309,50 +1331,62 @@ finish_routine(void)
...
   if (LL_MDREF_IS_NULL(cached_unroll_enable_metadata)) {
-    lvcomp[0] = ll_get_md_string(cpu_llvm_module, "llvm.loop.unroll.enable");
-    unroll= ll_get_md_node(cpu_llvm_module, LL_PlainMDNode, lvcomp, 1);
-    LL_MDRef md = ll_create_flexible_md_node(cpu_llvm_module);
-    ll_extend_md_node(cpu_llvm_module, md, md);
-    ll_extend_md_node(cpu_llvm_module, md, unroll);
-    cached_unroll_enable_metadata = md;
+   lvcomp[0] = ll_get_md_string(cpu_llvm_module, "llvm.loop.unroll.enable");
+   cached_unroll_enable_metadata = ll_get_md_node(cpu_llvm_module,
+     LL_PlainMDNode, lvcomp, 1);
```
PeixinQiao, line 1382: Nit: Missing two spaces. The same as above.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1777,34 +1828,34 @@ schedule(void)
...
         } else if (BIH_UNROLL_COUNT(bih)) {
-          LL_MDRef loop_md = cons_unroll_count_metadata(unroll_factor);
-          INSTR_LIST *i = find_last_executable(llvm_info.last_instr);
-          if (i) {
-            i->flags |= LOOP_BACKEDGE_FLAG;
-            i->misc_metadata = loop_md;
-          }
+          if (LL_MDREF_IS_NULL(loop_md))
+            loop_md = cons_loop_id_md();
+          ll_extend_md_node(cpu_llvm_module, loop_md, cons_unroll_count_metadata(unroll_factor));
```
PeixinQiao, line 1839: Nit: Exceed 80 columns.

tools/flang2/flang2exe/cgmain.cpp
bryanpkc: It's probably easier to run this hunk through clang-format, which would give:
```c
    cached_vectorize_disable_metadata =
        ll_get_md_node(cpu_llvm_module, LL_PlainMDNode, lvcomp, 2);
```

bryanpkc (approved):
LGTM, other than the formatting nits raised by @PeixinQiao. Our downstream build checked out fine too.

michalpasztamobica:
Many thanks for all the reviews, I just pushed a commit addressing @PeixinQiao 's findings. I did not squash it, as it is kind of Peter' code originally. I hope this is fine?

@kiranchandramohan , seems like all reviews are in place and we could try to merge this in?

bryanpkc:
@michalpasztamobica The formatting fixes should be squashed into the original commits that introduced those changes.

michalpasztamobica:
@bryanpkc , alright, squashed :)

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1777,34 +1828,35 @@ schedule(void)
...
           }
-        }
-        if (BIH_UNROLL(bih)) {
-          LL_MDRef loop_md = cons_unroll_metadata();
-          INSTR_LIST *i = find_last_executable(llvm_info.last_instr);
-          if (i) {
-            i->flags |= LOOP_BACKEDGE_FLAG;
-            i->misc_metadata = loop_md;
-          }
+	}
```
bryanpkc, line 1831: Incorrect indentation. @michalpasztamobica  Could you also fix this one?

kiranchandramohan:
> I would like to understand what fix_nodepcheck_flag does before completely removing it from the flow. Alternatively, introduce a fix_access_group_flag function in place of fix_nodepcheck_flag.

@bryanpkc what was your conclusion regarding this point?

tools/flang2/flang2exe/cgmain.cpp
michalpasztamobica: Done :)

michalpasztamobica:
@bryanpkc , @kiranchandramohan , it looks like Kiran's question is the last blocker standing in the way of merging this in? Any update on this?

bryanpkc:
As I mentioned at the last call, I tried to read the source and figure out what `fix_nodepcheck_flag` was doing, with not much success. I'll find time to dig some more and update this issue one way or another.

kiranchandramohan (approved):
LGTM.

PR #1097, opened 2021-06-22
Use ompstub by default on Windows on ARM

kaadam:
OpenMP is not supported on Windows on ARM yet.
Therefore this PR modified Cmakelist to use ompstub
on Windows on ARM for linking by default.


kaadam:
I will rebase it after #464 is merged

kaadam:
libomp library available for Windows on ARM (MSVC 14.29), this change is not necessary any more.

Issue #713, opened 2019-05-04
ICE: attempt to get size of assumed size character


nncarlson:
The following example gives this ICE with the March 2019 binary release:
```
F90-S-0000-Internal compiler error. size_of: attempt to get size of assumed size character       0  (flang-20190504.f90: 4)
```
```fortran
character(:), allocatable :: a(:)
class(*), allocatable :: b(:)
a = ['fubar']
allocate(b, source=a) ! <== ICE HERE
select type (b)
type is (character(*))
  if (len(b) /= 5) stop 1
  if (size(b) /= 1) stop 2
  if (b(1) /= 'fubar') stop 3
class default
  stop 4
end select
end
```

gklimowicz:
Thanks, Neil. We'll add it to the list. Sourced allocation is going to be a focus area for us soon. If you have more examples, please feel free to post them.

nncarlson:
Here's a variation of the example that more closely mimics the original code I'm working with. This one compiles fine but produces the wrong result. The source-allocated variable should be initialized with the string 'fubar', but instead gets a single 'f' followed by (what I think are) 4 null characters, and the code exits at the `STOP 3` statement.
```fortran
class(*), allocatable :: b(:)
call foo(['fubar'], b)
contains
  subroutine foo(a, b)
    class(*), intent(in)  :: a(:)
    class(*), allocatable :: b(:)
    allocate(b, source=a) ! <== VALUE OF B NOT CORRECTLY DEFINED
    select type (b)
    type is (character(*))
      if (len(b) /= 5) stop 1
      if (size(b) /= 1) stop 2
      if (b(1) /= 'fubar') stop 3
    class default
      stop 4
    end select
  end subroutine
end
```

nncarlson:
For completeness, here's a rank-2 matrix version of the previous example (again it closely mimics original failing code). This also compiles okay, but produces the wrong result and exits on the `STOP 3`
```fortran
character(:), allocatable :: ref(:,:)
class(*), allocatable :: b(:,:)
ref = reshape(['foo','bar'], shape=[1,2])
call foo(ref, b)
contains
  subroutine foo(a, b)
    class(*), intent(in)  :: a(:,:)
    class(*), allocatable :: b(:,:)
    allocate(b(size(a,1),size(a,2)), source=a) ! <== VALUE OF B NOT CORRECTLY DEFINED
    !allocate(b, source=a) ! unspecified bounds allowed by F2008
    select type (b)
    type is (character(*))
[...]

PR #1099, opened 2021-06-23
Fix the ICE caused by source-allocating a class(*) array from an array of deferred length character strings. See issue #713.

qiaozhang-hnc:
It also fixes the length parameter not being correctly assigned problem mentioned at the end of issue #713.

This patch piggybacks the logic to handle source-allocating a class(*) array from an array of deferred length character strings on the code handling source-allocating an unlimited_polymorphic array from a string.

### DIFF
--- a/test/f90_correct/inc/alloc_unpoly_str_01.mk
+++ b/test/f90_correct/inc/alloc_unpoly_str_01.mk
@@ -1,0 +1,22 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+$(TEST): run
+
+
+build:  $(SRC)/$(TEST).f90
+	-$(RM) $(TEST).$(EXESUFFIX) core *.d *.mod FOR*.DAT FTN* ftn* fort.*
+	@echo ------------------------------------ building test $@
+	-$(FC) -c $(FFLAGS) $(LDFLAGS) $(SRC)/$(TEST).f90 -o $(TEST).$(OBJX)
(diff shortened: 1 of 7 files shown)


PeixinQiao:
The following test case should be added since this patch revises the code handling source-allocating an unlimited_polymorphic variable from a deferred char.
```
program test
  implicit none
  class(*), allocatable :: a
  character(:), allocatable :: b
  character(:), allocatable :: c

  b = 'ab'
  c = 'abc'
  if (len(b) /= 2) stop 1

  allocate(a, source = b)

  select type (a)
  type is (character(*))
    if (len(a) /= len(b)) stop 2
    if (a /= b) stop 3
  class default
    stop 4
  end select
  deallocate(a)

  call sub(a, c)
  print *, 'PASS'

contains
  subroutine sub(x, y)
    class(*), allocatable :: x
    character(:), allocatable :: y

    allocate(character(len(y)) :: x)
    select type (x)
    type is (character(*))
      if (len(x) /= len(y)) stop 5
    class default
      stop 6
    end select
    deallocate(x)
  end subroutine
end
```

PeixinQiao:
What logic did you piggyback is not clear in commit message. I think what you revise in this patch are the following parts.

```
Use DDTG(dtype) instead of dtype to determine the source type of a
polymorphic allocation, in order to accommodate a string array as well
as a single string.

To prevent an internal compiler error, do not call chkstruct if the
element type of an array-type source is deferred-length string, and do
not set rank, lsize, gsize, and kind for the allocatable destination.
```

qiaozhang-hnc:
@PeixinQiao Test cases and commit description updated as suggested. Please have a look.

PeixinQiao:
LGTM.

pawosm-arm (approved):

shivaramaarao (approved):

bryanpkc (approved):

PR #1100, opened 2021-06-23
Fix 2 bugs related to passing a polymorphic array as an argument to a function

qiaozhang-hnc:
This patch fixes 2 errors:
1. When a polymorphic array is passed as an argument, flang directly uses the element type in the declaration statement of the dummy array to create the function calling node.
2. When a section of a polymorphic array is passed as an argument, flang doesn't get the first element correctly.

This patch fixes the 1st bug by adding a new ast node to obtain the actual element type if necessary and it fixes the 2nd error by calling `get_poly_element_arg` to get the correct first element.

### DIFF
--- a/test/f90_correct/inc/array_section.mk
+++ b/test/f90_correct/inc/array_section.mk
@@ -1,0 +1,26 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+########## Make rule for test array_section ########
+
+
+array_section: run
+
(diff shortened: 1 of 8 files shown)


bryanpkc (approved):

shivaramaarao (approved):

bryanpkc:
@pawosm-arm I don't think issues were opened for these bugs. Do you want @qiaozhang-hnc to create issues for documentation purpose, or can this PR be merged as-is?

pawosm-arm:
@bryanpkc any kind of reference (e.g. in commit message) what benchmarks were actually failing would do.

qiaozhang-hnc:
Hi guys. I've added a reference to the test case that exposes the bugs fixed by the PR. Have a good one!

bryanpkc:
I missed the fact that we already got three approvals for this PR. Merging.

PR #1101, opened 2021-06-23
Fix errors related to empty structure constructors nested in an array constructor.

qiaozhang-hnc:
The problem occurs：
    1.When an array constructor with an empty structure constructor is passed as an argument to a function，flang2 will generate wrong IR to use an undefined type named 'struct.STATICS*';
    2.When an array constructor has several empty structure constructors, semantic analysis handling the empty structure construct will be prevented with bogus error "F90-S-0069-Illegal implied DO expression".

This patch fixes the 1st bug by skipping over unnecessary processing when the element type is an empty drived data type in flang2. This patch fixes the 2nd bug by not processing the components in the function `dinit_data` from file 'dinit.c' and returning the real ACL pointer instead of 0 in `rewrite_acl` from file 'semutil2.c'.

### DIFF
--- a/test/f90_correct/inc/empty_type_01.mk
+++ b/test/f90_correct/inc/empty_type_01.mk
@@ -1,0 +1,20 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+$(TEST): run
+
+build:  $(SRC)/$(TEST).f90
+	-$(RM) $(TEST).$(EXESUFFIX) core *.d *.mod FOR*.DAT FTN* ftn* fort.*
+	@echo ------------------------------------ building test $@
+	-$(CC) -c $(CFLAGS) $(SRC)/check.c -o check.$(OBJX)
+	-$(FC) -c $(FFLAGS) $(LDFLAGS) $(SRC)/$(TEST).f90 -o $(TEST).$(OBJX)
+	-$(FC) $(FFLAGS) $(LDFLAGS) $(TEST).$(OBJX) check.$(OBJX) $(LIBS) -o $(TEST).$(EXESUFFIX)
+
+run:
+	@echo ------------------------------------ executing test $(TEST)
+	$(TEST).$(EXESUFFIX)
+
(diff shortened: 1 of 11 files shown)


bryanpkc:
@qiaozhang-hnc Could you rebase the branch?

pawosm-arm (approved):
There are overhead empty lines in empty_type_01.mk (with one at the end of file) that could be removed, but otherwise, it looks OK.

qiaozhang-hnc:
> There are overhead empty lines in empty_type_01.mk (with one at the end of file) that could be removed, but otherwise, it looks OK.

The empty lines have been removed.

bryanpkc (approved):
We have tested/integrated this in our downstream compiler.

bryanpkc:
@shivaramaarao Do you want to take a look at this?

kiranchandramohan:
@shivaramaarao can look into it this week.

pawosm-arm:
@qiaozhang-hnc can you look at this?

pawosm-arm:
I've looked into the generated `tx7a.mod` file. Turns out, invalid ICT's are being generated. Normally for this module two ICT codes are emitted: 'R' and 'A'. After applying this patch, two different codes are emitted instead, 'S' and 'J', where 'J' is unknown code, hence `getict()` function (defined in the `interf.c` file) terminates in the `default:` case of the `switch (p[0])`.

@bryanpkc should a new issue with regression report be opened for this problem?

qiaozhang-hnc:
> @qiaozhang-hnc can you look at this?

Hi @pawosm-arm , I looked into the regression, but couldn't reproduce the problem. Could you provide me with the test case that exposes the error? Thanks.

pawosm-arm:
@qiaozhang-hnc what architecture are you working with? I can't reproduce it on x86_64, but I can reproduce it on AArch64 with disabled quadfp. I'm using upstream open-source compiler.

pawosm-arm:
I've also reproduced it on AArch64 with enabled quadfp. Somehow x86_64 makes all of the difference.

qiaozhang-hnc:
> @qiaozhang-hnc what architecture are you working with? I can't reproduce it on x86_64, but I can reproduce it on AArch64 with disabled quadfp. I'm using upstream open-source compiler.

I have reproduced this regression with the simplified case in my AArch64 machine. I will fix it ASAP.

PR #1105, opened 2021-06-25
[DebugInfo] Suppress debugloc for actual argument bitcast

PeixinQiao:
This patch fixes the bug in debug location when using gdb and addr2line dump the line number. Compile the code and run it with gdb 10.1 as follows:
```
program example
  implicit none

  call func()
contains
  subroutine func()
    call func1()
    call func2(2)
  end
  subroutine func1()
    integer :: i = 1
    call func3()
    print *, i
  end
  subroutine func2(j)
    integer :: j, k
    k = j
  end
  subroutine func3()
  end
end
```

```
$ flang case.f90 -g && gdb ./a.out
...
(gdb) b func3
...
(gdb) run
...
(gdb) bt
#0  example::func3 () at case.f90:30
#1  0x00000000004009b8 in example::func1 () at case.f90:22
#2  0x0000000000400988 in example::func () at case.f90:17
#3  0x0000000000400968 in example () at case.f90:14
#4  0x0000000000400a98 in main (argc=1, argv=0xffffffffdfa8) at /home/qpx/llvm-project/flang/runtime/flangmain/flangmain.c:59
#5  0x0000ffffbedbfae0 in __libc_start_main () from /lib64/libc.so.6
#6  0x0000000000400884 in _start ()
```

Then, open another terminal and run the following command:
```
$ addr2line -p -f -e /home/qpx/case1/a.out 0x00000000004009b8
example_func1 at /home/qpx/case1/case.f90:?
```

The line number info is missing. To look at the assembly code, I find out that the bitcast debug info of `print *` causes the problem, which is  ` .loc    1 0 0 is_stmt 0         // case.f90:0:0`.

```
        .loc    1 22 1 prologue_end             // case.f90:22:1
        bl      example_func3
        .loc    1 0 0 is_stmt 0                 // case.f90:0:0
        adrp    x0, .C304_example_func1
        add     x0, x0, :lo12:.C304_example_func1
        adrp    x1, .C302_example_func1
        add     x1, x1, :lo12:.C302_example_func1
        mov     w8, #8
        mov     w2, w8
        .loc    1 23 1 is_stmt 1                // case.f90:23:1
        bl      f90io_src_info03a
        .loc    1 0 0 is_stmt 0                 // case.f90:0:0
        adrp    x0, .C305_example_func1
        add     x0, x0, :lo12:.C305_example_func1
        mov     x1, xzr
        adrp    x3, .C283_example_func1
        add     x3, x3, :lo12:.C283_example_func1
        .loc    1 23 1                          // case.f90:23:1
        mov     x2, x3
        bl      f90io_print_init
```

So, I think flang should not generate debug location metadata for bitcast instructions on actual arguments passed to a procedure call.

### DIFF
--- a/test/debug_info/procedure_argument.f90
+++ b/test/debug_info/procedure_argument.f90
@@ -1,0 +1,31 @@
+!RUN: %flang -g -S -emit-llvm %s -o - | FileCheck %s
+
+! check function pointer bitcast instructions should not have debug location
+!CHECK-NOT: bitcast i32* @.C{{[0-9]+}}_example_func to i64*, !dbg {{![0-9]+}}
+!CHECK-NOT: bitcast i32* @.C{{[0-9]+}}_example_func1 to i8*, !dbg {{![0-9]+}}
+!CHECK-NOT: bitcast [8 x i8]* @.C{{[0-9]+}}_example_func1 to i8*, !dbg {{![0-9]+}}
+!CHECK-NOT: bitcast void ()* @f90io_src_info03a to void (i8*, i8*, i64)*, !dbg {{![0-9]+}}
+!CHECK-NOT: bitcast i32 ()* @f90io_print_init to i32 (i8*, i8*, i8*, i8*)*, !dbg {{![0-9]+}}
+!CHECK-NOT: bitcast i32 ()* @f90io_sc_i_ldw to i32 (i32, i32)*, !dbg {{![0-9]+}}
+
+program example
+  implicit none
+
+  call func()
+contains
+  subroutine func()
+    call func1()
+    call func2(2)
+  end
+  subroutine func1()
+    integer :: i = 1
+    call func3()
+    print *, i
+  end
+  subroutine func2(j)
+    integer :: j, k
+    k = j
+  end
+  subroutine func3()
+  end
+end
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -2828,6 +2828,33 @@ should_suppress_debug_loc(INSTR_LIST *instrs)
   }
 }
 
+/* whether instruction is procedure argument bitcast */
+static bool
+is_proc_arg_bitcast(INSTR_LIST *instrs)
+{
+  if (!instrs)
+    return false;
+
+  if (instrs->i_name == I_BITCAST) {
+    INSTR_LIST *call_instr;
+    OPERAND *operand;
+    /* search for call function instruction */
+    for (call_instr = instrs->next; call_instr; call_instr = call_instr->next) {
+      if (call_instr->i_name == I_CALL) {
+        /* match the instruction with the procedure argument */
+        for (operand = call_instr->operands; operand; operand = operand->next) {
+          if (operand->ot_type == OT_TMP && operand->tmps == instrs->tmps &&
+              instrs->operands->ot_type == OT_VAR) {
+            return true;
+          }
+        }
+      }
+    }
+  } else {
+    return false;
+  }
+}
+
 /**
    \brief Write the instruction list to the LLVM IR output file
  */
@@ -3368,11 +3395,13 @@ write_instructions(LL_Module *module)
      *  - it is already written (dbg_line_op_written) or
      *  - it is a known internal (f90 runtime) call in prolog (fort_init &
      * f90_*)
+     *  - it is procedure argument
      */
     if (!(LL_MDREF_IS_NULL(instrs->dbg_line_op) || dbg_line_op_written ||
           ((instrs->dbg_line_op ==
             lldbg_get_subprogram_line(module->debug_info)) &&
-           should_suppress_debug_loc(instrs)))) {
+           should_suppress_debug_loc(instrs)) ||
+          ((instrs->i_name == I_BITCAST) && is_proc_arg_bitcast(instrs)))) {
       print_dbg_line(instrs->dbg_line_op);
     }
 #if DEBUG


SouraVX:
@kiranchandramohan could you please approve CI run for this PR. Thanks!

kiranchandramohan:
Invited @PeixinQiao to this repo with triage level access.

kiranchandramohan:
Could you share the file case.f90 as well?

PeixinQiao:
@kiranchandramohan Sorry for the imprecise explanation. Add the `case.f90` above, which is also the test case `test/debug_info/procedure_argument.f90` in this patch.

jinisusan:
@PeixinQiao, I have not been able to reproduce this issue. Do you mind cross checking ?

PeixinQiao:
@jinisusan I tested it on aarch64, and build flang using the commits as follows
llvm-project: 5b601378fcb0f78e880068b31c3e84f8756d0cba
flang: b97f949a8c7a99e3855113aa2720979621f80a85

What is your test results of the assembly code for `func`? Does it contain the statement of `.loc    1 0 0 is_stmt 0                 // case.f90:0:0`?

jinisusan:
@PeixinQiao, no, the assembly file that I obtained does not contain the particular .loc directive (.loc 1 0 0 is_stmt 0 // case.f90:0:0) as you mentioned above. 

One point which I did not understand is that in the gdb backtrace you pointed out above, we seem to be having the line table information for address  0x00000000004009b8 in  example::func1() as seen below (since the source position information (case.f90:22) is displayed). 
#1  0x00000000004009b8 in example::func1 () at case.f90:22

Would that be an addr2line issue by any chance? 

Could you pls send the IR file for this test case before and after your fix ?

test/debug_info/procedure_argument.f90
```diff
@@ -0,0 +1,31 @@
+!RUN: %flang -g -S -emit-llvm %s -o - | FileCheck %s
+
+! check non debug instructions should not have debug location
```
bryanpkc, line 3: Could you clarify here that the symbols being bitcast below are static variables or literals? It would be nice to increase coverage to actual arguments that are local variables or dummy arguments.

PeixinQiao:
@jinisusan I think gdb and addr2line use different method for backtracing function. Making the results of running addr2line correct is necessary.

jinisusan:
@PeixinQiao wouldn't that point to an addr2line issue rather than an flang issue ?

PeixinQiao:
@jinisusan When looking at the assembly code by compiling the IR `case-before-fix.txt`, the statement `.loc    1 0 0 is_stmt 0                 // case.f90:0:0` is unreasonable. Here is the assembly code I generated.
[case-before-fix.txt](https://github.com/flang-compiler/flang/files/6746083/case-before-fix.txt)

jinisusan:
Thank you for providing the files and the input, @PeixinQiao. I was able to reproduce the issue by providing -global-isel option to the backend. 

But IMHO, it is perfectly fine to have the .loc 1 0 0 is_stmt  directive (i.e., which translates to a line number of zero in the debug line table) . The backend produces this with the -global-isel  option (which would be the default in your case). 

The DWARFv5 standard (page 151) states the following wrt "line" in line number information state machine: 

An unsigned integer indicating a source line number. Lines are numbered beginning at 1. **The compiler may emit the value 0 in cases where an instruction cannot be attributed to any source line.**

Since this is a valid case as per the standard, it should be upto the tools to handle such cases appropriately. In gdb, a fix was made to ignore these entries in July 2020 with the following commit: 

commit 876518dd0a0b6fd6f4ad0a0b247db0d6a267db27
Author: Tom de Vries <tdevries@suse.de>
Date:   Sat Jul 25 00:23:06 2020 +0200

Relevant code change for this: 

diff --git a/gdb/dwarf2/read.c b/gdb/dwarf2/read.c
index 39ed455def5..4e70dccecbc 100644
--- a/gdb/dwarf2/read.c
+++ b/gdb/dwarf2/read.c
@@ -20437,8 +20437,9 @@ lnp_state_machine::record_line (bool end_sequence)
          bool file_changed
            = m_last_subfile != m_cu->get_builder ()->get_current_subfile ();
          bool ignore_this_line
-           = (file_changed && !end_sequence && m_last_address == m_address
-              && !m_is_stmt && m_stmt_at_address);
+           = ((file_changed && !end_sequence && m_last_address == m_address
+               && !m_is_stmt && m_stmt_at_address)
+              || (!end_sequence && **m_line == 0)**);

More details can be found for this at: https://sourceware.org/bugzilla/show_bug.cgi?id=26243 

So, I believe that the binutils addr2line needs to be fixed to handle this. FWIW, the llvm-addr2line handles this better (it reports the line number as 0 -- just as it is in the line table): 

$**llvm-addr2line** -p -f -e ./case  0x0000000000201b21
example_func1 at <path>/case.f90:**0**
$ **addr2line** -p -f -e ./case  0x0000000000201b21
example_func1 at /<path>/case.f90:?

Thanks to @SouraVX and @alokkrsharma for their inputs for this.

PeixinQiao:
@jinisusan Thanks so much for providing the detailed information to fix the line number from `?` to `0`.
It seems that the line number of 0 is still not reasonable. Let's look at the similar c code:
```
#include <stdio.h>
  
void func3() {}

void func2(int j) {
  int k;
  k = j;
}

void func1() {
  int i = 1;
  func3();
  printf("%d\n", i);
}

void func() {
  func1();
  func2(2);
}

int main() {
  func();
}
```
The assembly code for `func` is as follows:
```
func:                                   // @func
...
.Ltmp6:
        .loc    1 17 3 prologue_end             // case.c:17:3
        bl      func1
        mov     w0, #2
        .loc    1 18 3                          // case.c:18:3
        bl      func2
        .loc    1 19 1                          // case.c:19:1
        ldp     x29, x30, [sp], #16             // 16-byte Folded Reload
        ret
```
Running the following code and command:
```
$clang case.c -g && gdb ./a.out
...
$ b func3
...
$ run
...
$ bt
#0  func3 () at case.c:3
#1  0x0000000000400668 in func1 () at case.c:12
#2  0x0000000000400690 in func () at case.c:17
#3  0x00000000004006ac in main () at case.c:22
(gdb) quit
...
$ llvm-addr2line -p -f -e ./a.out 0x0000000000400668
func1 at /home/.../case.c:13
```

Compared with `clang` frontend, `flang` generates the additional `.loc 1 0 0 is_stmt 0` after `bl example_func1`, which is not reasonable.

It seems that `gdb` use the `.loc` above the `bl function` for tracing functions, while `add2line` use  the `.loc` below the `bl function` for tracing functions. The reason why I want to use `addr2line` is that sometimes `gdb` does not work well when running some workloads using multiple processes. One solution is to write some `c` code using `addr2line` for finding function line numbers and interoprate with the workload.

test/debug_info/procedure_argument.f90
PeixinQiao: Updated the test case and commit message. The symbols being bitcast are function pointers.
The above explaination of the corresponding assembly code is not correct, which should be `example_func3` instead of `example_func1`. Updated now. The test case contains the arguments of constant value `2` and local variable `i`.

```
$ llvm-addr2line -p -f -e ./a.out 0x0000000000400988
example_func at /home/.../case.f90:0
$ llvm-addr2line -p -f -e ./a.out 0x00000000004009b8
example_func1 at /home/.../case.f90:0
```

jinisusan:
@PeixinQiao, IMHO, having the compiler emit the line number 0 in the debug information is perfectly reasonable. This is also done by clang and is permitted by the DWARF standard. You can check the test case provided at [https://sourceware.org/bugzilla/show_bug.cgi?id=26243](https://sourceware.org/bugzilla/show_bug.cgi?id=26243). If you compile the test case provided there with clang++, you can notice .loc directives with line numbers specified as 0. I would re-iterate that it is upto the debug info consumer tools to deal with this the right way.

PeixinQiao:
@jinisusan Thanks for your advice. I will close this PR for now since there is no big deal in current flang usage without this patch.

PR #1109, opened 2021-07-16
[flang1] Fix bug in TK_G0FORMAT token pattern matching (encountered in workloads WAVEWATCH III and CMAQ)

PeixinQiao:
The following test case is generated from the workload WAVEWATCH III 6.07.1 (https://github.com/NOAA-EMC/WW3) and CMAQ 4.7.1 (https://github.com/USEPA/CMAQ).
```
program example
   integer :: i = 1, j = 2, k = 3
   character(len=2) :: expect(3) = (/'1 ', ' 2', ' 3'/)
   character(len=2) :: res

   write(res, 10) i
   if (res .ne. expect(1)) stop 1
   write(res, 20) j
   if (res .ne. expect(2)) stop 2
   write(res, 30) k
   if (res .ne. expect(3)) stop 3
   print *, 'PASS'

10 FORMAT (I0)
20 FORMAT (I02)
30 FORMAT (I2)
end
```
In format specification, the token type TK_G0FORMAT is for I0, not for I02 or I002, or something like that. Fix the token pattern matching of TK_G0FORMAT by checking if the next identifier of '0' in id buffer is null.

### DIFF
--- a/test/f90_correct/src/format_specification.f90
+++ b/test/f90_correct/src/format_specification.f90
@@ -1,0 +1,28 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Checking for the token type "TK_G0FORMAT" (I0) and
+! the special case of the token type "I" + "<integer>" (I02),
+! while the normal case is I2.
+
+program example
+   integer :: i = 1, j = 2, k = 3
+   character(len=2) :: expect(3) = (/'1 ', ' 2', ' 3'/)
+   character(len=2) :: res
+
+   write(res, 10) i
+   if (res .ne. expect(1)) stop 1
+   write(res, 20) j
+   if (res .ne. expect(2)) stop 2
+   write(res, 30) k
+   if (res .ne. expect(3)) stop 3
+
+   print *, 'PASS'
+
+10 FORMAT (I0)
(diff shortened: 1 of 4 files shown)


test/f90_correct/src/format_specification.f90
```diff
@@ -0,0 +1,28 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Checking for the token type "TK_G0FORMAT" (I0) and
+! the special case of the token type "I" + "<integer>" (I02),
+! while the normal case is I2.
+
+program example
```
pawosm-arm, line 10: There are complains in other reviews for giving the programs generic names like `test` or `example`. Personally, I'm not against, but others may ask you to change it.

pawosm-arm (approved):

test/f90_correct/src/format_specification.f90
bryanpkc: I think it's modules with non-unique names that could cause problems.

bryanpkc (approved):
LGTM.

PeixinQiao:
ping @shivaramaarao

shivaramaarao (approved):

kiranchandramohan:
@bryanpkc Looks like this has all the approvals.

PeixinQiao:
@bryanpkc Can we merge this now?

bryanpkc:
Sorry, I forgot about this.

PR #1110, opened 2021-07-26
Fix adding ST_MODPROC case in function mkvarref()

### LINKED ISSUE #1016: dbcsr: flang did not notice that pure function defined later in the 'contains' section is pure
Consider the following piece of Fortran code:

```
module mymod
  implicit none
  private
  public :: func, other_func
  type :: foo
    integer :: mydim
  end type

contains

subroutine func(tparam, oparam)
  type(foo), intent(in) :: tparam
  integer, dimension(other_func(tparam, 1)), intent(out), optional :: oparam

  oparam = 42
end subroutine

pure function other_func(tparam, idim)
  type(foo), intent(in) :: tparam
  integer, intent(in) :: idim
  integer :: other_func

  other_func = min(tparam%mydim, idim)
end function

end module
```

Note that gfortran can compile this code, while flang can not:

```
F90-S-0084-Illegal use of symbol other_func  (mymod.f90: 13)
  0 inform,   0 warnings,   1 severes, 0 fatal for func
```

Yet when the pure function `other_func` definition is moved before it is used (in a place where a pure function should be called), flang is able to compile it:

```
contains

pure function other_func(tparam, idim)
  type(foo), intent(in) :: tparam
  integer, intent(in) :: idim
  integer :: other_func

  other_func = min(tparam%mydim, idim)
end function

subroutine func(tparam, oparam)
  type(foo), intent(in) :: tparam
  integer, dimension(other_func(tparam, 1)), intent(out), optional :: oparam

  oparam = 42
end subroutine
```

abrahamtovarmob:
Adding ST_MODPROC to switch() options in mkvarref() to fix an issue
relating Pure Function not being properly detected.

The issue occours because the pure function type has assignated
ST_MODPROC and is not recognized.
In the case of the Workaround, the type is ST_PROC. Debug reveald
that the assignated types are correct, but the probem is that
ST_MODPROC has no action defined, hence an error in the compiler.
The soluton is to add ST_MODPROC into the same section as ST_PROC
in order to perfom the same actions

Fixes #1016

### DIFF
--- a/test/f90_correct/src/pure_function.f90
+++ b/test/f90_correct/src/pure_function.f90
@@ -1,0 +1,54 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+
+! Check that the type order between the declaration of "pure function" and its implementation
+! doesnt produce a compilation error
+! Test related to fix for issue #1016
+
+module pureFunction
+  implicit none
+  private
+  public :: func, other_func
+  type :: foo
+    integer :: mydim
+  end type
+
+contains
+
+subroutine func(tparam, oparam)
+  integer, intent(in) :: tparam
+  integer, dimension(other_func(tparam, 10)), intent(out), optional :: oparam
+
+  oparam = 42
+end subroutine
+
+pure function other_func(tparam, idim)
+  integer, intent(in) :: tparam
+  integer, intent(in) :: idim
+  integer :: other_func
+
+  other_func = min(tparam, idim)
+end function
+
+end module
+
+program testPureFunction
+	use pureFunction
+	implicit none
+
+	integer :: result(1)
+	integer :: expect
+	integer :: param1
+  
+
+	param1 = 20	
+	call func(param1,result)
+	expect = 42
+
+	call check(result, expect, 1)
+	print *,result, expect, param1
+
+end program
--- a/test/f90_correct/inc/pure_function.mk
+++ b/test/f90_correct/inc/pure_function.mk
@@ -1,0 +1,19 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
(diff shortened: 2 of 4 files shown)


michalpasztamobica:
@pawosm-arm , would you please double check the f90 test? It is a little bit simplified, compared to your reproducer (the type of the tparam is not `foo`, but `integer` instead). Abraham was still able to reproduce the issue without the fix. We had trouble using a private type from the main program, to be honest :)

pawosm-arm:
I'll build dbcsr and cp2k with it to make sure cp2k can make use of the dbcsr modules without any workarounds we had to use. That should prove this fix is ok.

pawosm-arm:
With this fix, I've successfully managed to build dbcsr with the corresponding workaround reverted (namely, CP2K can make use of it). Also, number of passing test cases did not drop. Therefore I can say that this fix does its job.

pawosm-arm (approved):

pawosm-arm:
@bryanpkc this is something we were waiting for for some long time. Can you look at this, please?

kiranchandramohan:
Please provide a description in the commit message. You can consider moving some of the points mentioned in 1016 to the description.

abrahamtovarmob:
> Please provide a description in the commit message. You can consider moving some of the points mentioned in 1016 to the description.

Done. Can you approve the workflow for CI to be triggered?

test/f90_correct/src/pure_function.f90
```diff
@@ -0,0 +1,44 @@
+module mymod
```
bryanpkc, line 1: Could you add a comment in this test case to indicate that this is a reproducer for #1016, and that the procedures `func` and `other_func` should never be reordered?

bryanpkc: Alternatively, rename the test case to `issue1016.f90`.

Also, I suggest using more unique names for the module `mymod` and the program `test`. We have run into test failures previously when module names collided.

pawosm-arm: I wouldn't rename the file to point at github issue. Let's not create a precedence.

bryanpkc: Would you object to mentioning the issue inside a comment?

abrahamtovarmob: Suggested changes added to the test. Also rebase for new master commits

abrahamtovarmob:
Can somebody approve the workflow? since I'm new I need a contributor to approve

bryanpkc (approved):

PR #1112, opened 2021-08-04
flang2: handle atomic load of logical variables of various size correctly

pawosm-arm:
This issue was haunting our OpenMP ATOMIC examples involving LOGICAL types. Finally I've found some time to fix it for good.


kiranchandramohan:
Upstream has atomic off by default. We will have to enable the option to test.

pawosm-arm:
Seems this is incomplete.

pawosm-arm:
Now it should work...

pawosm-arm:
cannot reopen! will open a new one.

PR #1113, opened 2021-08-04
flang2: handle atomic load of logical variables of various size correctly

pawosm-arm:
This issue was haunting our OpenMP ATOMIC examples involving LOGICAL types. Finally I've found some time to fix it for good.


pawosm-arm:
@kiranchandramohan we may not be able to find reviewers for the commit that affects native atomics only!

bryanpkc (approved):
We have reviewed this patch; it looks good to us.

Issue #896, opened 2020-05-18
Multiple use of ISO_FORTRAN_ENV causes flang to error


RichBarton-Arm:
The following example causes flang to emit an error. gfortran handles this fine.

> F90-S-0038-Symbol, compiler_version, has not been explicitly declared (single_source.F90)

This example is reduced from code in the [CP2k application](https://www.cp2k.org/).

As pointed out in the comments, it is the double use of _ISO_FORTRAN_ENV_ via an additional used module with use of _IMPLICIT_NONE_ that causes the issue.

```
MODULE input_section_types
   ! Removing this use statement fixes the issue and flang does the right thing.
   USE ISO_FORTRAN_ENV
END MODULE input_section_types

PROGRAM cp2k
   ! Removing the below use statement makes flang do the right thing.
   use input_section_types
   ! Removing the below line gives the same error (expected)
   USE ISO_FORTRAN_ENV, ONLY: compiler_version
   ! Removing implicit none lets flang build a working example.
   IMPLICIT NONE
   WRITE (*,*) "other: ", compiler_version()
END PROGRAM
```

michalpasztamobica:
I am first getting and information code 0035:
```
(lldb) bt
* thread #1, name = 'flang1', stop reason = breakpoint 4.1
  * frame #0: 0x00000000007d5db1 flang1`error(ecode=I_0035_Predefined_intrinsic_OP1_loses_intrinsic_property, sev=ERR_Informational, eline=13, op1="compiler_version", op2=0x0000000000000000) at error.c:226
    frame #1: 0x000000000072efee flang1`newsym(sptr=607) at semsym.c:1212
    frame #2: 0x00000000006e58cd flang1`ref_intrin(stktop=0x0000000000b43040, list=0x0000000000000001) at semfunc.c:5435
    frame #3: 0x00000000007455cf flang1`mkvarref(stktop=0x0000000000b43040, list=0x0000000000000001) at semutil.c:2127
    frame #4: 0x00000000006894d0 flang1`semant2(rednum=700, top=0x0000000000b43040) at semant2.c:1247
    frame #5: 0x00000000005e3823 flang1`_parser at parser.c:302
    frame #6: 0x00000000005e3266 flang1`parser at parser.c:152
    frame #7: 0x00000000005af7fc flang1`main(argc=2, argv=0x00007fffffffdfa8) at main.c:208
[...]

PR #1114, opened 2021-08-05
[Flang1] Fix bug in exporting iso_fortran_env (encountered in CP2K #896)

PeixinQiao:
The following test case is generated from the workload CP2K and reported by issue #896.
```
module usermod
  use iso_fortran_env
end module usermod

program cp2k
  use usermod
  use iso_fortran_env, only: compiler_version, compiler_options
  implicit none
  character(len=:), allocatable :: version_var
  character(len=:), allocatable :: options_var
  version_var = compiler_version()
  options_var = compiler_options()
  print *, version_var
  print *, options_var
end program cp2k
```
The functions `compier_version` and `compier_options` are exported in `iso_c_binding` library, which is shown in `usermod.mod` when compiling the code. The stype of the two functions is set to `ST_INTRIN` when they are imported by entering the if branch of `iso_c_binding`, which leads to the error.
```
$ flang test.f90
(error message)
$ head -n 10 test.f90
...
B 606 iso_c_binding compiler_options
B 607 iso_c_binding compiler_version
...
```

### DIFF
--- a/test/llvm_ir_correct/iso_fortran_env_functions.f90
+++ b/test/llvm_ir_correct/iso_fortran_env_functions.f90
@@ -1,0 +1,26 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Simply check the file compiles with no error.
+!
+! RUN: %flang -S -emit-llvm %s -o - | FileCheck %s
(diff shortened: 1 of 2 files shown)


PeixinQiao:
@RichBarton-Arm @michalpasztamobica This patch fixes the failed example you reported in #896.

bryanpkc (approved):
LGTM.

pawosm-arm (approved):
great!

pawosm-arm:
ping...

shivaramaarao (approved):
LGTM

PR #1117, opened 2021-08-06
win: Define ssize_t on Windows

kaadam:
ssize_t is not available type on Windows, so
it needs to use SSIZE_T from Windows Data Types.

### DIFF
--- a/runtime/include/memops.h
+++ b/runtime/include/memops.h
@@ -14,7 +14,12 @@ extern "C" {
 #endif
 
 #if defined(INLINE_MEMOPS)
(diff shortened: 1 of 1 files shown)


isuruf (approved):
It looks like the same workaround is used in `utils/errmsg/errmsg.cpp`, so +1 from me.

kiranchandramohan (approved):
LGTM.
@bryanpkc if you are OK, please merge.

bryanpkc (approved):

bryanpkc (approved):

PR #1118, opened 2021-08-09
win: include winsock header for timeval structure

kaadam:
On Windows timeval structure comes from 'winsock.h'.

### DIFF
--- a/runtime/flang/miscsup_com.c
+++ b/runtime/flang/miscsup_com.c
@@ -18,6 +18,8 @@
 #ifndef _WIN64
 #include <sys/time.h>
 #include <unistd.h>
+#else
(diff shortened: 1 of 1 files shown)


isuruf (approved):

kiranchandramohan:
File runtime/flang/stat_linux.c also has some usage of timeval, but that seems to be conditionally enabled for non-Windows platforms. 
Should we have the same handling here as well? Or is just including the header enough? Will the usage of timeval work correctly on Windows?

kaadam:
@kiranchandramohan could you look at it again? Thanks.

kaadam:
@bryanpkc , @pawosm-arm May I ask one of you to merge this as well?

PR #1122, opened 2021-08-15
[flang2] Fix member_with_offset in outliner.cpp

bryanpkc:
Commit fdd777cee941eb7df368a54abad2de4dff744194 added some code that creates a NME for an address in the uplevel table (conceptually a struct with many anonymous fields) if one does not exist, in order to make the dumps easier to read (presumably there would otherwise be many stores associated with the same NME, which would be confusing). However, instead of looking for a struct member with a matching offset, the function `member_with_offset` clobbers the `ADDRESS` field for the first `SPTR` in the given linked list of members. This patch corrects the function, and also fixes some formatting problems in the code that calls `member_with_offset`.

### DIFF
--- a/tools/flang2/flang2exe/outliner.cpp
+++ b/tools/flang2/flang2exe/outliner.cpp
@@ -1422,18 +1422,17 @@ handle_nested_threadprivate(LLUplevel *parent, SPTR uplevel, SPTR taskAllocSptr,
   }
 }
 
-/*
- * given a member of a struct datatype and an offset,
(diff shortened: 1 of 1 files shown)


pawosm-arm (approved):

kiranchandramohan (approved):
LGTM.

bryanpkc:
ping @shivaramaarao

bryanpkc:
ping @shivaramaarao

shivaramaarao (approved):

PR #1123, opened 2021-08-18
[ArgParser] Use const char * for command-line arguments

bryanpkc:
ArgParser returns pointers to elements in the `argv` array; such pointers should be declared `const` to prevent clobbering of the command-line arguments. Declaring the recipient variables `const` also allows flang1/flang2's `init` function to initialize such variables with string literals.

This patch fixes a number of `-Wcast-qual` warnings by changing the ArgParser API to require `const char *` variables, and updating `tools/flang1/flang1exe/main.c` and `tools/flang2/flang2exe/main.cpp` to declare such variables `const`. As a result, more non-mutator functions that accepted `char *` have been updated to accept `const char *` instead.

In particular, the roles of `gbl.src_file` (`const` field representing the full path to an input file) and the local variable `sourcefile` (non-`const`, `malloc`-ed/`free`-d variable representing the base file name only) are now clearly separated.


tools/flang1/flang1exe/main.c
```diff
@@ -622,41 +607,41 @@ static void
...
+          {".hpf", 0, 0}, {".f", 0, 0},   {".F", 0, 1},
+          {".f90", 1, 0}, {".F90", 1, 1}, {".f95", 1, 0},
+          {".F95", 1, 1}, {".for", 0, 0}, {".fpp", 0, 1},
+          {0, 0, 0},
   };
   char *followval;
   int followindex;
   time_t now;
 
+
```
pawosm-arm, line 644: No need for one more extra line

tools/flang1/flang1exe/main.c
```diff
@@ -78,41 +78,26 @@ static int debugfunconly = -1;
...
-                      "schedule", "assemble", "xref",   "unroll"};
+static const char *who[] = {"init",      "parser",   "bblock",
+                            "vectorize", "optimize", "schedule",
+                            "assemble",  "xref",     "unroll"};
 #define _N_WHO (sizeof(who) / sizeof(char *))
 static INT xtimes[_N_WHO];
 static LOGICAL postprocessing = TRUE;
 
 /* Feature names for Fortran front-end */
 #if defined(TARGET_LINUX_X8664)
```
pawosm-arm, line 89: I do realize that this is not the objective of this patch, but IMO this place really needs detachment of Linux platform from x86_64 architecture.

pawosm-arm (approved):
In general it's ok, there are some comments of minor importance.

tools/flang1/flang1exe/main.c
bryanpkc: I will create another PR for this concern.

tools/flang1/flang1exe/main.c
bryanpkc: Fixed.

pawosm-arm (approved):
LGTM

kiranchandramohan:
ping @shivaramaarao

shivaramaarao (approved):

pawosm-arm:
After applying this patch on our fork of the compiler, the `intrinsic_nearest.f90` test case after compiled prints `1` instead of `PASS`. I wonder how this patch could possibly cause that...

PR #1124, opened 2021-08-22
[flang1][flang2] Use const char * in file generation tools; NFCI

bryanpkc:
Generated files such as errmsgdf.h, machardf.h, proddf.h, ilinfodf.h, ilmtpdf.h, etc. initialize `char *` fields with string literals, leading to `-Wwritable-strings` and `-Wcast-qual` warnings. This patch updates the struct declarations to use `const char *` fields, and adapts many flang1/flang2 APIs and constants similarly, eliminating a large number of such warnings.


pawosm-arm (approved):
so good!

bryanpkc:
ping @shivaramaarao

shivaramaarao (approved):

PR #1130, opened 2021-08-31
[CCFF] Use const char * in CCFF APIs; NFCI

bryanpkc:
The string parameters in the common compiler feedback format (CCFF) APIs are never modified, and many of them are called with string literals as arguments. Declaring the parameters `const char *` eliminates a number of `-Wcast-qual` and `-Wwritable-strings` warnings.

### DIFF
--- a/tools/flang2/flang2exe/main.cpp
+++ b/tools/flang2/flang2exe/main.cpp
@@ -18,6 +18,7 @@
 #include "machar.h"
 #include "upper.h"
 #include "fih.h"
+#include "ccffinfo.h"
 #include "cgraph.h"
 #include "x86.h"
 #include "dbg_out.h"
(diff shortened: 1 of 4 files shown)


kiranchandramohan:
Is the CCFF API actually used in the classic flang compiler?

bryanpkc:
> Is the CCFF API actually used in the classic flang compiler?

I am not sure how to use CCFF output, but the APIs are called quite a bit in `comm.c`, `inliner.c`, `expsmp.cpp`, and `ompaccel.cpp`, apparently to create a hierarchical data structure that traces the AST/ILM/ILI transformations. I think NVIDIA/PGI's downstream compilers might still use them? If everyone agrees (especially @mleair on behalf of NVIDIA/PGI), I can certainly submit a PR to rip out CCFF completely.

pawosm-arm (approved):
very useful change.

bryanpkc:
@shivaramaarao Can I get your approval on this? I'd like to merge this now as it blocks some other warning fixes that I haven't yet posted PRs for. I think we can consider removing the CCFF API in the future.

PR #1133, opened 2021-09-09
Removing CI build workflow for release_100 from classic-llvm-flang

abrahamtovarmob:
CI workflow for release_11x and release12x still enabled, Flang workflow not broken

### DIFF
--- a/.github/workflows/build_flang.yml
+++ b/.github/workflows/build_flang.yml
@@ -21,13 +21,8 @@ jobs:
         cc: [clang]
         cpp: [clang++]
         version: [10, 11]
-        llvm_branch: [release_100, release_11x, release_12x]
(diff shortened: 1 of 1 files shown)


bryanpkc (approved):

pawosm-arm (commented):
Why are there two commits in this PR? What is the role of this merge commit?

abrahamtovarmob:
> Why are there two commits in this PR? What is the role of this merge commit?

The first commit removed the "release_100" from the build workflow, but a keen observation showed that we would like to keep building, just to avoid testing against release_100, at least for now, so the second commit rollback the changes from the 1st commit and add the changes to only remove the test for 'release_100'

pawosm-arm (commented):
OK

Issue #893, opened 2020-05-15
Implied shape arrays are not supported


RichBarton-Arm:
Flang does not support implied shape arrays (F2008 feature). Example:

`program main
INTEGER, DIMENSION(*), PARAMETER :: elpa_kernel_names = [INTEGER :: 2]
end program main
`
Flang gives this error:
> $ flang main.f90 
F90-S-0084-Illegal use of symbol elpa_kernel_names - a named constant array must have constant extents (main.f90: 3)
F90-S-0000-Internal compiler error. mk_id: invalid symbol table index       0  (main.f90: 3)
F90-S-0000-Internal compiler error. sym_of_ast: unexpected ast      16  (main.f90: 3)
  0 inform,   0 warnings,   3 severes, 0 fatal for main

pawosm-arm:
This is yet another CP2K issue. @bryanpkc do you know about any fix for this perhaps?

bryanpkc:
@pawosm-arm We have implemented implied shape arrays, and are going to upstream a patch for this feature. We should have some time to do so once the warning fixes are all merged.

zhaochuanfeng:
I submitted a PR  https://github.com/flang-compiler/flang/pull/1136  to resolve this issue.

pawosm-arm:
@zhaochuanfeng and @bryanpkc thanks for fixing this. Having more of unmodified CP2K being able to build, I'd like to ask you about one more CP2K compilation glitch. While building CP2K, did you have to introduce the following change in order to build it succesfully?:
```
diff --git a/src/motion/space_groups.F b/src/motion/space_groups.F
index 0cce6d7c0..a90d1388a 100644
--- a/src/motion/space_groups.F
+++ b/src/motion/space_groups.F
@@ -410,7 +410,7 @@ CONTAINS
       INTEGER                                            :: handle, i, ia, ib, ir, j, natom, nop, &
                                                             nshell
       REAL(KIND=dp)                                      :: diff
-      REAL(KIND=dp), DIMENSION(3)                        :: rb, ri, ro, tr
+      REAL(KIND=dp), DIMENSION(3)                        :: rb, ri, ro, tr, riro
       REAL(KIND=dp), DIMENSION(3, 3)                     :: rot

       CALL timeset(routineN, handle)
@@ -427,7 +427,7 @@ CONTAINS
          spgr%eqatom(:, ia) = ia
       END DO

[...]

PR #1136, opened 2021-09-16
[feature][flang1] Support implied-shape array

zhaochuanfeng:
Support `implied-shape array` feature of F2008.

An implied-shape array is a named constant that is declared with each upper bound given as
an asterisk. It takes its shape from its constant expression2, for example,
```
integer, parameter :: order(0:*) = [0, 1, 2, 3]
```
Also fix the issue #893

### DIFF
--- a/test/f90_correct/inc/implied_shape_array_01.mk
+++ b/test/f90_correct/inc/implied_shape_array_01.mk
@@ -1,0 +1,25 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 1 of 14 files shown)


pawosm-arm (approved):
Good to see it fixed!

zhaochuanfeng:
What is the reason why the check failed on X86, please? It seems not because of my commit.

pawosm-arm:
@zhaochuanfeng it's not because of your commit, all other commits suffer from the same problem.

bryanpkc (approved):

bryanpkc:
ping @shivaramaarao

shivaramaarao (approved):

Issue #1137, opened 2021-09-16
regression: several programs using ENTRY cannot be compiled after applying PR #1036


pawosm-arm:
I've managed to derive some common pattern of those failing examples and eventually I've derived the following minimal reproducer:
```
module ent
contains
  function f(n)
    character(n) :: f, e
    print *, "hello1"
    return
  entry e(n)
    print *, "hello2"
  end function
end module
```
This module fails to compile as such:
```
Lowering Error: visit symbol e(636) which was replaced by f(635)
Lowering Error: visit symbol e(636) which was replaced by f(635)
F90-F-0000-Internal compiler error. Errors in Lowering       2  (entry.f90: 9)
```
Before PR #1036 it could be compiled.

yaoliu-hnc:
Sorry about the regression and thank you for deriving a test case. I'm working on this issue.  The root cause is related with the replace_fval_in_params () function in lowersym.c as the return type of the entry function is character. I will fix it ASAP.

pawosm-arm:
Hey @yaoliu-hnc, no worries, regressions happen, this is inherent part of project development cycle. No need to hurry, just take as much time as it needs. Thanks for tackling this.

yaoliu-hnc:
Hi @pawosm-arm, I've submitted a PR #1139 to solve this problem. Could you help to check whether this patch fixes the NAG suite test cases?

pawosm-arm:
Fixed. Thanks :+1:

PR #1036, opened 2021-04-22
Fix error of processing entry statement.

yaoliu-hnc:
Fix error of processing entry statement:

* Compiling error when return type of entry is complex/pointer/array.

* Inconsistent with Fortran 2008 standard, ISO/IEC JTC 1/SC 22/WG 5/N1830
  section-12.6.2.6 (If the characteristics of the result of the function
  named in the ENTRY statement are the same as the characteristics of the
  result of the function named in the FUNCTION statement, their result
  variables identify the same variable, although their names need not be
  the same).

PR #1138, opened 2021-09-17
Revert "Fix error of processing entry statement."

shivaramaarao:
Reverts flang-compiler/flang#1036. please refer to issue #1137

PR #1139, opened 2021-09-18
Fix for regression introduced by PR #1036 (issue #1137).

yaoliu-hnc:
When the entry returns adjustable length character without POINTER attribute,
the field NEWARG of the return variable is not set which causes that the return
variable in dummy arguments is not replaced. This commit adds the check for
the field NEWARG to fix the regression.


pawosm-arm (changes requested):
Is this the most up to date version of this patch?

kiranchandramohan:
Merged the patch which reverts the original change. So I think your first commit should be to re-revert the change and the second commit can be the fix.

https://github.com/flang-compiler/flang/pull/1138

yaoliu-hnc:
> Is this the most up to date version of this patch?

I've fixed it and am doing more tests. Once these tests are finished, I will update the patch again.

yaoliu-hnc:
@pawosm-arm @kiranchandramohan @bryanpkc @shivaramaarao The patch has been updated. Please have a look.

pawosm-arm (approved):

pawosm-arm (approved):

shivaramaarao (approved):
LGTM

tools/flang1/flang1exe/lowersym.c
```diff
@@ -1239,14 +1239,25 @@ replace_fval_in_params(SPTR entry, SPTR entrysame)
...
   fval = FVALG(entry);
   fvalsame = FVALG(entrysame);
   newdsc = NEWDSCG(fval);
   newarg = NEWARGG(fval);
+  /* If the return variable is an adjustable length character without POINTER
+   * attribute, its NEWARG keeps SPTR_NULL(see init_newargs in bblock.c), and
+   * itself will be put into dummy parameters(see newargs_for_entry in
+   * dpm_out.c), so we match and replace the return variable directly.
+   */
+  if (newarg == 0)
```
bryanpkc, line 1255: Instead of 0, I suggest using `SPTR_NULL`.

tools/flang1/flang1exe/lowersym.c
```diff
@@ -1239,14 +1239,25 @@ replace_fval_in_params(SPTR entry, SPTR entrysame)
...
+    newarg = fval;
   newdscsame = NEWDSCG(fvalsame);
   newargsame = NEWARGG(fvalsame);
-  params = DPDSCG(entry);
-  narg = PARAMCTG(entry);
+  if (newargsame == 0)
```
bryanpkc, line 1259: Ditto.

tools/flang1/flang1exe/lowersym.c
yaoliu-hnc: > Instead of 0, I suggest using `SPTR_NULL`.

Fixed.

bryanpkc (approved):
LGTM.

PR #1142, opened 2021-09-22
[DebugInfo] Missing debug info when one of the module variable is renamed

alokkrsharma:
Currently ILM file generated by Flang1 does not distinguish between
below two casese
1. Renamed use
use mymod, var_z => var_y
2. Restricted and renamed use
use mymod, ONLY: var_z => var_y
In both the cases debug info generated for case-2.

Please consider below testcase
``````````````````````` 
module mod1
  integer :: var_x = 30, var_y = 31
  real :: var_f = 30.1
end module mod1
program p1
use mod1,var_z => var_x ! rename var_x with var_z, unable to access var_y now in gdb
  integer::aa=100
  print *,var_z
end
```````````````````````
Without this patch, Flang generated binary behaves inside debugger as
```````````````````````
gdb) print var_y^M
No symbol "var_y" in current context.
```````````````````````
Now Flang1 is modified to preserve restricted use status in f62 and Flang2 preserves in f95.
Now for non-restricted case debug info is generated for other (non aliased) variables as well

### DIFF
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -13981,17 +13981,29 @@ add_debug_cmnblk_variables(LL_DebugInfo *db, SPTR sptr)
   scope = SCOPEG(sptr);
   for (var = CMEMFG(sptr); var > NOSYM; var = SYMLKG(var)) {
     if ((!SNAME(var)) || strcmp(SNAME(var), SYMNAME(var))) {
+      if (gbl.rutype != RU_BDATA && NEEDMODG(scope) &&
+          lookup_modvar_alias(var)) {
+        has_alias = true;
+        break;
+      }
+    }
+  }
+  for (var = CMEMFG(sptr); var > NOSYM; var = SYMLKG(var)) {
+    if ((!SNAME(var)) || strcmp(SNAME(var), SYMNAME(var))) {
       if (CCSYMG(sptr)) {
         debug_name = new_debug_name(SYMNAME(scope), SYMNAME(var), NULL);
       } else {
         debug_name = new_debug_name(SYMNAME(scope), SYMNAME(sptr),
                                     SYMNAME(var));
       }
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by pawosm-arm — APPROVED** [review]

**Comment by alokkrsharma** [comment]
Hi @bryanpkc , @kiranchandramohan can you please review this ?

**Comment by alokkrsharma** [comment]
Hi @bryanpkc , @kiranchandramohan , Gentle reminder !!

File: tools/flang2/flang2exe/cgmain.cpp [hunk 1]
```diff
@@ -14088,11 +14096,10 @@ add_debug_cmnblk_variables(LL_DebugInfo *db, SPTR sptr)
         debug_name = new_debug_name(SYMNAME(scope), SYMNAME(sptr),
                                     SYMNAME(var));
       }
-      if (gbl.rutype != RU_BDATA && 
-          NEEDMODG(scope) && lookup_modvar_alias(var)) {
+      if (gbl.rutype != RU_BDATA && NEEDMODG(scope) && has_alias &&
+          (!RESTRICTEDG(sptr) || lookup_modvar_alias(var))) {
```
bryanpkc, line 14100: I think the intention here would be clearer if the conditions were reorganized like this (which should be logically equivalent):
```
      if (gbl.rutype != RU_BDATA && NEEDMODG(scope) &&
          (lookup_modvar_alias(var) ||
           (has_alias && !RESTRICTEDG(sptr)))) {
```

alokkrsharma: Yes, this looks far better. I am updating it.

alokkrsharma:
@bryanpkc do you have any more comments ?

alokkrsharma:
@bryanpkc Gentle reminder !!!

alokkrsharma:
> LGTM. Thanks.

Thanks a lot !!!

PR #1143, opened 2021-09-22
[DebugInfo] Debug support for module defined in a different file

alokkrsharma:
Please consider below testcases.
use_only_test.f90_mod.f90
````
module use_only_mod
  integer :: mod_var = 10
end module use_only_mod
````
use_only_test.f90
````
program main
  use use_only_mod, only: mod_var
  implicit none
  print*, mod_var
end
````
Without this patch, executable behaves inside debugger as
````
Breakpoint 1, main () at use_only_test.f90:4
4         print*, mod_var
(gdb) p mod_var
No symbol "mod_var" in current context.
````

### DIFF
--- a/test/debug_info/use_only_test.f90_mod.f90
+++ b/test/debug_info/use_only_test.f90_mod.f90
@@ -1,0 +1,7 @@
+!! This file is compiled with other test use_only_test.f90
+!! Below run is just to satisfy test framework as run is must for 
+!! testcase to exist in this directory.
+!RUN: true
+module use_only_mod
+  integer :: mod_var = 10
+end module use_only_mod
--- a/test/debug_info/use_only_test.f90
+++ b/test/debug_info/use_only_test.f90
@@ -1,0 +1,15 @@
+!RUN: %flang -g -S -emit-llvm %s_mod.f90 %s
+!RUN: cat use_only_test.ll | FileCheck %s
+
+!CHECK: [[MOD_VAR:![0-9]+]] = distinct !DIGlobalVariable(name: "mod_var"
+!CHECK: distinct !DICompileUnit
+!CHECK-SAME: imports: [[IMPLIST:![0-9]+]]
+!CHECK: [[IMPLIST]] = !{[[IMPORT:![0-9]+]]
+!CHECK: [[IMPORT]] = !DIImportedEntity(tag: DW_TAG_imported_declaration
+!CHECK-SAME: entity: [[MOD_VAR]]
+
+program main
+  use use_only_mod, only: mod_var
+  implicit none
+  print*, mod_var
+end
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by pawosm-arm — COMMENTED** [review]
How do you enforce the dependency between those two test cases? People should still be able to run testing in parallel, e.g. `make -j check-flang`

**Comment by alokkrsharma** [comment]
> 
> 
> How do you enforce the dependency between those two test cases? People should still be able to run testing in parallel, e.g. `make -j check-flang`

That should be fine. Please consider below line in test 'use_only_test.f90'
````
!RUN: %flang -g -S -emit-llvm %s_mod.f90 %s
```` 
it compiles both the files. (use_only_test.f90 and use_only_test.f90_mod.f90). This is the file which has all the CHECK statements.
Another file 'use_only_test.f90_mod.f90' has only place holder RUN statement (without which testing framework cribs), it has no CHECK statement.

**Comment by pawosm-arm** [comment]
I'm afraid this needs to be discussed more thoroughly. The tests started in parallel will build things in parallel. So if both of those tests end up building the same .mod file simultaneously, it may result in corrupted .mod file. This almost never happens on common desktop PC's with low number of cores, but on our biggest machines with 100+ cores, it (corrupted .mod file) happens almost always and is very annoying. One of the solutions we're taking here is to build things in separate directories, but it won't help here: we have clear dependency and those two things must be built sequentially.

**Comment by alokkrsharma** [comment]
> 
> 
> I'm afraid this needs to be discussed more thoroughly. The tests started in parallel will build things in parallel. So if both of those tests end up building the same .mod file simultaneously, it may result in corrupted .mod file. This almost never happens on common desktop PC's with low number of cores, but on our biggest machines with 100+ cores, it (corrupted .mod file) happens almost always and is very annoying. One of the solutions we're taking here is to build things in separate directories, but it won't help here: we have clear dependency and those two things must be built sequentially.

what if the place holder RUN command in second file is changed to (non interfering command) ?
````
diff --git a/test/debug_info/use_only_test.f90_mod.f90 b/test/debug_info/use_only_test.f90_mod.f90
index 3839768d..e2811c8c 100644
--- a/test/debug_info/use_only_test.f90_mod.f90
+++ b/test/debug_info/use_only_test.f90_mod.f90
@@ -1,4 +1,4 @@
-!RUN: %flang -g -S -emit-llvm %s
+!RUN: ls  %s
 module use_only_mod
   integer :: mod_var = 10
 end module use_only_mod
````

**Comment by pawosm-arm** [comment]
Tricky :) if it works, it should be fine.

**Comment by pawosm-arm** [comment]
but please add a word of comment that this module gets compiled by the other test file.

**Review by pawosm-arm — APPROVED** [review]

**Comment by alokkrsharma** [comment]
Hi @bryanpkc , @kiranchandramohan can you please review this ?

File: test/debug_info/use_only_test.f90_mod.f90 [hunk 1]
```diff
@@ -0,0 +1,7 @@
+!! This file is compiled with other test use_only_test.f90
+!! Below run is just to satisfy test framework as run is must for 
+!! testcase to exist in this directory.
+!RUN: ls  %s
```
bryanpkc, line 4: The LLVM convention for this common idiom is:
```
!RUN: true
```

alokkrsharma: Thanks. I am updating this,

alokkrsharma:
@bryanpkc, any other comment on this ?

alokkrsharma:
@bryanpkc Gentle reminder !!!

alokkrsharma:
> LGTM.

Thanks a lot !!!

PR #1144, opened 2021-09-22
[runtime] Use const char * where applicable; NFCI

bryanpkc:
This PR was part of #1087, and has been extracted to ease code reviews.

This patch fixes many `-Wincompatible-pointer-types-discards-qualifiers` and `-Wcast-qual` warnings that resulted from mixing `char *` variables with `const char *` values, especially string literals. Most of these can be trivially fixed by changing the types of the variables or functions in question to `const char *`. Some cases are handled by calling `strdup()` to put a literal in a non-const variable.

A couple of places in runtime/flang/eoshift.c cast string literals and `long long int *` values into `char *`. They are left as FIXMEs for now as the results of the casts are passed to many functions via `char *` parameters, and fixing all of those functions will not be trivial.

### DIFF
--- a/runtime/flang/allo.c
+++ b/runtime/flang/allo.c
@@ -391,9 +391,8 @@ I8(__alloc04)(__NELEM_T nelem, dtype kind, size_t len,
   if (*pointer && I8(__fort_allocated)(*pointer)
       && ISPRESENT(stat) && *stat == 2) {
     int i;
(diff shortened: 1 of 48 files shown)


pawosm-arm (approved):

shivaramaarao (approved):
looks good to me.  thanks

PR #1145, opened 2021-09-22
win: Fix the build on Windows

kaadam:
The build is broken with a redefinition error:

typedef redefinition with different types
      ('__INT_T' (aka 'long') vs 'int')
typedef __INT_T INT;       /* native integer at least 32 bits */

'INT' is already defined in winnt.h as 'int'.
The comment says it needs a 'native integer at least 32 bits',
'int' should be fine for it on all platform.

### DIFF
--- a/runtime/flang/global.h
+++ b/runtime/flang/global.h
@@ -16,6 +16,7 @@
 #include "fioMacros.h"
 #include "stdioInterf.h" /* stubbed version of stdio.h */
 #include "cnfg.h" /* declarations for configuration items */
+#include <stdint.h>
 
 #define GBL_SIZE_T_FORMAT "zu"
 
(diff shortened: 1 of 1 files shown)


pawosm-arm (commented):
maybe we should use int32_t instead? what is the availability of <stdint.h> on Windows?

kaadam:
@pawosm-arm yes, int32_t is available in <stdint.h> on Windows, so we could use it here. I update the PR.

pawosm-arm (approved):
It is OK for me, and I always have preference on stdint types over local reinventions, but I'm not sure if all the building environments that we're targeting expose such header. If it was that easy, no one would ever do that any other way!
I mean: I thought you'll surround this change with some ifdef ensuring you're targeting your particular environemt, but from the other side, it would make this patch ugly and I'd rather see us leaning towards standards and do exceptions for those who don't comply not the other way round.

kaadam:
@bryanpkc Could you take a look at it please? Thanks in advance.

kaadam:
@pawosm-arm Could you merge this PR as well? Thanks.

Issue #1146, opened 2021-09-28
Get the wrong variable after module use statement


Zek-Li:
This is a runtime error of ATM of CESM, and I reduced the case like below:

```
module public1
  ! This should be the value we get after calling foo() in program use_1_2
  integer, public :: var = 1
  integer, public :: var1 = 11 ! Removing this statement fixes the issue
end

module public2
  integer, public :: var = 2
end

module use_public2
  use public2, only : var
  private ! Removing this fixes the issue
contains
end

program use_1_2
  use public1, only : var1
  use use_public2

  integer :: ret = 0

  ret = foo()
  ! ret shoule be 1
  print *, ret
contains
  function foo() result(ret)
    use public1, only : var

    integer :: ret

    ret = var
  end
end

```

Zek-Li:
Hi @michalpasztamobica, do you have any idea to solve this problem?

kiranchandramohan:
Thanks @Zek-Li for raising the issue. @michalpasztamobica has moved to work on another project. If reverting https://github.com/flang-compiler/flang/commit/56e924c5b5bd5823dd383b72e5e1416492d49d17 fixes the issue then please go ahead and revert.
BTW, is this from @bryanpkc's team?

bryanpkc:
> BTW, is this from @bryanpkc's team?

Yes.

michalpasztamobica:
Hi, sorry for low responsiveness. Indeed, I am busy doing something else at the moment.
If a fix without revert is possible, that would be better. This commit was proven to fix an issue: https://github.com/flang-compiler/flang/pull/949

xinliu-hnc:
Hi @Zek-Li , I've submitted a PR #1149 to solve this problem. The above small test case passed. Could you help to check whether this patch fixes the error of ATM of CESM?

Zek-Li:
Good job! This patch passed the case of CESM. 
I'm still confused at [this line](https://github.com/flang-compiler/flang/blob/master/tools/flang1/flang1exe/module.c#L323). I think `global = SST_SYMG(RHS(3));` may not be the symbol we want, do we need to create the alias? In fact, everything works fine if I delete this line.

pawosm-arm:
Shouldn't it be closed since #1146 was merged?

kiranchandramohan:
@pawosm-arm if the issue is fixed then please close.

pawosm-arm:
[...]

PR #1149, opened 2021-10-09
Fix for the issue #1146 introduced by PR #949(issue #889).

xinliu-hnc:
The reason of the problem is that the symbols with the same name
in different modules are linked in the hash list, and the local symbol is
inserted in front of the matching symbol instead of the head of hash list.
This commit adds an existence check of rename to fix the problem.


pawosm-arm (approved):

bryanpkc (approved):
LGTM.

PR #1151, opened 2021-10-12
Remove Fortran entry symbol formats for Windows

kaadam:
We can use standard symbol names, it seems it is not
necessary to keep different symbol names for Windows.


pawosm-arm (approved):

kiranchandramohan:
ping @bryanpkc

PR #1161, opened 2021-11-01
[test] Stop writing to source directory; NFC

bryanpkc:
The test case `test/sema/aconst_with_null_subict.f90` was writing a `.qdbf` file to the source directory unnecessarily. This patch stops it from doing so.


pawosm-arm (approved):

kiranchandramohan (approved):
Lgtm.

PR #1170, opened 2021-11-08
Create retained node list for function arguments (fixes #1159)

### LINKED ISSUE #1159: Flang does not create retained node list for function arguments
LLVM13 has become very strict for a function arguments to be able to appear in debuginfo. As a result, if given function argument is not used in the function code, it now disappears from debugging data. The only way for preventing that is to do it the clang way. All of the arguments should be listed in the 'retained nodes list'.

Consider following Fortran code:
```
subroutine addf(a, b, c, d, e, f, g, h, i)
  integer :: a, b, c, d, e, f, g, h, i
end subroutine
```
In order for those variables to appear in the debuginfo, I had to manually add 'retained nodes list' to the IR as in the following snippet (see entry !28 which is referred from entry !15, the entry !28 is the thing that flang does not emit!):
```
!15 = distinct !DISubprogram(file: !3, scope: !10, name: "addf", line: 1, type: !13, spFlags: 8, unit: !10, scopeLine: 1, retainedNodes: !28)
!16 = !DILocation(scope: !15)
!17 = !DILocalVariable(scope: !15, name: "a", arg: 1, file: !3, line: 1, type: !11)
!18 = !DIExpression()
!19 = !DILocalVariable(scope: !15, name: "b", arg: 2, file: !3, line: 1, type: !11)
!20 = !DILocalVariable(scope: !15, name: "c", arg: 3, file: !3, line: 1, type: !11)
!21 = !DILocalVariable(scope: !15, name: "d", arg: 4, file: !3, line: 1, type: !11)
!22 = !DILocalVariable(scope: !15, name: "e", arg: 5, file: !3, line: 1, type: !11)
!23 = !DILocalVariable(scope: !15, name: "f", arg: 6, file: !3, line: 1, type: !11)
!24 = !DILocalVariable(scope: !15, name: "g", arg: 7, file: !3, line: 1, type: !11)
!25 = !DILocalVariable(scope: !15, name: "h", arg: 8, file: !3, line: 1, type: !11)
!26 = !DILocalVariable(scope: !15, name: "i", arg: 9, file: !3, line: 1, type: !11)
!27 = !DILocation(line: 3, column: 1, scope: !15)
!28 = !{!17, !19, !20, !21, !22, !23, !24, !25, !26}
```

pawosm-arm:
This commit adds the creation of the retained node list for function
arguments. The retained node list is required by LLVM13 in order
to have debuginfo metadata being created for function arguments
also in a situation where they are not used within the function code.


tools/flang2/flang2exe/lldebug.cpp
```diff
@@ -507,7 +507,8 @@ lldbg_create_subprogram_mdnode(
           ll_get_md_node(db->module, LL_PlainMDNode, &lv_list_mdnode, 1));
     else
       llmd_add_md(mdb, lv_list_mdnode);
```
bryanpkc, line 509: Apparently, this was already adding `lv_list_mdnode` for LLVM versions between 3.4 and 7.0, but in the corresponding `DISubprogram` templates (in `ll_write.cpp`), the name of the field was `variables` instead of `retainedNodes`.

pawosm-arm: Agree. Note that what was in the past stays in the past, I don't want to complicate the logic of this code by touching paths that deal with the backward compatibility.

bryanpkc (approved):
I haven't built and tested this patch but it looks straightforward to me.

tools/flang2/flang2exe/lldebug.cpp
bryanpkc: No objection.

pawosm-arm:
I've forgotten to `git add` the test case file :D

pawosm-arm:
...so it won't pass before LLVM13 merge is completed

kiranchandramohan:
> ...so it won't pass before LLVM13 merge is completed

Since we test LLVM 11 and 12 also, would that mean this test will fail there? Do we have to set up a way to check with llvm version?

pawosm-arm:
I have no idea. Is there a way to require particular LLVM version for a given test case?

pawosm-arm:
I have a solution for this. Will push it later.

pawosm-arm:
@shivaramaarao could you look at this too, please?

bryanpkc (approved):

pawosm-arm:
@alokkrsharma can I ask you to look at this, please?

alokkrsharma:
> 
> 
> @alokkrsharma can I ask you to look at this, please?

I plan to complete this in a day or two.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1479,6 +1479,7 @@ process_params(void)
       }
     }
   }
+  lldbg_emit_lv_list(di);
```
alokkrsharma, line 1482: Current function processes constants (variables with parameter attribute). The call to lldbg_emit_lv_list should be at better place may be at the end of function process_formal_arguments. That place makes sure db->param_stack[] is populated.
It will also be good to make the call lldbg_emit_lv_list under the condition (db->routine_idx > 0) to make sure we have DISubprogram ready.

alokkrsharma (commented):
Apart from one comment, rest LGTM.

tools/flang2/flang2exe/cgmain.cpp
pawosm-arm: Your comment leads to a significant change. But I like that change, I'll upload it later, when it passes tests.

pawosm-arm:
@alokkrsharma can I ask you to complete the review?

test/lit.cfg
```diff
@@ -454,9 +454,13 @@ def get_llvm_config_props():
 
 config.available_features.update(get_llvm_config_props())
 
+
```
alokkrsharma, line 457: unintended blank line ?

alokkrsharma (approved):
LGTM except one nit.

test/lit.cfg
pawosm-arm: indeed

PR #1171, opened 2021-11-12
[flang2] Delete dead loop metadata code and fix build error on master branch; NFCI

bryanpkc:
Commits 4c6964dabf06, 2838f5dbdc58, and 255fce7a3bf5 have re-written how loop metadata are generated for `NODEPCHK` and `VECTOR ALWAYS` directives. As a result, the `mark_rw_nodepchk` function from the original implementation is no longer called, which means the `rw_nodepchk` flag is never true, causing some code to never run. In particular, the unused `mark_rw_nodepchk` function causes `-Werror` to fail the build. This patch deletes all the dead code, merges `cons_no_depchk_metadata` into `cons_vec_always_metadata` (both are identical) and fixes up some incorrect comments.

### DIFF
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -183,7 +183,6 @@ static struct {
   unsigned _float_jmp : 1;
   unsigned _fcmp_negate : 1;
   unsigned _last_stmt_is_branch : 1;
-  unsigned _rw_no_dep_check : 1;
   unsigned _rw_acc_grp_check : 1;
 } CGMain;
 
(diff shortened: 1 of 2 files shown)


pawosm-arm (approved):
Always nice to review commits that remover more than they add :)

bryanpkc:
@kiranchandramohan @shivaramaarao Please review and merge this PR at your earliest convenience. CI is currently failing on the master branch.

shivaramaarao (approved):

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1701,10 +1669,8 @@ schedule(void)
     if ((!XBIT(69, 0x100000)) && BIH_NODEPCHK(bih) &&
         (!ignore_simd_block(bih)) || XBIT(191, 0x4)) {
       fix_nodepchk_flag(bih);
```
kiranchandramohan, line 1671: Is this used now?

kiranchandramohan (approved):
LGTM.

tools/flang2/flang2exe/cgmain.cpp
bryanpkc: Yes, it is. The function `fix_nodepchk_flag` could use a better name, but the new `llvm.access.group` metadata mechanism still depends on the `BIH_NODEPCHK2` flag set by the function.

PR #1173, opened 2021-11-15
[flang2] Make flang aware of LLVM Target features

RichBarton-Arm:
Add a -target_features option to flang2 that takes a comma-separated list of
LLVM target features suitable for embedding in LLVM metadata.

Teach flang2 to output this as a "target-features" metadata.

This is needed for flang to produce llvm objects that can be used with LTO and
a combination of targets, for example Arm SVE targets.

Depends on https://github.com/flang-compiler/classic-flang-llvm-project/pull/66 on the classic-flang-llvm-project to tell the driver to pass through the options.


pawosm-arm:
Covered by #1186 and #1187

PR #1174, opened 2021-11-17
Fix uninitialized variable-related warnings; NFCI

bryanpkc:
Fix `-Wuninitialized` and `-Wsometimes-uninitialized` warnings by initializing each affected variable (mostly to 0, NULL, or the type-appropriate equivalent). If possible, an `assert` call is added on the code path that would have resulted in the illegal use of an uninitialized variable. Also update CMakeLists.txt to turn on `-Werror` warning-to-error conversion for these types of warnings.


runtime/flang/scatter.c
```diff
@@ -533,10 +533,10 @@ sked *I8(__fort_gathscat)(gathscat_parm *z)
   DECL_HDR_PTRS(rd);
   DECL_HDR_PTRS(xd);
   DECL_F90_DIM_PTR(mdd);
-  DECL_DIST_DIM_PTR(mdd);
+  DECL_DIST_DIM_PTR(mdd) = NULL;
```
pawosm-arm, line 536: It is OK no matter how awkward this assignment looks like :)

pawosm-arm (approved):

bryanpkc:
ping @shivaramaarao

shivaramaarao (approved):
Looks good.. thanks

PR #1152, opened 2021-10-12
Flang test port for LLVM release_13x

abrahamtovarmob:
Minor changes to flang test in order to adapt them for LLVM release_13x.
Depends on https://github.com/flang-compiler/classic-flang-llvm-project/pull/62

PR #1175, opened 2021-11-19
Adding changes to tests files to adapt to LLVM13 rules.

pawosm-arm:
This is re-upload of PR #1152 which introduces two version of the prefetch test case. I decided to introduce separate version for LLVM13 (and later!) instead of turning a good test case into worse for a sake of backward compatibility.

### DIFF
--- a/test/directives/vector.f90
+++ b/test/directives/vector.f90
@@ -17,5 +17,4 @@ end subroutine
 ! IGNORE-DIRECTIVES-NOT: !"llvm.loop.vectorize.enable", i1 true
 ! CHECK: load <[[VF:[0-9]+]] x i32>
 ! CHECK: store <[[VF]] x i32>
-! VECTOR-NOT: load <[[VF:[0-9]+]] x i32>
-! VECTOR-NOT: store <[[VF]] x i32>
+! VECTOR-NOT: <{{[0-9]+}} x
--- a/test/directives/novector.f90
+++ b/test/directives/novector.f90
@@ -19,6 +19,4 @@ end subroutine
 ! CHECK: load <[[VF:[0-9]+]] x i32>
 ! CHECK: sub {{.*}} <[[VF]] x i32>
 ! CHECK: store <[[VF]] x i32>
-! NOVECTOR-NOT: load <[[VF:[0-9]+]] x i32>
-! NOVECTOR-NOT: sub {{.*}} <[[VF]] x i32>
-! NOVECTOR-NOT: store <[[VF]] x i32>
+! NOVECTOR-NOT: <{{[0-9]+}} x {{.+}}>
--- a/test/directives/omp_simd_clause.f90
+++ b/test/directives/omp_simd_clause.f90
@@ -13,9 +13,7 @@ subroutine sum(myarr1,myarr2,ub)
   end do
 end subroutine
 
-! CHECK-NOT:  {{.*}} add nsw <[[VF:[0-9]+]] x i32>{{.*}}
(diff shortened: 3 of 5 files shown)


pawosm-arm:
gentle ping...

kiranchandramohan:
@shivaramaarao @bryanpkc we need to merge this to unblock the llvm13 migration patches. Could you prioritise this? Ideally this week. So that we can merge llvm13 next week.

bryanpkc:
I was building and testing #1152 over the weekend, unaware that this had been raised instead. 😢

test/directives/prefetch-llvm_13.f90
```diff
@@ -0,0 +1,25 @@
+! REQUIRES: llvm-13
+
+! RUN: %flang -O1 -S -emit-llvm %s -o - | FileCheck %s --check-prefix=CHECK-PREFETCH
+! RUN: %flang -O1 -S -emit-llvm -Hy,59,0x4 %s -o - | FileCheck %s --check-prefix=CHECK-NOPREFETCH
```
bryanpkc, line 4: I still don't understand why we cannot have just one test, by making the `RUN` commands use `-O0`. I have verified that, at `-O0`, both LLVM 12 and LLVM 13 will produce the exact same IR, including both `llvm.prefetch` intrinsic calls as expected. The only change you need is matching a signed extension and multiplication of the 32-bit loop induction variable, instead of a left-shift of the 64-bit version of the same variable. For example:
```
! CHECK-PREFETCH: [[i:%[0-9]+]] = load i32
! CHECK-PREFETCH: [[TMP1:%[0-9]+]] = sext i32 [[i]] to i64
! CHECK-PREFETCH: [[TMP2:%[0-9]+]] = mul nsw i64 [[TMP1]], 4
! CHECK-PREFETCH: [[a2elem:%[0-9]+]] = getelementptr i8, i8* [[a2base]], i64 [[TMP2]]
! CHECK-PREFETCH: call void @llvm.prefetch{{.*}}(i8* [[a2elem]], i32 0, i32 3, i32 1)
```

test/directives/vector.f90
```diff
@@ -17,5 +17,5 @@ subroutine add(arr1,arr2,arr3,N)
 ! IGNORE-DIRECTIVES-NOT: !"llvm.loop.vectorize.enable", i1 true
 ! CHECK: load <[[VF:[0-9]+]] x i32>
 ! CHECK: store <[[VF]] x i32>
-! VECTOR-NOT: load <[[VF:[0-9]+]] x i32>
-! VECTOR-NOT: store <[[VF]] x i32>
+! VECTOR-NOT: <{{[0-9]+}} x
+
```
bryanpkc, line 21: Delete trailing blank line.

bryanpkc (changes requested):
Please rebase on the master branch. LLVM 13's CMake rules have added the `-fno-semantic-interposition` option (all else seemingly equal), and now my Debug build fails with a new warning-turned-error:
```
/home/bryanpkc/src/flang.orig/tools/flang1/utils/prstab/prstab2.c:890:7: warning: variable 'opntyp' set but not used [-Wunused-but-set-variable]
  INT opntyp, swtype;
      ^
```

test/directives/prefetch-llvm_13.f90
pawosm-arm: let's try...

pawosm-arm:
...forgot to remove one line. Now it should pass

bryanpkc (approved):
LGTM. Thanks!

shivaramaarao (approved):
looks fine to me. thanks

PR #1179, opened 2021-11-29
interface nested into subroutine causes symbol undefined, fix #1013

### LINKED ISSUE #1013: BigDFT: surprising interactions within the code: interface nested into subroutine causes imported symbol appear undefined
This is similar to #367 

In the following example:

```
module base
  implicit none
  integer, public :: a = 10
end module

module intermediate
  use base, only: a
  implicit none
  private
end module

program x
  use base
  implicit none

  interface
    subroutine sub2
      use intermediate
      implicit none
    end subroutine sub2
  end interface

  call sub

contains

subroutine sub
  use intermediate
  implicit none
  print *, a
end subroutine

end program
```
The symbol `a` imported by `use base` in `program x` became inaccessible to `subroutine sub` after the `interface` block was added. Namely, `use intermediate` line in the `subroutine sub2` declared in this `interface` block causes a problem, removal of this particular line makes this problem go away. The `intermediate` module contains `private` clause which limits the visibility of the `a` variable defined in the `base` module, yet the `base` module is also imported directly by the `use base` line in the `program x`, hence the imported `a` variable should be visible to each of the contained subroutines. Note that the `use intermediate` line itself does not cause a problem, e.g. in the contained `subroutine sub`, the `a` variable is still visible after that line. The only context in which this suspicious interaction occurs is in the `interface` block which for some reason spills whatever happens in it outside of its context. Similar problem was somehow fixed some time ago (see issue #367), yet the code behind handling interfaces in `flang1` seems to be so messy it still causes recurring problems.

pwisniewskimobica:
It's my proposal for fix for ticket
https://trello.com/c/V7wB3kC6/30-hpcl3-1205-bigdft-surprising-interactions-within-the-code-interface-nested-into-subroutine-causes-imported-symbol-appear-undefin

related to
https://github.com/flang-compiler/flang/issues/1013

More experienced devs - please share discussion about pros and cons of such solution, because I can be wrong- this is only my idea and i can not investigate it deeper without any comments from you of such solution.
Please pay attention mostly because I am afraid that this fix will put regression- will let to successfully compile the programs which should not.

### DIFF
--- a/test/sema/interface_nested_into_subroutine.f90
+++ b/test/sema/interface_nested_into_subroutine.f90
@@ -1,0 +1,41 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+! RUN: %flang -S -emit-llvm %s -o - | FileCheck %s --check-prefix=CHECK
+!
+! CHECK-LABEL: define void @MAIN_()
+module base
+  implicit none
+  integer, public :: a = 10
+end module
+
+module intermediate
+  use base, only: a
+  implicit none
+  private
+end module
+
+program x
+  use base
+  implicit none
+
+  interface
+    subroutine sub2
+      use intermediate
+      implicit none
+    end subroutine sub2
+  end interface
+
+  call sub
+
+contains
+
+subroutine sub
+  use intermediate
+  implicit none
+  print *, a
+end subroutine
+
+end program
--- a/tools/flang1/flang1exe/interf.c
+++ b/tools/flang1/flang1exe/interf.c
@@ -515,6 +515,21 @@ init_use_tree(void)
   use_tree = use_tree_end = 0;
 }
 
+void remove_from_use_tree(char *module)
+{
+  USES_LIST *prev = use_tree;
+  for (USES_LIST *n = use_tree; n; n = n->next) {
+    if (strcmp(n->use_module->modulename, module) == 0) {
+      if (n == use_tree) {
+        use_tree = use_tree->next;
+      } else {
+        prev->next = n->next;
+      }
+    }
+    prev = n;
+  }
+}
+
 static TOBE_IMPORTED_LIST *
 already_to_be_used(char *modulename, int public, int except)
 {
--- a/tools/flang1/flang1exe/module.c
+++ b/tools/flang1/flang1exe/module.c
@@ -67,6 +67,7 @@ typedef struct {
   LOGICAL submodule;    /* use of module by submodule */
   RENAME *rename;
   char *fullname; /* full path name of the module file */
+  unsigned int scope; /* scope of the module */
 } USED;
 
 struct {
@@ -371,7 +372,11 @@ apply_use_stmts(void)
   save_lineno = gbl.lineno;
 
   if (!gbl.currmod && gbl.internal <= 1) {
-    init_use_tree();
+    for (m_id = FIRST_USER_MODULE; m_id < usedb.avl; m_id++) {
+      if (usedb.base[m_id].scope > sem.scope_level) {
+        remove_from_use_tree(SYMNAME(usedb.base[m_id].module));
+      }
+    }
   }
   if (usedb.base[ISO_C_MOD].module) {
     if (usedb.base[ISO_C_MOD].module == ancestor_mod)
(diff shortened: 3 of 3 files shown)


michalpasztamobica:
Nice work @pwisniewskimobica . I was trying to approach the problem on the `usedb` level, but got completely stuck there. It's good that you looked at it from a fresh perspective. As you told me the basic `make check-all` run is passing, so I think this has a chance to work. It addresses the root cause of the problem, that might have caused many other issues - the `use` clause processing did not keep track of the scope level. Now it does.

@bryanpkc , @kiranchandramohan , your feedback and running some internal tests would let us find any issues with this solution. I see nothing obviously wrong with it and a lot of improvement it can potentially lead to.

tools/flang1/flang1exe/module.c
```diff
@@ -372,6 +373,11 @@ apply_use_stmts(void)
 
   if (!gbl.currmod && gbl.internal <= 1) {
     init_use_tree();
+    for (m_id = FIRST_USER_MODULE; m_id < usedb.avl; m_id++) {
+      if(usedb.base[m_id].scope > sem.scope_level){
+        remove_from_use_tree(SYMNAME(usedb.base[m_id].module));
```
liuxiao0909c, line 378: Since `init_use_tree` set `use_tree` to be NULL, `remove_from_use_tree` seems to cause no side effects.

kiranchandramohan:
I think the following two are internal issue numbers or trackers. Please avoid using these and instead provide a suitable title and description for the issue that you are fixing.
-> bigdft_hpcl3-1205
-> https://trello.com/c/V7wB3kC6/30-hpcl3-1205-bigdft-surprising-interactions-within-the-code-interface-nested-into-subroutine-causes-imported-symbol-appear-undefin

tools/flang1/flang1exe/interf.c
```diff
@@ -515,6 +515,22 @@ init_use_tree(void)
   use_tree = use_tree_end = 0;
 }
 
+void remove_from_use_tree(char* module){
+  USES_LIST* prev = use_tree;
+  for (USES_LIST* n = use_tree; n; n = n->next) {
+
+    if(strcmp(n->use_module->modulename,module)==0){
+        if(n == use_tree){
```
kiranchandramohan, line 523: Please use alignment that is consistent with the rest of the file.

kiranchandramohan (commented):
Please add a test case.

pawosm-arm:
Nice fix.
I can't see a commit message, only this: `bigdft_hpcl3-1205-symbol_undefined_problem,`, can you add one or more line of something human readable?
Also, the test case file seems to be missing.

tools/flang1/flang1exe/module.c
pwisniewskimobica: you are right- its fixed- i removed init_use_tree

pawosm-arm (approved):
Any mergers of this PR, please hit "Squash and merge" button, thank you :)

tools/flang1/flang1exe/interf.c
pwisniewskimobica: fixed

test/sema/interface_nested_into_subroutine.f90
```diff
@@ -39,3 +39,4 @@ subroutine sub
 end subroutine
 
 end program
+
```
pawosm-arm, line 42: newline character (mandatory) at the end of file is not the same thing as empty line at the end of file (not recommended).

bryanpkc:
> Any mergers of this PR, please hit "Squash and merge" button, thank you :)

I think @pwisniewskimobica should squash the commits himself and provide a better commit message.

tools/flang1/flang1exe/module.c
```diff
@@ -371,7 +372,11 @@ apply_use_stmts(void)
   save_lineno = gbl.lineno;
 
   if (!gbl.currmod && gbl.internal <= 1) {
-    init_use_tree();
+    for (m_id = FIRST_USER_MODULE; m_id < usedb.avl; m_id++) {
+      if(usedb.base[m_id].scope > sem.scope_level){
```
bryanpkc, line 376: Please insert spaces:
```c
      if (usedb.base[m_id].scope > sem.scope_level) {
```

pawosm-arm:
gentle ping: this seems complete now, can I ask the reviewers to look at it again?

pwisniewskimobica:
@shivaramaarao  please look at the task

shivaramaarao (approved):
Looks fine to me. thanks

pawosm-arm (commented):
void remove_from_use_tree(char *module) should be declared in a header file included by module.c: remove_from_use_tree() is defined in interf.c while it is called from module.c. After recent changes this causes a compilation error. @pwisniewskimobica please issue a fix promptly.

pwisniewskimobica:
I created new pull request with fix for CI job including changes from this PR :
https://github.com/flang-compiler/flang/pull/1206

PR #1180, opened 2021-11-29
Fix implicit function declaration warnings; NFCI

bryanpkc:
Various functions were used without proper declaration, or without including the headers that declare them. This PR adds back the missing includes and declarations to avoid `-Wimplicit-function-declaration` warnings, and also updates the compiler flags to stop exempting these warnings from `-Werror`.


bryanpkc:
To reviewers: I added comments to the `#include` statements, in order to make it obvious why a header needs to be included. They are a bit unsightly and I don't really want to create a precedent. If you agree then I will amend the patch and remove the comments. Thanks.

pawosm-arm (approved):
Fortunately, not too big this one

pawosm-arm:
> To reviewers: I added comments to the `#include` statements, in order to make it obvious why a header needs to be included. They are a bit unsightly and I don't really want to create a precedent. If you agree then I will amend the patch and remove the comments. Thanks.

I've seen these many times and I've been also adding comments justifying inclusion of headers many times in the past, so I'm not complaining.

shivaramaarao (approved):
lgtm.

PR #1185, opened 2021-12-01
Add a platform independent build script for Flang

kaadam:
This commit aims to replace shell script with a python scripts to build
Flang easier on Windows as well.

* Also add 'pyc' to .gitignore.

### DIFF
--- a/scripts/build_flang.py
+++ b/scripts/build_flang.py
@@ -1,0 +1,179 @@
+#!/usr/bin/env python
+
+import argparse
+import multiprocessing
+import platform
+import shutil
+import subprocess
+import sys
+from pathlib import Path, PurePath
+
+import settings
+
+
+def default_toolchain():
+    system = platform.uname()[0]
+    machine = platform.uname()[4]
+    toolchain = Path(settings.THIS_DIR,
+                             'cmake',
+                             f'toolchain_{system.lower()}_{machine.lower()}.cmake')
+    return toolchain if toolchain.is_file else None
+
+
+def get_arguments():
+    parser = argparse.ArgumentParser(
+        description='Help configure and build Flang')
+
+    buildopt = parser.add_argument_group('general build options')
+    buildopt.add_argument('-d', '--builddir', metavar='DIR', default='build',
+                          help='Specify build directory (default: "projects_root_dir"/%(default)s) \
+                          Note: The name of the build directory will be the same for flang and libpgmath')
+    buildopt.add_argument('--clean', action='store_true', default=False,
+                          help='Clean build')
+    buildopt.add_argument('-b', '--build-type', metavar='TYPE', choices=['Release', 'Debug', "RelWithDebInfo", "MinSizeRel"],
+                          default='Release', help='set build type (default: %(choices)s)')
+    buildopt.add_argument('--cmake-param', metavar='OPT', action='append', default=[],
+                          help='Add custom argument to CMake')
+    buildopt.add_argument('-p', '--install-prefix', metavar='PATH', nargs='?', default=None, const=False,
+                          help='Install after build, also specify LLVM dir')
+    buildopt.add_argument('--toolchain', metavar='FILE', default=default_toolchain().as_posix(),
+                          help='Specify toolchain file (default: %(default)s)')
+    buildopt.add_argument('-t', '--target', metavar='ARCH', choices=['X86', 'AArch64', 'PowerPC'], default='X86',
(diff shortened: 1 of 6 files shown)


kaadam:
Example if you run:
`build_flang.py --install_dir=path-to-installed-llvm-dir`
it produces the same behavior as the current shell script does.
The compiler toolchain can be set via toolchain.cmake file.

The scripts requires python version >=3.5  and cmake version >=3.13.

michalpasztamobica (approved):
In general it looks OK. If there is a use case for this I see no reason not to add this. I just wonder if we should perhaps move the existing bash script into `scripts/` and adjust all the CI jobs?

kaadam:
@isuruf @kiranchandramohan @bryanpkc Do you have any comment on this?

bryanpkc:
Please add the option to use CCache.

kaadam:
> Please add the option to use CCache.

Done, thanks.

kiranchandramohan:
@isuruf could you give this a try in windows?

ping @bryanpkc

scripts/cmake/toolchain_linux_x86_64.cmake
```diff
@@ -0,0 +1,5 @@
+set(dir /usr/local)
+
+set(CMAKE_C_COMPILER ${dir}/bin/clang)
+set(CMAKE_CXX_COMPILER ${dir}/bin/clang++)
+
```
bryanpkc, line 5: Delete trailing blank line.

bryanpkc:
@kaadam gentle ping

scripts/cmake/toolchain_linux_x86_64.cmake
```diff
@@ -0,0 +1,5 @@
+set(dir /usr/local)
```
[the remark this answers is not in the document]
kaadam: That's a mistake. I remove that line. Update PR soon.

kaadam:
@bryanpkc Updated the PR, could you look at it again? Thanks!

bryanpkc (approved):
LGTM. I think this can be merged to make builds easier for Windows users, but we are not dropping the original shell scripts at this time.

scripts/build_flang.py
```diff
@@ -0,0 +1,179 @@
...
+    parser = argparse.ArgumentParser(
+        description='Help configure and build Flang')
+
+    buildopt = parser.add_argument_group('general build options')
+    buildopt.add_argument('--builddir', metavar='DIR', default='build',
+                          help='specify build directory (default: "projects_root_dir"/%(default)s) \
+                          Note: The name of the build directory will be the same for flang and libpgmath')
+    buildopt.add_argument('--clean', action='store_true', default=False,
+                          help='clean build')
+    buildopt.add_argument('--build-type', metavar='TYPE', default='Release',
```
bryanpkc, line 34: I think the options should support a single-letter short form, similar to the ones in `build-flang.sh`. `--build-type`, `--install-prefix`, `--target`, and `--use-ccache` are especially important since a user will probably be typing them quite frequently.

bryanpkc: The options implemented in this script should be synchronized with the options in flang-compiler/classic-flang-llvm-project#134.

kaadam: > I think the options should support a single-letter short form, similar to the ones in `build-flang.sh`. `--build-type`, `--install-prefix`, `--target`, and `--use-ccache` are especially important since a user will probably be typing them quite frequently.

@bryanpkc Thanks for the review. That's good point.  I'm adding a new singel-letter form to the options.

kaadam: @bryanpkc  I've updated the build script.

- Added single-letter support for --builddir (-b), --debug (-d), --target (-t), --use-ccache (-c), -- install-prefix (-p)
- Removed superfluous  alias options such as build_type, llvm-root
- Refactor the path handling: os.path is replaced with pathlib, now the path handling is consistent.

bryanpkc: I suggest you revert the `--debug | -d` change, and keep the `--build-type | -b` option, so that a user can specify a `RelWithDebInfo` build or even `MinSizeRel` build. You can use the `--build-dir | -d` option to specify the location of the build. I'd also suggest adding a `--verbose | -v` option so that `ninja`, `make`, or `nmake` can be invoked with an appropriate option to print verbose messages.

bryanpkc (approved):
LGTM.

PR #1186, opened 2021-12-06
Adding vscale_range feature to flang

abrahamtovarmob:
Depends on https://github.com/flang-compiler/classic-flang-llvm-project/pull/68

Similar to https://github.com/flang-compiler/flang/pull/1173 but I'm adding the check for target-feature and adding the default value of vscale_range(1,16) in case the target-feature is known as sve capable, to check this there is a 'table' of different acceptable models that can be given as an argument input, such 'table' can be expanded as needed.
This is the first part of a larger task, next part is to be able to check values for vscale_range(min,max) that will be passed as command argument


pawosm-arm:
Covered by #1187

PR #1187, opened 2021-12-08
Implementing vscale_range with msve-vector-bit into Flang

abrahamtovarmob:
depends on https://github.com/flang-compiler/flang/pull/1173
depends on https://github.com/flang-compiler/flang/pull/1186

Implement msve-vector-bits in classic Flang. Implementing vscale_range with the functionallity to read from msve-vector-bits to be able to manipulate the range of the scale. for vscale_range(X/128, X/128)

Important to add the previous commits since the base of this work need those changes too.

### DIFF
--- a/test/llvm_ir_correct/attr.f90
+++ b/test/llvm_ir_correct/attr.f90
@@ -1,0 +1,13 @@
+!** Test checking attributes are set correctly
+! RUN: %flang -S -emit-llvm -march=armv8-a %s -o - | FileCheck %s -check-prefix=ATTRS1
+! RUN: %flang -S -emit-llvm -march=armv8-a+sve %s -o - | FileCheck %s -check-prefix=ATTRS2
+      program tz
+       integer :: i
+       integer ::acc(100)
+       do i=1,100
+            acc(i) = 5
+       end do
+      print *, acc(100)
+      end program
+! ATTRS1: attributes{{.*}}"target-features"="+neon"
+! ATTRS2: attributes{{.*}}"target-features"="+neon,+sve"
--- a/test/llvm_ir_correct/attr.f90
+++ b/test/llvm_ir_correct/attr.f90
@@ -1,6 +1,6 @@
 !** Test checking attributes are set correctly
-! RUN: %flang -S -emit-llvm -march=armv8-a %s -o - | FileCheck %s -check-prefix=ATTRS1
-! RUN: %flang -S -emit-llvm -march=armv8-a+sve %s -o - | FileCheck %s -check-prefix=ATTRS2
+! RUN: %flang -S -emit-llvm -target aarch64-linux-gnu -march=armv8-a %s -o - | FileCheck %s -check-prefix=ATTRS1
+! RUN: %flang -S -emit-llvm -target aarch64-linux-gnu -march=armv8-a+sve %s -o - | FileCheck %s -check-prefix=ATTRS2
       program tz
        integer :: i
        integer ::acc(100)
--- a/test/llvm_ir_correct/attr.f90
+++ b/test/llvm_ir_correct/attr.f90
@@ -1,4 +1,6 @@
 !** Test checking attributes are set correctly
+! REQUIRES: aarch64-registered-target
+
 ! RUN: %flang -S -emit-llvm -target aarch64-linux-gnu -march=armv8-a %s -o - | FileCheck %s -check-prefix=ATTRS1
 ! RUN: %flang -S -emit-llvm -target aarch64-linux-gnu -march=armv8-a+sve %s -o - | FileCheck %s -check-prefix=ATTRS2
       program tz
--- a/test/llvm_ir_correct/vscale-mbits.f90
+++ b/test/llvm_ir_correct/vscale-mbits.f90
@@ -1,0 +1,32 @@
+!** Test checking msv-vector-bits are passed correctly
+! REQUIRES: aarch64-registered-target
+
+! RUN: %flang -S -emit-llvm -target aarch64-linux-gnu -march=armv8-a+sve -msve-vector-bits=128 %s -o - | FileCheck %s -check-prefix=ATTRS1
+! RUN: %flang -S -emit-llvm -target aarch64-linux-gnu -march=armv8-a+sve -msve-vector-bits=256 %s -o - | FileCheck %s -check-prefix=ATTRS2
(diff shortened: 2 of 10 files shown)


test/llvm_ir_correct/attr.f90
```diff
@@ -0,0 +1,13 @@
...
+      program tz
+       integer :: i
+       integer ::acc(100)
+       do i=1,100
+            acc(i) = 5
+       end do
+      print *, acc(100)
+      end program
+! ATTRS1: attributes{{.*}}"target-features"="+neon"
+! ATTRS2: attributes{{.*}}"target-features"="+neon,+sve"
```
pawosm-arm: Newline missing after the last line.

pawosm-arm (changes requested):

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -13567,6 +13568,13 @@ print_function_signature(int func_sptr, const char *fn_name, LL_ABI_Info *abi,
     print_token(" \"target-features\"=\"");
     print_token(flg.target_features);
     print_token("\"");
+    int ret = extract_arch_feat(flg.target_features, mod);
+    if(ret) {
```
pawosm-arm, line 13572: `if (ret) {` it should be

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -13567,6 +13568,13 @@ print_function_signature(int func_sptr, const char *fn_name, LL_ABI_Info *abi,
...
+      ret = is_vscale_feat(mod);
+      if(ret) {
```
pawosm-arm, line 13574: same as above

test/llvm_ir_correct/vscale-mbits.f90
```diff
@@ -0,0 +1,32 @@
...
+      end program
+
+! ATTRS1: attributes #{{[0-9]+}}
+! ATTRS1-DAG:"target-features"="+neon,+sve"
+! ATTRS1-DAG:vscale_range(1,1)
+! ATTRS2: attributes #{{[0-9]+}}
+! ATTRS2-DAG:"target-features"="+neon,+sve"
+! ATTRS2-DAG:vscale_range(2,2)
+! ATTRS3: attributes #{{[0-9]+}}
+! ATTRS3-DAG:"target-features"="+neon,+sve2"
```
pawosm-arm, line 25: `-march=armv8-a+sve2` implies "+neon,+sve2,+sve", this check will not pass.

The same goes to lines below.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -13598,6 +13606,78 @@ INLINE void static add_property_struct(char *func_name, int nreductions,
...
+      //flg.min = argument1;
+      //flg.max = argument2;
+      break;
+    }
+  }
+
+  return result;
+}
+
+bool
```
pawosm-arm, line 13655: There's something wrong with this function, it spilled mod equals to `+sve2,+sve` which resulted in `is_vscale_feat()` couldn't recognize `+sve2` in simple `strcmp()`.

pawosm-arm (changes requested):
It won't work.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -13598,6 +13606,78 @@ INLINE void static add_property_struct(char *func_name, int nreductions,
...
+  if(i==len) {
+    result = false;
+    return result;
+  }
+
+  //here is the feature that will be extracted
+  for(j=0; i<len; ++j, ++i) {
+      str1[j] = arch_feat[i];
+  }
+  mod = (char*) calloc(j+1,sizeof(char));
```
pawosm-arm, line 13684: who frees that?!

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -13598,6 +13606,78 @@ INLINE void static add_property_struct(char *func_name, int nreductions,
...
+  print_token(" \"vscale-range\"=\"");
+  print_token(token);
+  print_token("\"");
+}
+
+bool
+is_vscale_feat(char *mod)
+{
+  bool result = false;
+  int len = sizeof(*CPU_VSCALE)-1;
```
pawosm-arm, line 13623: that's not how sizeof should be used! `CPU_VSCALE` is an array of pointers, so a single element of it (`*CPU_VSCALE`) is a pointer, on 64-bit machine it is 8 bytes in size, so sizeof returns always 8 here. You were lucky that `CPU_VSCALE` contains 7 elements and that `8 - 1` is also 7.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -13598,6 +13606,78 @@ INLINE void static add_property_struct(char *func_name, int nreductions,
...
+{
+  char token[60];
+  sprintf(token, "%d,%d", flg.min, flg.max);
+  print_token(" \"vscale-range\"=\"");
+  print_token(token);
+  print_token("\"");
+}
+
+bool
+is_vscale_feat(char *mod)
```
pawosm-arm, line 13620: should take `const char *mod` instead. The `mod` variable is not modified by this function.

pawosm-arm:
I need to make a final blow to this PR: flang2 should remain architecture independent, hence vscale_range computation depending on particular VLA implementations should not be computed here. It should be computed in LLVM's frontent driver and passed to flang2 as yet another command line option. Emission of vscale_range should therefore depend on presence of that option.

jdomke (commented):
Sorry to jump in, but I tested this PR and found multiple issues in tools/flang2/flang2exe/cgmain.cpp:

- str1[30] in extract_arch_feat is too small for a64fx which yields '+v8.2a,+fp-armv8,+neon,+crc,+crypto,+fullfp16,+ras,+lse,+rdm,+sve,+sha2,+aes' in arch_feat string
- extract_arch_feat extracts everything after the 1st comma
- is_vscale_feat assumes single feature token, but gets multiple comma-separed tokens, hence strcmp fails to find anything unless you are lucky and +sve is in 2nd position of original string
- provided test cases are insufficiant since they all yield a similar '+neon,+sve..' arch_feat string and hence don't find any of the errors which I ran into with realistic march/mcpu flags

pawosm-arm:
A decision was made that this feature needs to be provided by a new PR introducing different implementation shifting more of the logic to the frontend driver.

PR #1188, opened 2021-12-13
[debuginfo] solve issues with DILocation nodes not being generated for the CP2K sources

pawosm-arm:
CP2K uses fypp, a Fortran preprocessor based on Python expressions. It
can generate large code blocks spanning several functions from a
single line directive. This severely confuses flang's lldebug routines
making them unable to emit DILocation nodes (e.g. for the loops).

The consequences of this can vary, the most visible is an assertion
failure experienced while building CP2K with -g flag using
an assertion-enabled compiler.

The patch presented in this PR fixes the issue.

### DIFF
--- a/test/llvm_ir_correct/fypp-dilocation.f90
+++ b/test/llvm_ir_correct/fypp-dilocation.f90
@@ -1,0 +1,33 @@
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+! RUN: %flang -cpp -g -S -emit-llvm %s -o - | FileCheck %s
+
+# 111 "fypp-dilocation.F"
+function fypp1()
+# 111 "fypp-dilocation.F"
+  implicit none
+# 111 "fypp-dilocation.F"
+  integer :: fypp1
+# 111 "fypp-dilocation.F"
+  fypp1 = 0
+# 111 "fypp-dilocation.F"
+end function fypp1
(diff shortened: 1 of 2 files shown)


pawosm-arm:
Gentle ping. Any chances for this being reviewed?

bryanpkc (approved):
LGTM.

kiranchandramohan:
@alokkrsharma can you have a look?

bhuvanendrakumarn:
@pawosm-arm, i am trying to review this fix
>It can generate large code blocks spanning several functions from a single line directive
the fix address this limitation/issue, that is if the line numbers are same(last_line and lineno passed to lldbg_emit_line()) and subprogram nodes are different, then DILocation in this case does not get emitted, where it should be. Now this fix resolve this issue. Hope my understanding about the fix is correct, if yes, its LGTM from my/AMD side, thanks

>2810   if ((last_line != lineno) ||

pawosm-arm:
> @pawosm-arm, i am trying to review this fix
> 
> > It can generate large code blocks spanning several functions from a single line directive
> 
> the fix address this limitation/issue, that is if the line numbers are same(last_line and lineno passed to lldbg_emit_line()) and subprogram nodes are different, then DILocation in this case does not get emitted, where it should be. Now this fix resolve this issue. Hope my understanding about the fix is correct, if yes, its LGTM from my/AMD side, thanks

@bhuvanendrakumarn this is the issue this patch addresses. Also it is not merely a limitation, with assertions-enabled compiler an assertion is failing without this fix.

kiranchandramohan:
@shivaramaarao can you approve?

shivaramaarao (approved):

PR #1205, opened 2022-01-09
Revert "interface nested into subroutine causes symbol undefined, fix #1013"

### LINKED ISSUE #1013: BigDFT: surprising interactions within the code: interface nested into subroutine causes imported symbol appear undefined

kiranchandramohan:
Reverts flang-compiler/flang#1179


kiranchandramohan:
@pwisniewskimobica @pawosm-arm I reverted [this patch ](https://github.com/flang-compiler/flang/pull/1179) to keep the master branch clean since some people are trying to package classic flang now.

PR #1209, opened 2022-01-20
Fix lowering of move_alloc() subroutine

pawosm-arm:
This is not the best solution for this problem. But it's the one that does not require a major rewrite of the code in this area.

The Fortran's internal move_alloc() subroutine failed to accept more complex field selection expressions used for the subroutine parameters. This patch solves the problem by adding more eager traversal through the member selection expression.

### DIFF
--- a/tools/flang1/flang1exe/semutil2.c
+++ b/tools/flang1/flang1exe/semutil2.c
@@ -11916,6 +11916,16 @@ get_parm_ast(int parent, SPTR sptr, DTYPE dtype)
       return ast;
     }
   }
+  /* Field not found, so try again and take more recursive approach */
+  for (mem = DTY(dtype + 1); mem > NOSYM; mem = SYMLKG(mem)) {
+    /* Act only on the cases that were not considered before */
+    if ((!PARENTG(mem)) && strcmp(SYMNAME(sptr), SYMNAME(mem))) {
+      ast = mk_member(parent, mk_id(mem), dtype);
+      rslt = get_parm_ast(ast, sptr, DTYPEG(mem)); /* recurse here */
+      if (rslt)
+        return rslt;
+    }
+  }
   return 0;
 }
 
--- a/test/f90_correct/inc/move_alloc.mk
+++ b/test/f90_correct/inc/move_alloc.mk
@@ -1,0 +1,22 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+EXE=move_alloc.$(EXESUFFIX)
+
+build:  $(SRC)/move_alloc.f90
+	-$(RM) move_alloc.$(EXESUFFIX) core *.d *.mod FOR*.DAT FTN* ftn* fort.*
+	-$(RM) $(OBJ)
+	-$(CC) -c $(CFLAGS) $(SRC)/check.c -o check.$(OBJX)
+	@echo ------------------------------------ building test $@
+	$(FC) $(FFLAGS) $(LDFLAGS) $(SRC)/move_alloc.f90 check.$(OBJX) -o move_alloc.$(EXESUFFIX)
+
+run:
+	@echo ------------------------------------ executing test move_alloc
+	move_alloc.$(EXESUFFIX)
+
+verify: ;
+
+move_alloc.run: run
--- a/test/f90_correct/lit/move_alloc.sh
+++ b/test/f90_correct/lit/move_alloc.sh
@@ -1,0 +1,10 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+# Shared lit script for each tests. Run bash commands that run tests with make.
+
+# RUN: KEEP_FILES=%keep FLAGS=%flags TEST_SRC=%s MAKE_FILE_DIR=%S/.. bash %S/runmake | tee %t
+# RUN: cat %t | FileCheck %S/runmake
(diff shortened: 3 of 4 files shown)


kiranchandramohan:
The CI is fixed. Please rebase to run CI.

pawosm-arm:
> The CI is fixed. Please rebase to run CI.

Done :)

tools/flang1/flang1exe/semutil2.c
```diff
@@ -11916,6 +11916,16 @@ get_parm_ast(int parent, SPTR sptr, DTYPE dtype)
       return ast;
     }
   }
+  /* Field not found, so try again and take more recursive approach */
+  for (mem = DTY(dtype + 1); mem > NOSYM; mem = SYMLKG(mem)) {
+    /* Act only on the cases that were not considered before */
+    if ((!PARENTG(mem)) && strcmp(SYMNAME(sptr), SYMNAME(mem))) {
```
kiranchandramohan, line 11922: Could the real issue be that the `PARENTG` field is not set for some cases?

Assuming PARENT is for finding the parent type in an inheritance/extends hierarchy, this might only be required at the leafs in which case the recursion is not needed.

pawosm-arm: I don't think so. I've added this check for the `!PARENTG(mem)` just to prevent doing the recursive call to `get_param_ast()` for the second time. In this place we already know it won't find anything. Initially, there was no `!PARENTG(mem)` here and it worked the same, so I doubt `PARENTG` field is involved.

bryanpkc:
I tried the test case with the master branch, and the failure (while lowering `sub2`) is not immediately obvious:
```
Lowering Error: symbol a$sd is an inconsistent array descriptor
Lowering Error: symbol a$sd is an inconsistent array descriptor
Lowering Error: symbol a$sd is an inconsistent array descriptor
F90-F-0000-Internal compiler error. Errors in Lowering       3  (pr1209.f90: 87)
F90/x86-64 Linux Flang - 1.5 2017-05-01: compilation aborted
```
Initially I thought it might be related to the confusion between the names `a` (the member of `some`) and `a` (the local variable in `move_alloc_test`) but that is not the case since the problem persisted even after I renamed the derived type member. The problem doesn't affect `sub1`, so it's clear that the use of a multi-level derived type is key. However, I am not sure why in such a case `PARENTG(mem)` would be 0.

Could you explain the problem and the fix in a bit more detail?

pawosm-arm:
> I tried the test case with the master branch, and the failure (while lowering `sub2`) is not immediately obvious:
> 
> ```
> Lowering Error: symbol a$sd is an inconsistent array descriptor
> Lowering Error: symbol a$sd is an inconsistent array descriptor
> Lowering Error: symbol a$sd is an inconsistent array descriptor
> F90-F-0000-Internal compiler error. Errors in Lowering       3  (pr1209.f90: 87)
> F90/x86-64 Linux Flang - 1.5 2017-05-01: compilation aborted
> ```
> 
> Initially I thought it might be related to the confusion between the names `a` (the member of `some`) and `a` (the local variable in `move_alloc_test`) but that is not the case since the problem persisted even after I renamed the derived type member. The problem doesn't affect `sub1`, so it's clear that the use of a multi-level derived type is key. However, I am not sure why in such a case `PARENTG(mem)` would be 0.
> 
> Could you explain the problem and the fix in a bit more detail?

I'm trying to get some idea how to explain the problem better (I mean, better than I did in this PR initial comment). When lowering the built-in move_alloc() subroutine, flang1 may run into trouble when a parameter value is of derived type with several levels of nesting. And this type nesting seems to be the key issue here: it requires recursive AST traversal, more eager than flang1 attempts. The effect of this sloppy traversal is that the field it finds is not an array descriptor (as it stopped on a wrong level), hence the error message `an inconsistent array descriptor`. My fix forces (quite brutally I presume) more eager traversal.

tools/flang1/flang1exe/semutil2.c
kiranchandramohan: Reposting a message shared internally here.

I was asking because the m5.f90 program work without this patch while m6.f90 needs this patch. m5.f90 and m6.f90 given below.

It will be helpful to proceed with the review if an explanation is provided for the above as well as a testcase demonstrating the change in the AST.
```
~/classic_flang$ cat m5.f90
program move_alloc_test

  implicit none
  type bar
    integer, allocatable :: x
  end type
  type some
    type(bar), allocatable :: a(:)
  end type
  type some_more
    type(some) :: b
  end type
  type some_more_more
    type(some_more) :: c
  end type
  type(bar), allocatable :: a(:)
  type(some_more_more) :: e
  allocate(a(128))
  call move_alloc(e%c%b%a,a)
end program
```
```
~/classic_flang$ cat m6.f90
program move_alloc_test

  implicit none
  type bar
    integer, allocatable :: x
  end type
  type some
    integer, allocatable :: a(:)
  end type
  type some_more
    type(some) :: b
  end type
  type some_more_more
    type(some_more) :: c
  end type
  integer, allocatable :: b(:)
  type(some_more_more) :: e
  allocate(b(128))
  call move_alloc(e%c%b%a, b)
end program
```

pawosm-arm: @kiranchandramohan those two examples execute completely different execution paths, and I suspect the m5.f90 example exposes a different bug that is not relevant to this pull request. The code corrected by this patch is never reached for m5.f90 as the allocatable array of a derived type seems to be wrongly treated.

Let's start with a simple trace: the patch in this PR corrects the `get_parm_ast()` function. In case of m6.f90 the trace is following:

transform_move_alloc() calls copy_surrogate_to_bnds_vars() which calls add_parent_to_bounds() which eventually calls get_parm_ast().

In order for transform_move_alloc() to call copy_surrogate_to_bnds_vars(), the following condition must be held:
```
if (A_SHAPEG(fptr2) && sptr != sptr2 && !SDSCG(sptr2)) {
...
}
```
...where `sptr` represents the first param to `move_alloc()`, while `sptr2` represents the second param to `move_alloc()`.
`A_SHAPEG(fptr2)` is true in case of both m5.f90 and m6.f90, `sptr != sptr2` is true by nature, the only issue is with `SDSCG(sptr2)`. This macro checks if there is a static descriptor available for the second parameter. In case of the allocatable arrays, this should never be available and `!SDSCG(sptr2)` should yield true for both m5.f90 and m6.f90. Thing is, it doesn't. For some reason, static descriptor seems to be available for the allocatable array of a derived type! This simplifies the execution paths since the buggy function that this PR tries to fix is never being executed.

bryanpkc (approved):
This tested okay on our side.

PR #1214, opened 2022-01-24
Introduce vscale_range attribute (with 'behind the scenes' support for -msve-vector-bits) into flang

pawosm-arm:
This is a new reimplementation of the PR #1187

Corresponding change to the classic-flang-llvm-project needs to be applied.


kiranchandramohan:
The CI is fixed. Please rebase to run CI.

pawosm-arm:
> The CI is fixed. Please rebase to run CI.

Done. Still it will fail as it requires corresponding PR to be merged on the LLVM side.

pawosm-arm:
This is ready to be reviewed. The failure of AArch64 builds is caused by classic-flang-llvm-project part not being merged yet.

pawosm-arm:
As the relevant part was merged into classic-llvm-project, this one is ready to go. Although the LLVM part is on release_13x branch, this won't disrupt those who need to build classic flang with older LLVM versions: the new features introduced by this PR just won't be in use.

bryanpkc (approved):
LGTM. But I wish that `vscale_range_min` and `vscale_range_max` didn't have to be members of the global `FLG` structure. The command-line option names could also use a prefix (similar to Clang's `-m` options) to indicate that they are target-specific options.

pawosm-arm:
@bryanpkc I guess it's too late for changing the option names, the classic-llvm-project part which introduces them was already merged after a positive review. Also, I was using neutral naming already (previous attempt was referring to SVE explicitly), you never know if other architectures we support one day start to introduce their own variation of VLA. I'd better keep it as it is now.

PR #1217, opened 2022-01-27
Remove CI testing of release_11x and add release_13x 

pwisniewskimobica:
There is a CI error for release_11x branch. While discussing we came to the conclusion that no one is using release_11x and hence it is safe to remove from CI. Also adding release_13x.


kiranchandramohan (approved):
LGTM.

Merging directly since we are fixing a CI failure here.

PR #1219, opened 2022-02-04
[runtime] Fix memory leak when freeing FIO_FCB

bryanpkc:
Commit d3ee7f250c0331f7381d6bffe7b40037304f275e made the "name" field
in FIO_FCB objects a pointer to malloc'ed memory in all cases. When
freeing a FIO_FCB object, the name should be freed in all cases as well.

### DIFF
--- a/runtime/flang/close.c
+++ b/runtime/flang/close.c
@@ -65,8 +65,6 @@ __fortio_close(FIO_FCB *f, int flag)
     else if (f->status == FIO_SCRATCH)
       unlink(f->name);
 #endif
-
-    free(f->name);
   } else { /* stdin, stdout, stderr - just flush */
 #if defined(TARGET_OSX)
     if (f->unit != 5 && f->unit != -5)
--- a/runtime/flang/utils.c
+++ b/runtime/flang/utils.c
@@ -122,6 +122,11 @@ __fortio_alloc_fcb(void)
 extern void
 __fortio_free_fcb(FIO_FCB *p)
 {
+  if (p->name) {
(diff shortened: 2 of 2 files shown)


runtime/flang/close.c
```diff
@@ -65,15 +65,14 @@ __fortio_close(FIO_FCB *f, int flag)
...
   } else { /* stdin, stdout, stderr - just flush */
 #if defined(TARGET_OSX)
     if (f->unit != 5 && f->unit != -5)
 #endif
       if (__io_fflush(f->fp) != 0)
         return __fortio_error(__io_errno());
   }
+  free(f->name);
 
   __fortio_free_fcb(f); /*  free up FCB for later use  */
```
pawosm-arm, line 77: but why __fortio_free_fcb() does not take care of it?

bryanpkc: That's a very good point. I'll change the code.

pawosm-arm: Note that this function has lots of early `return`s in the middle. Is it safe to assume each early return means the FCB is still in use and should not be freed and nothing is leaking in memory?

pawosm-arm (approved):

runtime/flang/close.c
bryanpkc: I checked the error handling code. It seems that in all cases, the runtime will print an error messages, and either aborts the process completely, or stops all further read/write operations (`fioFcbTbls.error = TRUE`). Not freeing the FCB probably isn't a problem.

PR #1129, opened 2021-08-31
[flang1][flang2][runtime] Support quad-precision FP type base

zhaochuanfeng:
We plan to add Quad-Precision Float Point to classic flang. There are multiple phases were planed:
- Support quad-precision FP types in flang1/flang2
- Add I/O support for quad-precision FP numbers to runtime
- Add quad-precision FP intrinsic functions to runtime
- Support quad-precision FP complex types in flang1/flang2
- Add I/O support for quad-precision FP complex numbers to runtime
- Add quad-precision FP complex intrinsic functions to runtime

This PR is the main part of the first phase, it contains requirements as follow: 
- Specification for quad-precision floating point type (quadFP)
    - REAL(KIND=16), REAL(16), REAL*16
- Syntax for quad-precision FP literals
   - 1.0_16, 1.0Q3
- Type conversion between quad-precision FP and other types
- Basic operations
  - Arithmetics: +, -, *, /, **
  - Comparison: >, <, >=, <=, ==, /=
  - Assignment: =
- Actual arguments, dummy arguments, return types
- Several necessary intrinsics for verification

PR #1222, opened 2022-02-09
[runtime] Fix bug in iso_fortran_env.f90 (encountered in OpenMPI)

zhaochuanfeng:
After #1129 commited, a regression is generated when compiling OpenMPI. 
_Originally posted by @pawosm-arm in https://github.com/flang-compiler/flang/issues/1129#issuecomment-1030595159_

This PR fix this issue by correctly declaring REAL128 kind parameter.

### DIFF
--- a/runtime/flang/iso_fortran_env.f90
+++ b/runtime/flang/iso_fortran_env.f90
@@ -59,14 +59,14 @@
 	integer REAL64
 	parameter (REAL64 = 8)
 	integer REAL128
-	parameter (REAL128 = -1)
+	parameter (REAL128 = 16)
 
         integer INTEGER_KINDS(4)
         parameter (INTEGER_KINDS = (/INT8, INT16, INT32, INT64/))
         integer LOGICAL_KINDS(4)
         parameter (LOGICAL_KINDS = (/LOGICAL8, LOGICAL16, LOGICAL32, LOGICAL64/))
-        integer REAL_KINDS(2)
-        parameter (REAL_KINDS = (/REAL32, REAL64/))
+        integer REAL_KINDS(3)
+        parameter (REAL_KINDS = (/REAL32, REAL64, REAL128/))
 
         end module  ISO_FORTRAN_ENV
 
--- a/tools/flang1/flang1exe/semutil2.c
+++ b/tools/flang1/flang1exe/semutil2.c
@@ -9544,6 +9544,10 @@ eval_selected_real_kind(ACL *arg)
     r = 4;
   else if (con <= 15)
     r = 8;
+#ifdef TARGET_SUPPORTS_QUADFP
+  else if (con <= MAX_EXP_OF_QMANTISSA)
+    r = REAL_16;
+#endif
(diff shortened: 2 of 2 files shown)


pawosm-arm (approved):

pawosm-arm (changes requested):
This indeed fixes the problem for OpenMPI. Still I'm seeing the same error message in nag test suite, e.g. pd5e.f90 test:
```
$HOME/upstream/llvm-debug/bin/flang -c pd5e.f90
F90-S-0081-Illegal selector - KIND value must be non-negative  (pd5e.f90: 4)
  0 inform,   0 warnings,   1 severes, 0 fatal for pd5e
F90-S-0081-Illegal selector - KIND value must be non-negative  (pd5e.f90: 7)
F90-S-0081-Illegal selector - KIND value must be non-negative  (pd5e.f90: 22)
F90-S-0081-Illegal selector - KIND value must be non-negative  (pd5e.f90: 23)
F90-S-0081-Illegal selector - KIND value must be non-negative  (pd5e.f90: 24)
  0 inform,   0 warnings,   3 severes, 0 fatal for test
```
First seven lines are:

pawosm-arm:
I think this PR is premature. It would be better to wait for the completion of Quad-FP works.

I've tried the new changes in our CI, but sadly, although pd5e test case compiles with the new patch, it segfaults when executed:
```
Test pd5e FAILED: on execution:
*** res/pd5e.res        Thu Feb 10 15:27:47 2022
--- pd5e.out    Thu Feb 10 15:28:02 2022
***************
*** 1,4 ****
!  F
!  T
!       1.2345678901234567890123
!  ok
--- 1 ----
! Segmentation fault (core dumped)
```

bryanpkc:
@pawosm-arm I can reproduce the segfault even after I reverted commit 46ec115bed6c. In fact, the integer test case pd5.f90 also crashes for me. The problem is that Flang doesn't seem to handle `value, optional` dummy arguments correctly, and generates LLVM IR for the actual argument that loads from a null pointer. Does pd5.f90 run fine in your tests?

[Our wiki page on Fortran 2008 features](https://github.com/flang-compiler/flang/wiki/Fortran-2008) seems to list this as a known unsupported feature.

pawosm-arm:
@bryanpkc this only confirms my suspicion that with QuadFP enabled, more tests are executed. I've looked into our CI logs and see that after reverting QuadFP commits, the only `pd`-prefixed tests that were executed are `pd0.f`, `pd0a.f`, `pd0b.f` and `pd0c.f`. So I assume, `pd5` and `pd5e` cases need to be added to disabled list. Sadly, this still does not end the problem, as even with this PR applied, there are more of the test cases failing at compile time with `KIND value must be non-negative`.

pawosm-arm (approved):
This patch fixes the initial issue with OpenMPI, so it can be merged. All other issues need to be addressed with further actions.

bryanpkc:
Thanks for the approvals.

PR #1223, opened 2022-02-15
introduce dm option

pwisniewskimobica:
It was decided to add clang -dM option to flang. 
It needs to print out all the macros defined - similarly as clang or gfortran does it.


test/sema/dm_option.f90
```diff
@@ -0,0 +1,15 @@
...
+! RUN: %flang -dM -E %s | FileCheck %s --check-prefix=CHECK-MACROS
+!
+! CHECK-MACROS: __amd64__
+! CHECK-MACROS: __SIZE_TYPE__
+! CHECK-MACROS: __FLANG_MAJOR__
+! CHECK-MACROS: __FLANG_MINOR__
+! CHECK-MACROS: __FLANG_PATCHLEVEL__
+program hello
+  print *, "hello world"
+end program hello
```
pawosm-arm, line 15: missing newline at EOF

tools/flang1/flang1exe/accpp.c
```diff
@@ -701,6 +701,20 @@ print_and_check(FILE *ff, char *str, char end_c)
     fputc(end_c, ff);
 }
 
+extern bool g_list_macros;
```
pawosm-arm, line 704: why not move it to some header?

tools/flang1/flang1exe/fpp.c
```diff
@@ -376,6 +376,20 @@ skipbl(char *tokval, int flag)
   return toktyp;
 }
 
+extern bool g_list_macros;
```
pawosm-arm, line 379: why not move it to some header?

tools/flang1/flang1exe/fpp.c
```diff
@@ -533,6 +547,20 @@ fpp(void)
     /* undef it */
     delete (*cp); /* should check if IDENT */
   }
+  if(g_list_macros)
```
pawosm-arm, line 550: if<SPACE>(<condition>)<SPACE>{

tools/flang1/flang1exe/fpp.c
```diff
@@ -533,6 +547,20 @@ fpp(void)
...
+  {
+    fpp_display_macros();
+
+    FREE(argbuf);
+    FREE(deftab);
+    FREE(hashrec);
+    if (incllist != 0) {
```
pawosm-arm, line 557: ...and here you knew how to format it properly... strange...

tools/flang1/flang1exe/accpp.c
```diff
@@ -913,6 +927,22 @@ accpp(void)
...
+  if (g_list_macros) {
+    acc_display_macros();
+    
+    FREE(deftab);
+    FREE(hashrec);
+    /* FREE(inclstack); Do not free this, gbl.curr_file points to it */
+    FREE(_ifs);
+    FREE(linebuf);
+
+    if (incllist != 0) {
```
pawosm-arm, line 939: please do not compare pointers with bare integers... `if (incllist != NULL)` or just `if (!incllist)` would do.

tools/flang1/flang1exe/accpp.c
```diff
@@ -701,6 +701,20 @@ print_and_check(FILE *ff, char *str, char end_c)
     fputc(end_c, ff);
 }
 
+extern bool g_list_macros;
+void acc_display_macros()
```
pawosm-arm, line 705: is this function used in any other file? should be annotated with `static`.

tools/flang1/flang1exe/fpp.c
```diff
@@ -533,6 +547,20 @@ fpp(void)
...
+      FREE(incllist);
+    }
+    
+    flg.es=TRUE;
```
pawosm-arm, line 561: `flg.es<SPACE>=<SPACE>TRUE;`

tools/flang1/flang1exe/main.c
```diff
@@ -94,6 +94,7 @@ static const char *feature = "pgfortran";
 static const char *feature = "pgfortran";
 #endif
 
+bool g_list_macros = FALSE;
```
pawosm-arm, line 97: is this variable read anywhere? I can see it written below, but never actually read.

tools/flang1/flang1exe/main.c
```diff
@@ -761,6 +763,7 @@ init(int argc, char *argv[])
   register_boolean_arg(arg_parser, "es", (bool *)&(flg.es), false);
   register_boolean_arg(arg_parser, "pp", (bool *)&(flg.p), false);
 
+  register_boolean_arg(arg_parser, "list-macros", &list_macros, false);
   /* Set values form command line arguments */
```
pawosm-arm, line 767: add one empty line in between. The author of the code below wanted it to be separated from the `register...` part and we should obey that.

tools/flang1/flang1exe/main.c
```diff
@@ -761,6 +763,7 @@ init(int argc, char *argv[])
   register_boolean_arg(arg_parser, "es", (bool *)&(flg.es), false);
   register_boolean_arg(arg_parser, "pp", (bool *)&(flg.p), false);
 
+  register_boolean_arg(arg_parser, "list-macros", &list_macros, false);
```
pawosm-arm, line 766: add one empty line below this one. The author of the code below wanted it to be separated from the register... part and we should obey that.

pawosm-arm (changes requested):
Although this code has a good chance to work, I'm having a lot of typographic remarks.
- why `extern` variable isn't defined in a header file?
- please annotate functions not used in other files with `static`. And if they are used in other files, please provide their declarations in a header file.
- stop mixing upper/lower case TRUE/FALSE vs true/false; please maintain similarity with the rest of the file; brief look at the file reveals that lowercase is used for boolean literals
- remember to put spaces in `if`, `for`, `while` and other constructs, namely `if<space>(<condition>)`
- do not move `{` to the next line in `if`, `for`, `while`, etc.; DO that ONLY when opening function definitions.
- keep pointer declaration convention of the file!!! it is `char *variable`, not `char* variable`
- always put space after comas, e.g. `strcmp(name,<space>"__LINE__")`
- and always put space around operators, e.g. `<space>==<space>`
this list could go on forever... Just try to use clang-format over your code changes (but NOT over entire file: you're responsible for your changes only, not for the entire file, if it's non-conformant already, so be it).

pawosm-arm:
And pleas change the commit message. dm could mean anything, Depeche Mode or anything else. From the context I know it's `-dM`, but the person who will read it in the future may not be lucky enough.

tools/flang1/flang1exe/fpp.c
pawosm-arm: but still, please don't compare pointers with merely integers.

tools/flang1/flang1exe/accpp.c
```diff
@@ -701,6 +701,20 @@ print_and_check(FILE *ff, char *str, char end_c)
...
+{
+  for (unsigned int i = 1; i<next_hash; ++i) {
+    char* name = &deftab[hashrec[i].name];
+    char* value = &deftab[hashrec[i].value];
+    if((strcmp(name,"__LINE__")==0) ||
+       (strcmp(name,"__DATE__")==0) ||
+       (strcmp(name,"__FILE__")==0) ||
+       (strcmp(name,"__TIME__")==0)) continue;
+    printf("%s : %s\r\n",name,value);
```
pawosm-arm, line 714: there is a good chance it would be the only use of `\r\n` in the entire file. This reminds me my DOS years, but that was more than 30 years ago. `\n` is enough to end a line.

tools/flang1/flang1exe/fpp.c
```diff
@@ -376,6 +376,20 @@ skipbl(char *tokval, int flag)
...
+void fpp_display_macros()
+{
+  for (unsigned int i = 1; i<next_hash; ++i) {
+    char* name = &deftab[hashrec[i].name];
+    char* value = &deftab[hashrec[i].value];
+    if((strcmp(name,"__LINE__")==0) ||
+       (strcmp(name,"__DATE__")==0) ||
+       (strcmp(name,"__FILE__")==0) ||
+       (strcmp(name,"__TIME__")==0)) continue;
+    printf("%s : %s\r\n",name,value);
```
pawosm-arm, line 389: `\n` should suffice to terminate a line.

tools/flang1/flang1exe/accpp.c
```diff
@@ -701,6 +701,20 @@ print_and_check(FILE *ff, char *str, char end_c)
     fputc(end_c, ff);
 }
 
+extern bool g_list_macros;
+void acc_display_macros()
+{
+  for (unsigned int i = 1; i<next_hash; ++i) {
```
pawosm-arm, line 707: this is C code, so although today's compilers understand this C++ way of defining variables, you should follow C rules.
should be:
```
{
  unsigned int i;

  for (i = 1U; i < next_hash; ++i) {
```

tools/flang1/flang1exe/fpp.c
```diff
@@ -376,6 +376,20 @@ skipbl(char *tokval, int flag)
   return toktyp;
 }
 
+extern bool g_list_macros;
+void fpp_display_macros()
+{
+  for (unsigned int i = 1; i<next_hash; ++i) {
```
pawosm-arm, line 382: see my comment above on defining variables in C

tools/flang1/flang1exe/fpp.c
```diff
@@ -376,6 +376,20 @@ skipbl(char *tokval, int flag)
   return toktyp;
 }
 
+extern bool g_list_macros;
+void fpp_display_macros()
```
pawosm-arm, line 380: This code looks very similar as in the other file. Personally, I would move it to some header and create `inline` function in there (or a `do..while (0)` preprocessor macro, but that requires some experience). All code duplications are bad and create maintenance nightmare. But since these two files are already plagued with duplication, I'm leaving decision whether to move it or to leave it to your personal taste.

tools/flang1/flang1exe/main.c
pawosm-arm: ok, I'm seeing it now, this is the variable shared across those files. it is ok to define it and initialize it here, but it's declaration (with `extern`) should also appear in `main.h`.

pawosm-arm, line 97: Please declare it (with `extern`) in main.h and include `"main.h"` in both fpp.c and accpp.c

pawosm-arm: Actually, avoid main.h. As I looked into it, I've found it terribly messed up. Use global.h instead, it is already included by fpp.c and accpp.c

tools/flang1/flang1exe/fpp.c
pawosm-arm: Namely, to global.h, which is already included by this file.

pwisniewskimobica:
Other PR created as duplication for it:
https://github.com/flang-compiler/flang/pull/1226

pwisniewskimobica:
Other PR created as duplication for it:
https://github.com/flang-compiler/flang/pull/1226

PR #1228, opened 2022-02-23
[DebugInfo] Artificial var debug info for common block should not be gererated

alokkrsharma:
Currently a debug metadata of artificial global variable type for common block is generated (in addition to user visible common block members). With Flang extension of LLVM it has no name, otherwise it has name. Both the cases are problematic and redundant. which should be removed.

Please consider below test case.
```
program main
  integer :: cvar1, cvar2
  common /cname/ cvar1, cvar2
  cvar1 = 1
  cvar2 = 2
  print *, cvar1
  print *, cvar2
end program main
```
Generated IR.

```
!1 = distinct !DIGlobalVariable(name: "cvar1", scope: !2, file: !4, type: !16, isLocal: false, isDefinition: true)
!2 = distinct !DICommonBlock(scope: !3, declaration: !9, name: "cname")
!8 = !DIGlobalVariableExpression(var: !9, expr: !DIExpression())
!9 = distinct !DIGlobalVariable(scope: !2, type: !10, isLocal: false, isDefinition: true, flags: DIFlagArtificial)
!14 = !DIGlobalVariableExpression(var: !15, expr: !DIExpression(DW_OP_plus_uconst, 4))
!15 = distinct !DIGlobalVariable(name: "cvar2", scope: !2, file: !4, type: !16, isLocal: false, isDefinition: true)
```
!9 is artificial global variable, and when we allow the name, it is 
```
!9 = distinct !DIGlobalVariable(name: "cname", scope: !2, type: !10, isLocal: false, isDefinition: true)
```
It translates to debug info as 
```
0x00000040:     DW_TAG_common_block
                  DW_AT_name    ("cname")
                  DW_AT_location        (DW_OP_addr 0x0)

0x0000004f:       DW_TAG_variable
                    DW_AT_name  ("cname")
                    DW_AT_type  (0x000000a2 "byte[8]")
                    DW_AT_external      (true)
                    DW_AT_location      (DW_OP_addr 0x0)

0x00000062:       DW_TAG_variable
                    DW_AT_name  ("cvar1")
                    DW_AT_type  (0x000000bc "integer")
                    DW_AT_external      (true)
                    DW_AT_location      (DW_OP_addr 0x0)

0x00000075:       DW_TAG_variable
                    DW_AT_name  ("cvar2")
                    DW_AT_type  (0x000000bc "integer")
                    DW_AT_external      (true)
                    DW_AT_location      (DW_OP_addr 0x0, DW_OP_plus_uconst 0x4)

0x0000008a:       NULL
``` 
Please note the DIE (0x0000004f) is artificial global var.
when it has name, it fails gdb test case gdb.fortran/common-block.exp.
In case it has no name it is useless and results space penalty.
Only use of it is the address it has which is anyway already available at DIE tagged as DW_TAG_common_block.

So it is redundant and problematic and we should get rid of it.

### DIFF
--- a/tools/flang2/flang2exe/lldebug.cpp
+++ b/tools/flang2/flang2exe/lldebug.cpp
@@ -4085,50 +4085,8 @@ lldbg_create_cmblk_mem_mdnode_list(SPTR sptr, SPTR gblsym)
 }
 
 static LL_MDRef
-lldbg_create_cmblk_gv_mdnode(LL_DebugInfo *db, LL_MDRef cmnblk_mdnode,
-                             SPTR sptr)
-{
-  LL_MDRef mdref, type_mdnode, subscripts_mdnode, subscript_mdnode,
-      elem_type_mdnode;
-  DBLINT64 align;
-  ISZ_T sz, lb, ub, dim_ele;
-  DTYPE elem_dtype;
-  LLMD_Builder mdb = llmd_init(db->module);
-  const char *display_name;
-
-  elem_dtype = DT_BINT;
-  sz = SIZEG(sptr);
-  dim_ele = sz - 1;
-  lb = 0;
-  ub = dim_ele;
-  align[1] = ((alignment(elem_dtype) + 1) * 8);
-  align[0] = 0;
-  if (ll_feature_debug_info_ver90(&db->module->ir))
-    subscript_mdnode = lldbg_create_subrange_mdnode(
-        db, ll_get_md_i64(db->module, lb), ll_get_md_i64(db->module, ub),
-        ll_get_md_null());
-  else
-    subscript_mdnode = lldbg_create_subrange_mdnode_pre11(db, lb, sz);
-  llmd_add_md(mdb, subscript_mdnode);
-  sz *= ZSIZEOF(elem_dtype) * 8;
-  elem_type_mdnode =
-      lldbg_emit_type(db, elem_dtype, sptr, 1, false, false, false);
-  subscripts_mdnode = llmd_finish(mdb);
-  type_mdnode = lldbg_create_array_type_mdnode(
-      db, ll_get_md_null(), 0, sz, align, elem_type_mdnode, subscripts_mdnode,
-      ll_get_md_null(), ll_get_md_null(), ll_get_md_null(), ll_get_md_null());
-  display_name = SYMNAME(sptr);
-  mdref = lldbg_create_global_variable_mdnode(
-      db, cmnblk_mdnode, display_name, SYMNAME(sptr), "", ll_get_md_null(),
-      DECLLINEG(sptr), type_mdnode, 0, 1, NULL, -1, DIFLAG_ARTIFICIAL, 0,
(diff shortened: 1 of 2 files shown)


tools/flang2/flang2exe/lldebug.cpp
```diff
@@ -4089,11 +4089,6 @@ lldbg_emit_common_block_mdnode(LL_DebugInfo *db, SPTR sptr)
   cmnblk_mdnode = lldbg_create_common_block_mdnode(
       db, scope_modnode, ll_get_md_null(), SYMNAME(sptr));
```
bryanpkc, line 4090: Do we still need a null metadata node for the optional `declaration` field?

tools/flang2/flang2exe/lldebug.cpp
```diff
@@ -4090,11 +4090,6 @@ lldbg_emit_common_block_mdnode(LL_DebugInfo *db, SPTR sptr)
...
   db->cur_cmnblk_mdnode = cmnblk_mdnode;
-  cmnblk_gv_mdnode = lldbg_create_cmblk_gv_mdnode(db, cmnblk_mdnode, sptr);
-  slot = LL_MDREF_value(cmnblk_gv_mdnode) - 1;
-  node = db->module->mdnodes[slot];
```
kiranchandramohan, line 4095: Looks like `slot` and `node` have become unused variables and hence the CI is failing.
```
/home/runner/work/flang/flang/tools/flang2/flang2exe/lldebug.cpp:4081:14: error: unused variable 'node' [-Werror,-Wunused-variable]
  LL_MDNode *node;
             ^
/home/runner/work/flang/flang/tools/flang2/flang2exe/lldebug.cpp:4082:12: error: unused variable 'slot' [-Werror,-Wunused-variable]
  unsigned slot;```

alokkrsharma: Thanks Kiran. After the change, tests passed.

tools/flang2/flang2exe/lldebug.cpp
alokkrsharma: Sorry. I am not sure whether I got this comment correctly. Do you mean we should get rid of this field as it is unused ?

pawosm-arm (approved):

tools/flang2/flang2exe/lldebug.cpp
bryanpkc: My understanding was that this null node is meant for the `declaration` field in a `DICommonBlock` metadata, which according to `LLParser.cpp` is not a required field. So I was just asking if `lldbg_create_common_block_mdnode` could be modified to not require this parameter, since the actual argument is always `ll_get_md_null()`.

alokkrsharma: Thanks @bryanpkc for clarification.
Yes, correct. I have updated patch for this.

bryanpkc (approved):

PR #1169, opened 2021-11-08
[Debug] Display common block variable in sub-routines

vigbalu:
Issue: Common block variables are shown as “optimized out” in subroutines. (Due to missing metadata references).
Fix: In the “write_comm” function, instead of emitting the global variable with defn to IR, we capture the same into a member “cmblk_init_data” of variable agb.  At the end of processing a source file, in the function “assemble_end”, we will generate the same with all the debug info metadata references.

PR #1233, opened 2022-03-21
[DebugInfo] Display common block variable in sub-routines

vigbalu:
Issue: Common block variables are shown as “optimized out” in subroutines. (Due to missing metadata references).
Fix: In the  “write_comm” function, instead of emitting the global variable with defn to IR, we capture the same into a member “cmblk_init_data” of variable agb.  At the end of processing a source file, in the function “assemble_end”, we will generate the same with all the debug info metadata references.

Note: This is same as #1169. Review comments are addressed here.

### DIFF
--- a/test/debug_info/cmn_blk_dbg_info.f90
+++ b/test/debug_info/cmn_blk_dbg_info.f90
@@ -1,0 +1,45 @@
+!RUN: %flang -gdwarf-4 -S -emit-llvm %s -o - | FileCheck %s
+
+!CHECK-DAG: ![[B3i1:[0-9]+]] = distinct !DIGlobalVariable(name: "b3i", scope: ![[BLK:[0-9]+]]
+!CHECK-DAG: ![[B3j1:[0-9]+]] = distinct !DIGlobalVariable(name: "b3j", scope: ![[BLK]]
+!CHECK-DAG: ![[BLK]] = distinct !DICommonBlock(scope: ![[SCOPE1:[0-9]+]]
+!CHECK-DAG: ![[SCOPE1]] = distinct !DISubprogram(name: "bdata"
+
+!CHECK-DAG: ![[B3i2:[0-9]+]] = distinct !DIGlobalVariable(name: "b3i", scope: ![[MOD:[0-9]+]]
+!CHECK-DAG: ![[B3j2:[0-9]+]] = distinct !DIGlobalVariable(name: "b3j", scope: ![[MOD]]
+!CHECK-DAG: ![[MOD]] = !DIModule({{.*}}, name: ".blockdata."
+
+!CHECK-DAG: ![[B3i3:[0-9]+]] = distinct !DIGlobalVariable(name: "b3i", scope: ![[SUB:[0-9]+]]
+!CHECK-DAG: ![[B3j3:[0-9]+]] = distinct !DIGlobalVariable(name: "b3j", scope: ![[SUB]]
+!CHECK-DAG: ![[SUB]] = distinct !DICommonBlock(scope: ![[SCOPE2:[0-9]+]]
+!CHECK-DAG: ![[SCOPE2]] = distinct !DISubprogram(name: "sub_block_data"
+
+!CHECK-DAG: ![[REF1:[0-9]+]] = !DIGlobalVariableExpression(var: ![[B3i1]]
+!CHECK-DAG: ![[REF2:[0-9]+]] = !DIGlobalVariableExpression(var: ![[B3j1]]
+!CHECK-DAG: ![[REF3:[0-9]+]] = !DIGlobalVariableExpression(var: ![[B3i2]]
+!CHECK-DAG: ![[REF4:[0-9]+]] = !DIGlobalVariableExpression(var: ![[B3j2]]
+!CHECK-DAG: ![[REF5:[0-9]+]] = !DIGlobalVariableExpression(var: ![[B3i3]]
+!CHECK-DAG: ![[REF6:[0-9]+]] = !DIGlobalVariableExpression(var: ![[B3j3]]
+!CHECK-DAG: @blk_ = global{{.*}}![[REF1]]{{.*}}![[REF2]]{{.*}}![[REF3]]{{.*}}![[REF4]]{{.*}}![[REF5]]{{.*}}![[REF6]]
+
+PROGRAM bdata
+   integer b3i, b3j, adr
+   COMMON/BLK/b3i, b3j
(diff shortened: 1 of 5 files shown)


tools/flang2/flang2exe/llassem.cpp
```diff
@@ -1162,11 +1162,16 @@ assemble_end(void)
...
       tdefsym = find_ag(gname);
       typed = AG_TYPENAME(tdefsym);
-      fprintf(ASMFIL, "%%struct%s = type < { %s } > \n", name, typed);
-      fprintf(ASMFIL, "@%s = %s global %%struct%s ", name,
-              AG_ISMOD(gblsym) ? "external" : "common", name);
-      fprintf(ASMFIL, "%s, align %d",
-              AG_ISMOD(gblsym) ? "" : " zeroinitializer", align_value);
+      if (AG_CMBLKINITDATA(gblsym)) {
+        fputs (AG_CMBLKINITDATA(gblsym), ASMFIL);
+        free (AG_CMBLKINITDATA(gblsym));
```
bryanpkc, line 1167: Please avoid the dangling pointer:
```c
AG_CMBLKINITDATA(gblsym) = NULL;
```

tools/flang2/flang2exe/llassem.cpp
```diff
@@ -1698,6 +1703,19 @@ write_comm(void)
 
     align_value = CACHE_ALIGN + 1;
 
+    /*
+      Create a tmp file and capture the initialization data.
+      Emitting the cmn blk var defn now will miss out the dbg info
+      generated for other references of this variable.
+    */
+    FILE *llvm_ir_bkup = gbl.asmfil;
+    FILE *fd;
```
bryanpkc, line 1712: I suggest renaming these variables to clarify their purpose. `fd` is not a terribly useful name.

tools/flang2/flang2exe/llassem.cpp
vigbalu: thanks for the catch, Done.

tools/flang2/flang2exe/llassem.cpp
```diff
@@ -1698,6 +1703,19 @@ write_comm(void)
...
+    if ((fd = tmpfile()) == NULL)
+      errinfo(F_0005_Unable_to_open_temporary_file);
+    else
+      gbl.asmfil = fd;
+
+
```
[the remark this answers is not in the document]
vigbalu: Removed the extra line.

tools/flang2/flang2exe/llassem.cpp
vigbalu: Changed. Thanks.

tools/flang2/flang2exe/llassem.cpp
```diff
@@ -1162,11 +1162,17 @@ assemble_end(void)
...
       tdefsym = find_ag(gname);
       typed = AG_TYPENAME(tdefsym);
-      fprintf(ASMFIL, "%%struct%s = type < { %s } > \n", name, typed);
-      fprintf(ASMFIL, "@%s = %s global %%struct%s ", name,
-              AG_ISMOD(gblsym) ? "external" : "common", name);
-      fprintf(ASMFIL, "%s, align %d",
-              AG_ISMOD(gblsym) ? "" : " zeroinitializer", align_value);
+      if (AG_CMBLKINITDATA(gblsym)) {
+        fputs (AG_CMBLKINITDATA(gblsym), ASMFIL);
+        free (AG_CMBLKINITDATA(gblsym));
```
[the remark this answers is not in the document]
vigbalu: Done.

tools/flang2/flang2exe/llassem.cpp
```diff
@@ -1698,6 +1704,18 @@ write_comm(void)
...
 
+    /*
+      Create a tmp file and capture the initialization data.
+      Emitting the cmn blk var defn now will miss out the dbg info
+      generated for other references of this variable.
+    */
+    FILE *llvm_ir_bkup = gbl.asmfil;
+    FILE *cmn_blk_ir;
+    if ((cmn_blk_ir = tmpfile()) == NULL)
+      errinfo(F_0005_Unable_to_open_temporary_file);
```
bryanpkc, line 1715: Shouldn't this call `errfatal` just like all the failing `fopen` attempts in `tools/flang2/flang2exe/main.cpp`?

vigbalu: My bad. Changed.

test/debug_info/cmn_blk_dbg_info.f90
```diff
@@ -0,0 +1,25 @@
+!RUN: %flang -gdwarf-4 -S -emit-llvm %s -o - | FileCheck %s
+
+!CHECK: @blk_ = global {{.*}} align 64, !dbg !{{[0-9]+}}, !dbg !{{[0-9]+}}, !dbg !{{[0-9]+}}, !dbg !{{[0-9]+}}, !dbg !{{[0-9]+}}, !dbg !{{[0-9]+}}
```
bryanpkc, line 3: I ran this test case, and have two questions.

1. There are actually 9 `!dbg` metadata nodes associated with `@blk_`. The FileCheck pattern here matches only six. Moreover, you are not checking that the metadata refer to the declarations of `b3i` and `b3j` on three different lines, which I feel is the purpose of this test.
2. By debugging `llassem.cpp`, I found that this test does not actually exercise the newly added code following `if (cmn_blk_ir) {` (line 1738), even though the output IR does contain the additional `!dbg` metadata nodes as expected. Isn't this test supposed to provide coverage for the newly added code?

vigbalu: Earlier, the intention was to check whether the global "@blk" common block variable had at least 6 reference one for ever b3i and b3j in 3 different scopes. Before the fix, only 2 metadata nodes were associated with @blk even though it were present in the "ll" file.
Now changed the CHECKs to verify that "@blk" variable associated metadata nodes are comes from every scope.
Please let me know if you feel too many CHECK lines. I restrict to one variable for each scope.

bryanpkc: That looks much better. Thanks!

bryanpkc (approved):

kiranchandramohan (approved):
LGTM. I have reviewed this in https://github.com/flang-compiler/flang/pull/1169.

kiranchandramohan:
@shivaramaarao can this be merged?

Issue #1211, opened 2022-01-21
Disable quad-precision fp on Windows ARM64


tdusnoki:
Currently Windows on ARM64 flang build fails with undefined symbol linker errors:

```lld-link: error: undefined symbol: __floatsitf
referenced by runtime\flang\CMakeFiles\flang_shared.dir\ieee_arithmetic.F95.obj:(ieee_arithmetic_ieee_logbr16_)
referenced by runtime\flang\CMakeFiles\flang_shared.dir\ieee_arithmetic.F95.obj:(ieee_arithmetic_ieee_logbr16_)

lld-link: error: undefined symbol: __extendsftf2
referenced by runtime\flang\CMakeFiles\flang_shared.dir\ieee_arithmetic.F95.obj:(ieee_arithmetic_ieee_rem4x16_)
referenced by runtime\flang\CMakeFiles\flang_shared.dir\ieee_arithmetic.F95.obj:(ieee_arithmetic_ieee_rem16x4_)

lld-link: error: undefined symbol: __extenddftf2
referenced by runtime\flang\CMakeFiles\flang_shared.dir\ieee_arithmetic.F95.obj:(ieee_arithmetic_ieee_rem8x16_)
referenced by runtime\flang\CMakeFiles\flang_shared.dir\ieee_arithmetic.F95.obj:(ieee_arithmetic_ieee_rem16x8_)
```

Disabling `TARGET_SUPPORTS_QUADFP` for Windows on ARM64 platform should fix the build until we implement these functions later.

PR #1235, opened 2022-03-25
AArch64: add ability to disable QUADFP

pawosm-arm:
To unblock those who not always need this feature, this commit adds ability to disable it in CMake invocation.


kaadam:
We're waiting for the last approval for this #1211, but seems your PR is more general,  maybe we should close #1211.

bryanpkc (approved):
LGTM.

pawosm-arm:
I've presented this one prematurely. I'm still having problems with it.

bryanpkc:
Converting CMake values into Python values can be tricky! But that's implementation details; the design looks fine.

pawosm-arm:
Sadly, this causes more damage than improvement in our local builds, so I'm giving up on this.

pawosm-arm:
What it needs is an equivalent of `llvm_canonicalize_cmake_boolean`, something that is used throughout LLVM's CMake files. It may deserve a separate PR to introduce this missing bit.

PR #1239, opened 2022-04-01
Fix path in demo call

ianhbell:


bryanpkc (approved):

pawosm-arm (approved):
Don't know why X86 received a treatment so special, but otherwise, looks OK

shivaramaarao (approved):

PR #1242, opened 2022-05-03
[directives] Make VECTOR test cases compatible with LLVM 14

bryanpkc:
With the `!DIR$ VECTOR ALWAYS` directive, LLVM 14 vectorizes the test case
`test/directives/vector.f90`, but the chosen VPlan replicates the `i32` load
instead of widening it; the `i32` store at the end of the loop is still
widened. This patch changes the FileCheck patterns in the test case to
match only the widened store.

For `test/directives/vector_directive3.f90`, LLVM 14 also replicates the `i32`
load instead of widening it. Furthermore, the test case originally matched
`add <2 x i64>` which referred to the vectorized add of section descriptor
values (rather than actual array elements). This patch changes the FileCheck
patterns in the test case to match the vectorized add of the `i32` array
elements.

The release_13x branch of classic-flang-llvm-project still passes the
updated tests.

### DIFF
--- a/test/directives/vector.f90
+++ b/test/directives/vector.f90
@@ -15,6 +15,5 @@ subroutine add(arr1,arr2,arr3,N)
 end subroutine
 ! METADATA: !"llvm.loop.vectorize.enable", i1 true
 ! IGNORE-DIRECTIVES-NOT: !"llvm.loop.vectorize.enable", i1 true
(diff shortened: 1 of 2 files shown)


pawosm-arm (approved):

Chirag-Khandelwal (approved):

kiranchandramohan:
@bryanpkc I think you have the approvals to merge.

PR #1243, opened 2022-05-04
Fix the im30.f test case

pawosm-arm:
LLVM15+ brings a new default for creation of the PIE binaries on Linux
systems. In effect, position independed executables are always
assumed.

With the recent change in the default for creation of PIE binaries,
some of the fragile test cases could start to fail.

In the im30 test case, the result of the first use of %loc() function
was truncated (due to the way the variable holding the result is
created), while the second use of %loc() was properly used (in the
comparison expression) as a 64-bit value. Up until recently it wasn't
hurting as the most significant bits were always 0. Sadly, it has
changed with the introduction of the PIE binaries.

Due to the way the `locf` array is created, we have no ability to
change its assumed type. In order to keep this test objectives intact,
the result of the second %loc() call is now kept in a 32-bit integer
variable, so the values used in the comparison are both of the same
type.

### DIFF
--- a/test/f90_correct/src/im30.f
+++ b/test/f90_correct/src/im30.f
@@ -23,10 +23,11 @@ c  --- check the results:
 	common /fast/ a(100)
 	real*8 a
 	real*8 x1wxyz
-	integer istuff
+	integer istuff, stuff
 	locf(x1wxyz) = ishft(%loc(x1wxyz),-3)
 	istuff = locf(a(1))
-	if (istuff .ne. ishft(%loc(a(1)),-3)) then
+	stuff = ishft(%loc(a(1)),-3)
+	if (istuff .ne. stuff) then
 	    if1 = 99
 	else
 	    if1 = 1


kiranchandramohan:
This is a really strange testcase. Is %LOC an extension? (listed so in llvm/flang https://releases.llvm.org/11.0.0/tools/flang/docs/Extensions.html)

BTW, `locf` is a statement function.

Different compilers seem to have different behaviours for this test.
-> gfortran produces an error message. Is there a flag to use?
```
  15 |         locf(x1wxyz) = ishft(%loc(x1wxyz),-3)
      |                       1
Error: The argument list functions %VAL, %LOC or %REF are not allowed in this context at (1)
```

-> llvm/flang crashes.

-> ifortran/ifx produces incorrect results for both the new and the old version.

kiranchandramohan (approved):
Although I have the comments above, the change in the test seems reasonable. Hence approving.

pawosm-arm:
This is a strange test case indeed, and LOC built-in functions are not portable across compilers. Namely, flang does not have locf, so it is valid to create a variable (or an array as it is done here) with that name. Flang defines %loc() and loc() functions (they both do the same), while gfortran defines locf() (also doing the same thing), this test case cannot be compiled with gfortran.

shivaramaarao (approved):
looks good to me.

PR #1244, opened 2022-05-05
[test] Adapt AArch64 target-features attribute tests for LLVM 14

bryanpkc:
In LLVM 14, the `-march=armv8-a` option adds `+v8a` to the `target-features`
attribute. This patch updates Flang test cases that match the attribute,
so that they will work with both LLVM 13 and LLVM 14.


pawosm-arm (approved):

pawosm-arm:
@RichBarton-Arm or @kiranchandramohan can any of you guys add your review?

PR #1248, opened 2022-05-15
[flang2] Fix generation of upperilm.h

bryanpkc:
The CMake recipe for running `upperl` to generate `upperilm.h` from `upperilm.in` incorrectly leaves out a terminating newline character, without which the parser inside `upperl` will not print the last opcode (`IM_KXOR`, a.k.a. `XOR8`) in its output. This patch adds back the missing newline and fixes the `XOR8` opcode.


pawosm-arm (approved):

kaadam (approved):
Good catch, LGTM.

Issue #872, opened 2020-02-19
Procedure pointers to internal procedures (part 2)


RichBarton-Arm:
This is a new issue similar to issue [#577 ](https://github.com/flang-compiler/flang/issues/577). The example below is adapted from that issue but with the procedure taking an array of unknown length as its argument rather than an integer.
```
module m
  implicit none
  abstract interface
    subroutine int_printer(i)
      integer, dimension(:) :: i  ! Change from previous example
    end subroutine
  end interface
  procedure(int_printer), pointer :: iprinter
contains
  subroutine work
    implicit none
    integer, dimension(2) :: i1 = (/ 1, 2 /)        ! Change from previous example
    call iprinter(i1)
  end subroutine
end module
subroutine sub
  use m
  implicit none
  integer num_prints
  num_prints = 0
  iprinter => myprint
  call work
  print *, 'num_prints=', num_prints
contains
  subroutine myprint(i)
    integer, dimension(:) :: i       ! Change from previous example
    print *, i
    num_prints = num_prints + 1
  end subroutine
end subroutine

use m
implicit none
call sub
! call work  ! error
end
```

When compiled at any optimisation level with flang built from the [latest commit](https://github.com/flang-compiler/flang/commit/cb42a171b6462b2f3acd3e1841a56a1ae08291ef) on AArch64 , the resulting image segfaults:
`> ./a.out 
            1            2
Segmentation fault
`
gfortran and F18 compile this correctly.

Looking at the generated image (built with -O3) the call through iprinter in work calls to the wrong address 

> objdump -rd example.o
0000000000000004 <m_work_>:
<snip>
5c:   90000008        adrp    x8, 60 <m_work_+0x5c>
                              5c: R_AARCH64_ADR_PREL_PG_HI21  _m_0_
60:   91000108        add     x8, x8, #0x0
                              60: R_AARCH64_ADD_ABS_LO12_NC   _m_0_
64:   f9402d02        ldr     x2, [x8, #88]
68:   f9400108        ldr     x8, [x8]

This regression was caused by the commit [82683130](https://github.com/flang-compiler/flang/commit/82683130aeebdbe6d5c70044cd79ef22e54f4f22) which was part of the fix for issue #577 .

RichBarton-Arm:
Reverting [82683130](https://github.com/flang-compiler/flang/commit/82683130aeebdbe6d5c70044cd79ef22e54f4f22) makes this go away while still fixing the original bug in #577.

[...]

PR #1252, opened 2022-05-24
A test confirming that the issue #872 is actually resolved.

pawosm-arm:
This change is based on the solution discussed in #872.

The test case code comes from the bug report.

### DIFF
--- a/test/f90_correct/inc/ptr_to_internal_proc.mk
+++ b/test/f90_correct/inc/ptr_to_internal_proc.mk
@@ -1,0 +1,22 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+EXE=ptr_to_internal_proc.$(EXESUFFIX)
+
+build:  $(SRC)/ptr_to_internal_proc.f90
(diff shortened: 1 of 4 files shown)


kiranchandramohan:
@pawosm-arm it might be helpful to expand the commit message outlining the issue and the fix.

tools/flang1/flang1exe/lowersym.c
```diff
@@ -5532,7 +5532,7 @@ static void
 llvm_check_retval_inargs(int sptr)
 {
   int fval = FVALG(sptr);
-  if (fval) {
+  if (fval && !IS_PROC_PTR_IFACEG(sptr)) {
```
liuxiao0909c, line 5535: I think the new added code may cannot be covered because all procedure in `ptr_to_internal_proc.f90` is SUBROUTINE and the field FVAL is 0. 
So may be the issue#872 has been fixed by other commit and we only need to revert [82683130](https://github.com/flang-compiler/flang/commit/82683130aeebdbe6d5c70044cd79ef22e54f4f22).

pawosm-arm:
@liuxiao0909c I'm giving your claim a try here. If all tests passes, you're statement is correct. Nevertheless, my PR stays as there is a value in the test case it adds (and it reverts the offending change too)

tools/flang1/flang1exe/lowersym.c
pawosm-arm: @liuxiao0909c you were right! I've updated my commit and removed the overhead change. But the PR itself stays: the new commit not only reverts offending commit but also adds a test case which allows closure of #872.

pawosm-arm:
The tests have passed, the PR is still valid as the test it provides can support closure of #872.

Issue #1253, opened 2022-05-24
Test case failing after PR #1212 (huge allocations for a small array)


pawosm-arm:
The mw4 test case which operates on just a 2-element array makes way too large allocations after merging PR #1212. Compare these (mw4-safe was compiled before merging this PR):
```
$ memusage -m ./mw4-safe
 ok

Memory usage summary: heap total: 104294, heap peak: 104230, stack peak: 1808
         total calls   total memory   failed calls
 malloc|         10         104294              0
realloc|          0              0              0  (nomove:0, dec:0, free:0)
 calloc|          0              0              0
   free|          2             64
mmap(r)|          0              0              0
mmap(w)|          0              0              0
mmap(a)|          0              0              0
 mremap|          0              0              0  (nomove: 0, dec:0)
 munmap|          0              0              0
Histogram for block sizes:
    0-15              5  50% ==================================================
   32-47              2  20% ====================
12288-12303           1  10% ==========
19200-19215           1  10% ==========
   large              1  10% ==========
$ memusage -m ./mw4
 ok

Memory usage summary: heap total: 9257652166, heap peak: 9257632928, stack peak: 2112
         total calls   total memory   failed calls
 malloc|         11     9257652166              0
realloc|          0              0              0  (nomove:0, dec:0, free:0)
 calloc|          0              0              0
   free|          2     9257547904
mmap(r)|          0              0              0
mmap(w)|          0              0              0
mmap(a)|          0              0              0
 mremap|          0              0              0  (nomove: 0, dec:0)
 munmap|          0              0              0
Histogram for block sizes:
    0-15              5  45% ==================================================
   32-47              2  18% ====================
12288-12303           1   9% ==========
19200-19215           1   9% ==========
   large              2  18% ====================
```
This causes failure in our CI where memory restrictions do apply.

pawosm-arm:
As it turned out, the problem with the code added by upstream PR#1212 is with the call to the `move_range_after()` function in the `semutil2.c` file.

The `move_range_before()` and `move_range_after()` functions were introduced by Gary Klimowicz a couple of years ago and never actually used, even in the commit that has introduced them. Therefore there is no good example on how this function can be used legally.

[...]

PR #1254, opened 2022-05-31
Add a test case exposing potential issues with implied do

pawosm-arm:
The purpose of this test is to demonstrate symptoms described
in the issue #1253

Note that this test case intercepts a call to the flang runtime
(the f90_alloc04a_i8() function) in order to verify that the
number of temporarily allocated elements do not exceed the
array size.


pawosm-arm:
This test is failing now (as expected) on `abort()`:
```
make: *** [/home/github/flang/test/f90_correct/lit/../inc/implied_do10.mk:16: run] Aborted (core dumped)
```
this is due to an issue I've signaled in #1253

pawosm-arm:
I've made a slight change which supports the evidence even more.

pawosm-arm:
This test case was finally added by PR #1277

PR #1255, opened 2022-06-03
Make those two tests safe to run in parallel

pawosm-arm:
Initially, I could not replicate this error outside the testing
environment. Then I figured out that both these tests share the
same names of the exported modules. In effect, two flangs running
in parallel are overwritting the same .mod files.

This patch renames the modules so they are different for each test.


pawosm-arm:
I'm merging it immediately in order to unblock the CI.

PR #1256, opened 2022-06-08
[flang1] Fix final subroutine attribute

PeixinQiao:
As Fortran 2018 7.5.6.1 NOTE 1, final subroutines are effectively always
"accessible". So, the final subroutine can never be private.


pawosm-arm (approved):

bryanpkc:
@shivaramaarao gentle ping

PR #1259, opened 2022-06-09
loop metadata: add two more test cases and improve loop discovery

pawosm-arm:
Those two test cases were supposed to be contributed to PR #1094

Unfortunately, at that time, the -emit-flang-llvm flag was not supported by the frontend driver yet.

### DIFF
--- a/test/llvm_ir_correct/loop-line-metadata.f90
+++ b/test/llvm_ir_correct/loop-line-metadata.f90
@@ -1,0 +1,60 @@
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+! REQUIRES: llvm-13
+
+! RUN: %flang -g -O2 -emit-flang-llvm %s -o /dev/stdout | FileCheck %s
+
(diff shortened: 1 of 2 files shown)


pawosm-arm:
I've added a small code change which mitigates the problem

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1800,12 +1800,9 @@ schedule(void)
           (void)cons_vec_always_metadata();
         }
         if ((check_for_loop_directive(ILT_LINENO(ilt), 191, 0x4))) {
```
bryanpkc, line 1802: It seems that this condition can be merged into the condition of the previous block, since the blocks are the same?

bryanpkc: @pawosm-arm ping?

pawosm-arm: done

pawosm-arm:
I suspect cgmain changes aren't needed after rebase

pawosm-arm:
These test cases are wrongly designed. With too many CHECK-DAG, the right occurrence of `!llvm.loop` is impossible to be found reliably (namely, a slight change in LLVM middle-end will break these test cases again). Therefore it's not worth to have them.

PR #1264, opened 2022-06-23
Set dwarf version to 4 in debug_info tests

pawosm-arm:
This PR is a series of the 3 small commits.

The LLVM14 uses dwarf version 5 by default, but those tests are only compatible with the older dwarf version.


bryanpkc (approved):
LGTM

bryanpkc:
I recommend squashing the commits before merging.

pawosm-arm:
I would normally do that, but these commits have different authors, and we have tendency to preserve authorship of our commits.

PR #1265, opened 2022-06-23
Fix flang testing with newer gcc versions

pawosm-arm:
When flang is being built with compiler version newer than
the one installed on the system, we need to make sure that
COMPILER_PATH variable is set for lit testing,
otherwise there might be compatibillity problems with the libgcc.a library.
Flang tests are building and executing programs, so some symbols
could be missing when those programs are linked with older libgcc.a
than the one used to build compiler.

### DIFF
--- a/test/lit.cfg
+++ b/test/lit.cfg
@@ -15,6 +15,17 @@ import tempfile
 
 import lit.formats
 import lit.util
+import lit.llvm
+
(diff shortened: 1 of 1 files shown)


pawosm-arm:
@bryanpkc I moved the change to the other file as you requested. Could you look at it once more?

PR #1270, opened 2022-07-07
portability: on Mac systems __WORDSIZE is defined in stdint.h

pawosm-arm:
Apparently, stdint.h must be included explicitly.

This is emergency fix as the previous commit might have broken builds for some.


bryanpkc (approved):

bryanpkc:
Don't we have a macOS CI job? Why did it not catch the problem in the previous PR?

PR #1271, opened 2022-07-10
portability: `__WORDSIZE` isn't defined in MinGW

mmuetzel:
Afaict, `__WORDSIZE` is only defined on glibc (and XCode?) platforms.
This change should fix compilation for MinGW.


bryanpkc (approved):

kaadam (approved):
LGTM

PR #1272, opened 2022-07-11
portability: Fix alignment and size for LLP64

mmuetzel:
On platforms with LLP64 data model (e.g. Windows 64-bit), `long` has 4 bytes. Use `kmp_int64` (a typedef for `long long`) to fix alignment and size of `kmp_critical_name` on such platforms (to match the comments).

### DIFF
--- a/runtime/include/komp.h
+++ b/runtime/include/komp.h
@@ -190,9 +190,9 @@ typedef void ident_t;
 typedef int kmp_int32;
 typedef long long kmp_int64;
 #if __WORDSIZE == 32
-typedef int kmp_critical_name[8];  /* must be 32 bytes */
+typedef kmp_int32 kmp_critical_name[8];  /* must be 32 bytes */
 #else
-typedef long kmp_critical_name[4]; /* must be 32 bytes and  8-byte aligned */
+typedef kmp_int64 kmp_critical_name[4];  /* must be 32 bytes and 8-byte aligned */
 #endif
 
 extern kmp_int32 __kmpc_global_thread_num(ident_t *);


pawosm-arm (approved):

runtime/include/komp.h
```diff
@@ -190,9 +190,9 @@ typedef void ident_t;
 typedef int kmp_int32;
 typedef long long kmp_int64;
 #if __WORDSIZE == 32
-typedef int kmp_critical_name[8];  /* must be 32 bytes */
+typedef kmp_int32 kmp_critical_name[8];  /* must be 32 bytes */
 #else
-typedef long kmp_critical_name[4]; /* must be 32 bytes and  8-byte aligned */
+typedef kmp_int64 kmp_critical_name[4];  /* must be 32 bytes and 8-byte aligned */
```
bryanpkc, line 195: LLVM's OpenMP implementation has never required 8-byte alignment of this type. In `openmp/runtime/src/kmp.h`, the type is defined unconditionally as:
```c
typedef kmp_int32 kmp_critical_name[8];
```
Do we need the alignment here for some reason?

mmuetzel: To be honest, I don't know if that alignment is required here. I just modified the code to match the existing comments...
IIUC, the current implementation leads to a type that is to small on Windows 64-bit. If the alignment doesn't matter, using the `typedef` you showed unconditionally is also a possible fix.

bryanpkc: #1280 shows that it doesn't work for some reason I don't yet understand.

bryanpkc (approved):

shivaramaarao (approved):

PR #1275, opened 2022-07-14
runtime: fix building of alarm3f.c on Linux systems with modern glibc

pawosm-arm:
Popular UNIX systems (e.g. Linux, MacOS) have alarm() function
declared in unistd.h, there is no need to provide a separate forward
declaration, which could go easily out of date and create a conflict
at compile time.

### DIFF
--- a/runtime/flang/alarm3f.c
+++ b/runtime/flang/alarm3f.c
@@ -11,12 +11,13 @@
 
 #ifndef WINNT
 #include <signal.h>
+#include <unistd.h>
 #include "ent3f.h"
 
 /*
 extern void (*signal(int, void (*)(int)))(int);
-*/
 extern int alarm();
+*/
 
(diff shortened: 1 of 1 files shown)


shivaramaarao (approved):

runtime/flang/alarm3f.c
```diff
@@ -11,12 +11,13 @@
...
 #ifndef WINNT
 #include <signal.h>
+#include <unistd.h>
 #include "ent3f.h"
 
 /*
 extern void (*signal(int, void (*)(int)))(int);
-*/
 extern int alarm();
+*/
```
bryanpkc, line 20: Why did you decide to keep these lines commented out?

pawosm-arm: A habit of following the encountered pattern (namely, I just followed what original PGI code started)

PR #1276, opened 2022-07-15
tests: The IEEE mode is now being used by default

pawosm-arm:
To make this pass, https://github.com/flang-compiler/classic-flang-llvm-project/pull/122 (and similar for other LLVM release branches) needs to be merged first.

### DIFF
--- a/test/driver/ieee_mode.f90
+++ b/test/driver/ieee_mode.f90
@@ -4,7 +4,10 @@
 ! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 !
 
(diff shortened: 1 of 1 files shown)


shivaramaarao (approved):

bryanpkc (approved):

bryanpkc:
NIT: Should this test check that the `-ieee` option is passed to `flang1` correctly too?

pawosm-arm:
> NIT: Should this test check that the `-ieee` option is passed to `flang1` correctly too?

Yes. This test case must have come from the time when only flang2 was receiving this flag.

PR #1277, opened 2022-07-16
tests: fix test case failing after attempted use of LLVM15

pawosm-arm:
An early attempt of using LLVM15 encountered just one failure


bryanpkc (approved):
LGTM

PR #1279, opened 2022-07-22
Fix for regression introduced by PR #1101.

qiaozhang-hnc:
The root cause is that when an empty derived type parameter is declared in module,
an ACL structure for that parameter is generated, but its subc node is empty and
exported to the MOD file. But when importing the MOD file, flang cannot get the
information of this empty subc node.

This commit adds a flag to the MOD file to indicate whether the ACL structure
for the derived type parameter has a subc node.

### DIFF
--- a/test/f90_correct/inc/empty_type_02.mk
+++ b/test/f90_correct/inc/empty_type_02.mk
@@ -1,0 +1,19 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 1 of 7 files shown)


pawosm-arm (approved):
Thanks for prompt reaction!

bryanpkc (approved):

kiranchandramohan:
Is this changing the module format? If so, we should change the version number.
Also there should be a test to check the module generated.

pawosm-arm (changes requested):
please update `IVSN` and add a note in the `interf.h` comment.

pawosm-arm (approved):

pawosm-arm:
@kiranchandramohan are you happy with current shape?

PR #1280, opened 2022-07-27
[runtime] sync up kmp_critical_name definition with openmp

bryanpkc:
OpenMP unconditionally typedefs `kmp_critical_name` as `int[8]`, without alignment requirement. This commit syncs up our definition with it, and also removes the dependence on `__WORDSIZE`.


bryanpkc:
AArch64 test is failing with this change; I guess there was a good reason for requiring 8-byte alignment on `kmp_critical_name`.

PR #1282, opened 2022-07-30
[NFC] Add -fopenmp to test cases which use openmp runtime routines

PeixinQiao:
Previously, classic flang provides ompstub for cases using openmp
runtime routines, but not adding the option "-fopenmp". gfortran/ifort
report errors of "undefined reference to ...".

This adds the option "-fopenmp" for the two test cases, which is
prepared for removing the "-lompstub" in the driver. After that, the two
test cases compiled without the option "-fopenmp" will report the error.

### DIFF
--- a/test/f90_correct/inc/rio2.mk
+++ b/test/f90_correct/inc/rio2.mk
@@ -5,7 +5,7 @@
 #
 
 build: $(SRC)/rio2.f90
-	-$(FC) $(FFLAGS) $(LDFLAGS) $(SRC)/rio2.f90 -o rio2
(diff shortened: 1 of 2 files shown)


PeixinQiao:
This is preparation for https://github.com/flang-compiler/classic-flang-llvm-project/pull/128.

pawosm-arm (approved):

bryanpkc (approved):
LGTM

bryanpkc:
@shivaramaarao ping?

bryanpkc:
This PR is simple enough that I think two reviews should suffice. Let's merge this to unblock the classic-flang-llvm-project PRs.

PR #1284, opened 2022-08-23
win: use long double instead of __float128 on Windows platform

tdusnoki:
128 bit float is not supported on Windows platform and it causes build error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,6 +303,8 @@ if (LLVM_COMPILER_IS_GCC_COMPATIBLE)
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-error=unused-but-set-variable")
      endif()
   endif ()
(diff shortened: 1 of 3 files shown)


isuruf (commented):
`long double` is not 128 bits in size on windows? Does it make sense to do `typedef long double float128_t`?

kaadam:
> `long double` is not 128 bits in size on windows? Does it make sense to do `typedef long double float128_t`?

By default long double is 8 byte on Windows. It seems it needs to use '-mlong-double-128' compiler option to force long double to be 128 bits.

kaadam:
@bryanpkc @kiranchandramohan gentle ping?

bryanpkc (approved):
LGTM.

bryanpkc:
@isuruf Do you approve this PR now? Was your comment sufficient addressed?

isuruf (approved):

PR #1286, opened 2022-09-13
tests: the fp_reduce_intrinsic.f90 test case was using non-existing features

pawosm-arm:
In order to make this test pass, this patch modifies it towards being more generic and less specific.

### DIFF
--- a/test/llvm_ir_correct/fp_reduce_intrinsic.f90
+++ b/test/llvm_ir_correct/fp_reduce_intrinsic.f90
@@ -2,28 +2,31 @@
 ! See https://llvm.org/LICENSE.txt for license information.
 ! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-! RUN: %flang -O3                             -emit-flang-llvm %s -o - | FileCheck %s  --check-prefixes=CHECK
-! RUN: %flang -Ofast                          -emit-flang-llvm %s -o - | FileCheck %s  --check-prefixes=FAST
-! RUN: %flang -O3 -Mx,216,0x8                 -emit-flang-llvm %s -o - | FileCheck %s  --check-prefixes=NSZ
-! RUN: %flang -O3 -Mx,216,0x10                -emit-flang-llvm %s -o - | FileCheck %s  --check-prefixes=REASSOC
-! RUN: %flang -O3 -Mx,216,0x8 -Mx,216,0x10    -emit-flang-llvm %s -o - | FileCheck %s  --check-prefixes=NSZ_REASSOC
-! RUN: %flang -Ofast -Mx,216,0x8 -Mx,216,0x10 -emit-flang-llvm %s -o - | FileCheck %s  --check-prefixes=FAST
+! REQUIRES: non-ms-sdk
+! REQUIRES: llvm-13
+
+! RUN: %flang -O3                             -emit-flang-llvm %s -o /dev/stdout | FileCheck %s  --check-prefixes=CHECK
+! RUN: %flang -Ofast                          -emit-flang-llvm %s -o /dev/stdout | FileCheck %s  --check-prefixes=FAST
+! RUN: %flang -O3 -Mx,216,0x8                 -emit-flang-llvm %s -o /dev/stdout | FileCheck %s  --check-prefixes=NSZ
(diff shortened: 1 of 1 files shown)


pawosm-arm:
I'm merging it immediately in order to unblock our CI

test/llvm_ir_correct/fp_reduce_intrinsic.f90
```diff
@@ -2,28 +2,31 @@
...
-! FAST-NOT: call reassoc float @llvm.fmuladd.f32
-! NSZ: call nsz float @llvm.fmuladd.f32
-! NSZ-NOT: call fast float @llvm.fmuladd.f32
-! NSZ-NOT: call reassoc {{.*}}float @llvm.fmuladd.f32
-! REASSOC-NOT: call fast float @llvm.fmuladd.f32
-! REASSOC-NOT: call nsz {{.*}}float @llvm.fmuladd.f32
-! REASSOC: call reassoc float @llvm.fmuladd.f32
-! NSZ_REASSOC: call nsz reassoc float @llvm.fmuladd.f32
-! NSZ_REASSOC-NOT: call fast float @llvm.fmuladd.f32
+! CHECK-NOT: {{[[:space:]]}}fast
```
bryanpkc, line 19: What problem did this fix? Can you describe the situation that would cause this test to fail incorrectly?

You might want to make sure that the word "fast" terminates too, e.g.
```
! CHECK-NOT: {{[[:space:]]fast[[:space:]]}}
```

pawosm-arm: those command lines have "fp-contract"="fast", it's not the 'fast' we don't want to see here. Looking for a space before 'fast' is minimalistic yet sufficient.

bryanpkc: Actually my question was, what was the problem with `CHECK-NOT: call fast float @llvm.fmuladd.f32`?

test/llvm_ir_correct/fp_reduce_intrinsic.f90
```diff
@@ -2,28 +2,31 @@
...
-! RUN: %flang -O3                             -emit-flang-llvm %s -o - | FileCheck %s  --check-prefixes=CHECK
-! RUN: %flang -Ofast                          -emit-flang-llvm %s -o - | FileCheck %s  --check-prefixes=FAST
-! RUN: %flang -O3 -Mx,216,0x8                 -emit-flang-llvm %s -o - | FileCheck %s  --check-prefixes=NSZ
-! RUN: %flang -O3 -Mx,216,0x10                -emit-flang-llvm %s -o - | FileCheck %s  --check-prefixes=REASSOC
-! RUN: %flang -O3 -Mx,216,0x8 -Mx,216,0x10    -emit-flang-llvm %s -o - | FileCheck %s  --check-prefixes=NSZ_REASSOC
-! RUN: %flang -Ofast -Mx,216,0x8 -Mx,216,0x10 -emit-flang-llvm %s -o - | FileCheck %s  --check-prefixes=FAST
+! REQUIRES: non-ms-sdk
+! REQUIRES: llvm-13
+
+! RUN: %flang -O3                             -emit-flang-llvm %s -o /dev/stdout | FileCheck %s  --check-prefixes=CHECK
```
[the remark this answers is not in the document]
pawosm-arm: That's why I added non-ms-sdk as a requirement. AFAIK, we don't support any other systems that don't have /dev/stdout.

There is a major difference between `-emit-llvm` and `-emit-flang-llvm`, mostly in who is emitting the output. Sadly, when passing '-' to flang2, many problems occur that were not currently addressed.

test/llvm_ir_correct/fp_reduce_intrinsic.f90
pawosm-arm: now I can see that this test file is entirely redundant, it tests the same things as fp_reduce.f90, while it was supposed to test attributes of the calls to the LLVM intrinsics, these are not emitted by the upstream compiler.

test/llvm_ir_correct/fp_reduce_intrinsic.f90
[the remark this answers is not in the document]
bryanpkc: Right, I remember now. `-emit-flang-llvm` doens't work with `-o -`.

PR #1289, opened 2022-09-18
[LLVM 15][runtime] Define overloaded function types and improve type checking

bryanpkc:
Classic Flang code uses tables of functions (with different argument types) to achieve polymorphism. Previously, such a function was declared as `void (*p)()`, to allow any function to be assigned to the function pointer variable `p`. But when a call to such a function pointer is attempted, Clang 15 now emits warnings like these:
```
.../flang/runtime/flang/comm.c:97:14: error: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]
    sk->start(sk->arg, rb, sb, rd, sd);
             ^
.../flang/runtime/flang/comm.c:117:13: error: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]
    sk->free(sk->arg);
            ^
```
To work around these issues, this PR replaces polymorphic pointer parameters in such functions with `void *` parameters, and casts the `void` pointers in the function body to the intended types of the respective arguments. By creating `typedef`s of the fully prototyped function pointers, some degree of type checking is possible without too much overhead. Relaxing the parameter type to `void *` also eliminates `-Wincompatible-function-pointer-types` warnings when assigning functions to function pointers.

The changes are organized into multiple relatively independent commits for easier review. If necessary, I can break this up into multiple pull requests.

### DIFF
--- a/runtime/flang/comm.c
+++ b/runtime/flang/comm.c
@@ -61,12 +61,11 @@ sked *I8(__fort_comm_sked)(chdr *ch, char *rb, char *sb, dtype kind, int len)
   return &sk->sked;
 }
 
-void *ENTFTN(COMM_START, comm_start)(sked **skp, void *rb, F90_Desc *rd,
+void *ENTFTN(COMM_START, comm_start)(void **skp, void *rb, F90_Desc *rd,
                                      void *sb, F90_Desc *sd)
 {
-  sked *sk;
+  sked *sk = (sked *)*skp;
 
-  sk = *skp;
   if (sk != NULL) {
 #if defined(DEBUG)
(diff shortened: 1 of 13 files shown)


pawosm-arm (approved):
No need to split it into more PRs, it's better to keep this change self-contained

bryanpkc:
@pawosm-arm  Thanks for the quick review. I have force-pushed the branch, reworded commit messages, and added one more commit. Would you like to take another look?

shivaramaarao (approved):

PR #1290, opened 2022-09-18
[LLVM 15][runtime] Stop using K&R function declarations

bryanpkc:
Legacy Classic Flang code uses K&R C function declarations, which differ from ANSI C function declarations in that the parameters are not part of the function prototype, which means that the functions can have unknown numbers of parameters, of unknown types. With Clang 15, they will now cause a warning message, e.g.
```
.../flang/runtime/flang/curdir.c:31:6: error: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]
void __fort_fixmnt(new, old)
     ^
```
This commit rewrites all function declarations in modern C.

### DIFF
--- a/runtime/flang/backspace.c
+++ b/runtime/flang/backspace.c
@@ -127,9 +127,7 @@ _f90io_backspace(__INT_T *unit, __INT_T *bitv, __INT_T *iostat, int swap_bytes)
 }
 
 __INT_T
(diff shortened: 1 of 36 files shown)


pawosm-arm (approved):
This is it, an end of an era of the old C.

shivaramaarao (approved):

PR #1291, opened 2022-09-18
[LLVM 15][runtime] Remove extern function declarations

bryanpkc:
`extern` function declarations have been replaced by headers (standard headers if possible), except for single unique uses. This not only reduces code duplication, but also fixes `-Wdeprecated-non-prototype` warnings, e.g.
```
.../flang/runtime/flang/curdir.c:25:14: error: a function declaration without a prototype is deprecated in all versions of C and is treated as a zero-parameter prototype in C2x, conflicting with a previous declaration [-Werror,-Wdeprecated-non-prototype]
extern char *getcwd();
             ^
/usr/include/unistd.h:514:14: note: conflicting prototype is here extern char *getcwd (char *__buf, size_t __size) __THROW __wur;
             ^
```
Some other dead code is also removed.

### DIFF
--- a/runtime/flang/curdir.c
+++ b/runtime/flang/curdir.c
@@ -22,8 +22,6 @@
 #define getcwd _getcwd
 #endif
 
-extern char *getcwd();
-
 WIN_MSVCRT_IMP char *WIN_CDECL getenv(const char *);
 
 /* fix pathname for "funny" NFS mount points */
--- a/runtime/flang/access3f.c
+++ b/runtime/flang/access3f.c
@@ -16,9 +16,7 @@
 
 #include "io3f.h"
 #include "ent3f.h"
-
-extern char *__fstr2cstr();
-extern void __cstr_free();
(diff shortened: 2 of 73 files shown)


runtime/flang/curdir.c
```diff
@@ -22,8 +22,6 @@
 #define getcwd _getcwd
```
pawosm-arm, line 22: the function `__fort_getdir()` below for which this `getcwd` was defined doesn't seem to be used anywhere in flang and its runtime library. It's possible that there is more to delete in this file

pawosm-arm (approved):
looks ok, see my comment on curdir.c, possibly it can be cleaned more, but it may go slightly out of scope of this commit.

bryanpkc:
That's a good observation. It seems that the entire `curdir.c` could be removed. Perhaps we can submit a separate PR for that.

shivaramaarao (approved):

PR #1294, opened 2022-09-18
[LLVM 15][libpgmath] Decouple from libflangrti by using errno directly; NFCI

bryanpkc:
libpgmath should be able to use `errno` directly, without having to go through `libflangrti`. Decoupling the libraries not only allows libpgmath to be used independently, but also fixes compiler warnings about undeclared functions.


pawosm-arm (approved):
Are there any other dependencies like that?

shivaramaarao (approved):

bryanpkc:
@pawosm-arm No, there isn't any other. The calls to `__io_errno` and `__io_set_errno` were the only references that caused the compiler warnings. `libpgmath.so` does not have any link-time dependence on `libflangrti.so`.

PR #1297, opened 2022-09-22
[LLVM 15][runtime] Fix compilation warning in {ftni64bitsup,utilsi64}.c; NFCI

bryanpkc:
ftni64bitsup.c and utilsi64.c define a static function `shf64` (a helper for shifting signed 64-bit values) identically, while only ftni64bitsup.c defines the unsigned verison `ushf64`. Both files call `shf64` on various unsigned values, causing the compiler to complain. They also declare conflicting prototypes for the same functions, declare some functions as `extern` unnecessarily, and omit forward declarations when they are necessary. This patch fixes all these problems, and makes both files use the same definitions for `shf64` and `ushf64`, reducing code duplication.

### DIFF
--- a/runtime/flang/ftni64bitsup.c
+++ b/runtime/flang/ftni64bitsup.c
@@ -11,8 +11,9 @@
 #include "dattype.h"
 #include "ftni64.h"
 
-static void shf64();
-static void ushf64();
+/* forward declarations */
+void shf64(DBLINT64, int, DBLINT64);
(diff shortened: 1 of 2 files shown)


pawosm-arm (approved):

bryanpkc:
@shivaramaarao gentle ping

shivaramaarao (approved):

PR #1298, opened 2022-09-22
[LLVM 15][runtime] Replace LINUX8664 with TARGET_X8664/TARGET_LINUX_X8664

bryanpkc:
Various files under `runtime/flang` use the macro `LINUX8664` to guard X86-only code; but this is incorrect because `LINUX8664` is never defined by any CMakeLists.txt (it is only defined when building `runtime/libpgmath`). Uses of the macro are modified to use `TARGET_X8664` and/or `TARGET_LINUX_X8664`.

This patch also guards some Windows-specific code with the `WIN64` macro to avoid compilation warnings.


runtime/flang/ftni64.h
```diff
@@ -16,7 +16,7 @@
 
 #define __HAVE_LONGLONG_T
 
-#if defined(LINUX8664) || defined(OSX8664) || defined(TARGET_LLVM_ARM64)
```
pawosm-arm, line 19: This patch not only replaces but also removes some of the uses of LINUX8664. was that intentional? If so, it needs to be mentioned in the commit message...

bryanpkc: Yes, it's intentional. I have updated the commit message. Is it better?

pawosm-arm (approved):

bryanpkc:
@shivaramaarao gentle ping

shivaramaarao (approved):

PR #1308, opened 2022-10-19
[LLVM 15] Make check-all invoke check-flang again

bryanpkc:
D121838 introduced a new macro for adding lit test suites to an umbrella test suite (e.g. "check-all"). It updated clang/CMakeLists.txt to change how "check-clang" is added to "check-all". This patch applies a similar transformation to Classic Flang's CMakeLists.txt, but retains the original code path so that the code still works for release_14x and older branches. Without this patch, when building Classic Flang with LLVM 15, running "make check-all" would have no effect (it does not invoke "make check-flang").

### DIFF
--- a/.github/workflows/build_flang.yml
+++ b/.github/workflows/build_flang.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Build and install flang & libpgmath
         run: |
           ${{ env.install_prefix }}/bin/clang --version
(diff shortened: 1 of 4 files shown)


pawosm-arm (approved):

bryanpkc:
@pawosm-arm  I added a CI-only patch to this PR; hope that's okay with you.

pawosm-arm:
> @pawosm-arm I added a CI-only patch to this PR; hope that's okay with you.

Yup, my approval still holds.

PR #1311, opened 2022-10-24
[LLVM 15][libpgmath] Fix -Wimplicit-function-declaration

kenjiang67:
Some functions in fltfenv.c were not being properly declared, causing a -Wimplicit-function-declaration warning, which is now an error. This commit updates the CMakeLists file to fix this issue.


bryanpkc (approved):
LGTM. The CI job shows that the `-Wimplicit-function-declaration` warnings have indeed disappeared with this fix.

pawosm-arm:
We were adding `-D_GNU_SOURCE` as extra CFLAGS passed to CMake for a very long time. It saves the problem with missing declarations, but creates tons of warnings on `_GNU_SOURCE` re-declarations, which we just simply ignore. Did you happen to observe those warnings after applying this patch?

bryanpkc:
@pawosm-arm The AArch64 CI job doesn't seem to show any problem like that, but in the X86 CI job, there are warnings of this form, even before this PR:
```
/home/runner/work/flang/flang/runtime/libpgmath/lib/x86_64/math_tables/mth_acosdefs.h:86:38: warning: incompatible redeclaration of library function 'cacosf' [-Wincompatible-library-redeclaration]
```

pawosm-arm (approved):
Ok, this seems to be straight to the point fix.

bryanpkc:
@shivaramaarao  Does this look okay to you?

PR #1313, opened 2022-10-27
[workflows] Build and test master branch with release_15x only

bryanpkc:
As discussed at the 2022/11/02 call, going forward the `master` branch will only
support opaque pointer IR, while the `legacy` branch will continue to support
legacy IR with older classic-flang-llvm-project release branches.


bryanpkc:
The CI jobs look okay to me.

pawosm-arm (approved):
Radical step. I'm ok with that, but I'd rather see the others to have their say too.

shivaramaarao (approved):

PR #1316, opened 2022-10-31
[workflows] Add build script options to allow customization

bryanpkc:
This patch adds the `-b` option to `build-flang.sh` which lets the user specify an alternate location for the build directories of libpgmath and flang (both will be placed under the same build prefix), and the `-x` option which lets the user specify additional CMake options when configuring flang.

### DIFF
--- a/build-flang.sh
+++ b/build-flang.sh
@@ -6,10 +6,12 @@ OPTIND=1         # Reset in case getopts has been used previously in the shell.
 # Initialize our own variables:
 TARGET="X86"
 BUILD_TYPE="Release"
+BUILD_PREFIX="./build"
 INSTALL_PREFIX="/usr/local"
 NPROC=1
 USE_CCACHE="0"
 USE_SUDO="0"
+EXTRA_CMAKE_OPTS=""
 VERBOSE=""
 
 set -e # Exit the script on first error.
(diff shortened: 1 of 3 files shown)


build-flang.sh
```diff
@@ -60,13 +66,22 @@ if [ $USE_CCACHE == "1" ]; then
...
 fi
 
-# Build and install libpgmath
-cd runtime/libpgmath
-mkdir -p build && cd build
+if [ -n "$EXTRA_CMAKE_OPTS" ]; then
+  CMAKE_OPTIONS="$CMAKE_OPTIONS $EXTRA_CMAKE_OPTS"
+fi
+
+# Assume the script is run where it is located.
```
pawosm-arm, line 73: Not sure about this assumption. Also this patch removes `cd` instructions. How this script should be executed? Was there a valid usage scenario agreed?

bryanpkc: The `cd` command on line 64 in the original version of the script had made the assumption; I just made it more explicit. The script always had to be run as `./build-flang.sh`, in the project root.

It's possible to make the script PWD-independent, but that is not in scope for this PR.

pawosm-arm (approved):

shivaramaarao (approved):

PR #1326, opened 2022-11-11
win: Disable QUADFP build in libpgmath for Windows on ARM

kaadam:
Basically QUADFP is disabled in Flang on Windows on ARM, so this setting should be applied in libpgmath as well until we make sure about it supported corretly.

### DIFF
--- a/runtime/libpgmath/CMakeLists.txt
+++ b/runtime/libpgmath/CMakeLists.txt
@@ -63,7 +63,7 @@ if(${LIBPGMATH_SYSTEM_PROCESSOR} MATCHES "aarch64" AND NOT ${LIBPGMATH_WITH_GENE
   else()
     message(STATUS "Libpgmath CPU target explicitly set to ${LLVM_FLANG_CPU_TARGET}")
   endif()
-  if(FLANG_ENABLE_QUADFP)
+  if(FLANG_ENABLE_QUADFP AND NOT ("${LIBPGMATH_SYSTEM_NAME}" STREQUAL "Windows"))
     add_definitions(-DTARGET_SUPPORTS_QUADFP)
     set(TARGET_SUPPORTS_QUADFP True)
   endif()


runtime/libpgmath/CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ if(${LIBPGMATH_SYSTEM_PROCESSOR} MATCHES "aarch64" AND NOT ${LIBPGMATH_WITH_GENE
   else()
     message(STATUS "Libpgmath CPU target explicitly set to ${LLVM_FLANG_CPU_TARGET}")
   endif()
-  if(FLANG_ENABLE_QUADFP)
+  if(FLANG_ENABLE_QUADFP AND NOT ${TARGET_OS} STREQUAL "Windows")
```
pawosm-arm, line 66: Arm test cases are now failing in precommit. Could it be that this logic has caused that? Maybe it should be put into parentheses, e.g. `((FLANG_ENABLE_QUADFP) AND NOT (${TARGET_OS} STREQUAL "Windows"))`

kaadam: Thanks for your comment. Hope I will work for now. That was my fault, I should have used LIBPGMATH_SYSTEM_PROCESSOR instead of TARGET_OS.

pawosm-arm: NB why this variable is named `LIBPGMATH_SYSTEM_PROCESSOR`? It's so confusing, if it summarizes CPU and OS, it should be named simply `LIBPGMATH_PLATFORM`.

pawosm-arm: nah, I looked into other parts of pgmath and found this:
```
elseif(${LIBPGMATH_SYSTEM_NAME} MATCHES "Windows")
```
I guess you should have used `(${LIBPGMATH_SYSTEM_NAME} MATCHES "Windows")` instead...

pawosm-arm (approved):

bryanpkc (approved):
LGTM

PR #1332, opened 2022-11-23
[workflow] Fix push workflow in build_flang_arm64.yml (legacy branch)

bryanpkc:
`build_flang_arm64.yml` has custom code for cloning the repository, since the workflow runs in a container on a hosted runner. The workflow needs to be given an explicit branch name to check out, otherwise it would always checkout the `master` branch. This patch adds back the missing branch name to fix a push workflow issue on the `legacy` branch, and also shuffles the steps slightly (NFC) to make `build_flang.yml` and `build_flang_arm64.yml` more similar.

### DIFF
--- a/.github/workflows/build_flang.yml
+++ b/.github/workflows/build_flang.yml
@@ -39,6 +39,10 @@ jobs:
           cmake --version
           make --version
 
+      # gcc11 and g++11 are needed to install llvm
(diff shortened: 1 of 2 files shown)


pawosm-arm (approved):
Not much of my experience in the area, but the changed lines themselves look reasonable.

bryanpkc:
Merging CI-only PR with one approval.

PR #1335, opened 2022-12-01
win: Fixed missing prototype error reported by Clang 15

kaadam:
error: passing arguments to 'GetVolumeInformationA' without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]

Also removed the superfluous x86 case.
Added proper Windows types for 'GetVolumeInformationA' function.

### DIFF
--- a/runtime/flang/getvolinfo3f.c
+++ b/runtime/flang/getvolinfo3f.c
@@ -8,27 +8,23 @@
 /*
  *	getvolinfo3f.c - DFWIN GetVolumeInformation subprogram
  *	This is just the interface to the actual system routine from
- *	the dwfin module routine.  This routine is called as an 'alias'
+ *	the dfwin module routine.  This routine is called as an 'alias'
  *	when the module is called; this routine makes sure that the
  *	output character arguments are blank-filled.
  */
 
 #include <string.h>
 
-typedef char *LPSTR;
-typedef int DWORD;
+#if defined (_WIN32)
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
 
-#if defined(_WIN64)
-typedef long long LDWORD;
-extern int GetVolumeInformationA();
 #define ENTNAM(ss) _##ss
-
-typedef LDWORD *LPDWORD;
 
 static void fill(LPSTR, int, int);
 
 extern int ENTNAM(pgdfw_GetVolumeInformation)(
-    LPSTR RootPathName, LPSTR VolumeNameBuffer, DWORD VolumeNameSize,
+    LPCSTR RootPathName, LPSTR VolumeNameBuffer, DWORD VolumeNameSize,
     LPDWORD VolumeSerialNumber, LPDWORD MaximumComponentLength,
     LPDWORD FileSystemFlags, LPSTR FileSystemNameBuffer,
     DWORD FileSystemNameSize)


runtime/flang/getvolinfo3f.c
```diff
@@ -16,26 +16,19 @@
...
+typedef const char *LPCSTR;
 typedef int DWORD;
+typedef DWORD *LPDWORD;
 
-#if defined(WIN64) || defined(WIN32)
-#if defined(WIN64)
-typedef long long LDWORD;
-extern int GetVolumeInformationA();
+#if defined(_WIN64)
+extern int GetVolumeInformationA(LPCSTR, LPSTR, DWORD, LPDWORD, LPDWORD,
```
bryanpkc, line 24: Is it not a better idea to `#include <fileapi.h>`?

kaadam: Hi Bryan, sorry for the delayed response, I've bumped into another issue with tip of tree Flang on Windows after the rebase.
Unfortunately  we cannot include <fileapi.h> directly. Basically 'fileapi' is automatically included in 'windows.h' and its dependencies as well. So this header is not a self-contained header because of its dependencies.

In this case we can only use 'windows.h' for example with WIN32_LEAN_AND_MEAN definition in order to exclude a bunch of unnecessary Windows header files.

runtime/flang/getvolinfo3f.c
```diff
@@ -8,27 +8,25 @@
...
  */
 
 #include <string.h>
 
-typedef char *LPSTR;
-typedef int DWORD;
+#if defined (_WIN64)
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
```
bryanpkc, line 21: Would `#define WIN32_LEAN_AND_MEAN` cause a warning or error if the macro is already defined? Is the `#ifndef` necessary?

kaadam: Yes, that could be removed.

bryanpkc (approved):

bryanpkc (approved):

runtime/flang/getvolinfo3f.c
```diff
@@ -8,27 +8,23 @@
...
+ *	the dfwin module routine.  This routine is called as an 'alias'
  *	when the module is called; this routine makes sure that the
  *	output character arguments are blank-filled.
  */
 
 #include <string.h>
 
-typedef char *LPSTR;
-typedef int DWORD;
+#if defined (_WIN64)
```
Neumann-A, line 18:
```suggestion
#if defined (_WIN32)
```
?

kaadam: Currently we only support Windows x64 and ARM64. That's why I used _WIN64 predefined macro instead of _WIN32.  _WIN32 defined as 1, when the compilation target is 32bit as well.

Neumann-A: even if only x64 win is supported shouldn't the constrain here not be as lose as possible? 
You want to know if the target is windows not if the target is x86 or x64. Basically I would use _WIN32 always if I want to know if something is Windows while I would use _WIN64 (or even better _M_X64) only if I need to differentiate between x64 and x86.

(Not that I care about x86)

kaadam: Yes, I can accept it :) I will update this.

shivaramaarao (approved):

bryanpkc:
Merging with two approvals, assuming @Neumann-A is okay.

Issue #1334, opened 2022-12-01
Forcing long double to be 128bit fails with assertion on newer VS 2019 version


kaadam:
It seems when we force long double to be 128bit on Windows (#1284 ), it causes an assertion in newer Windows system headers[1] (VS 2019 14.29.30133). By default double and long double is equal to each other, both are 64bit on Windows. FLOAT128 is not supported.

`C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\xlocnum(618,9): error: static assertion failed due to requirement 'sizeof(double) == sizeof(long double)': Bad assumption: sizeof(double) == sizeof(long double).
`

Neumann-A:
Is there a solution/workaround for this issue?

kaadam:
> Is there a solution/workaround for this issue?

Hi Alexander,
Well, the workaround is currently to remove this compiler option which fixes the build on Windows. Of course with that limitation long double will be 64 bits until we figure out the final solution to support float128 on Windows.

bryanpkc:
As long as we don't enable quad-precision FP support on Windows, the problematic code (which assumes that `float128_t` is 128 bits in size) will in theory not be used, and hence cause no problem.

bryanpkc:
Fixed in #1340.

PR #1340, opened 2023-01-20
win: Remove '-mlong-double-128' compiler option

kaadam:
On Windows 'long double' type size is 64 bits, same as double. As a workaround we started to use clang's '-mlong-double-128' option to force 'long double' to be 128 bits.
With newer version of Windows SDK it causes an assertion in one of system header.
Currently it breaks the Windows build, so it needs to remove.

Issue: #1334


bryanpkc (approved):

runtime/include/float128.h
```diff
@@ -12,7 +12,16 @@
 typedef __float128 float128_t;
 #else
 /* __float128 is not available on AArch64 or other generic targets;
-   on AArch64 at least, long double is 128 bits in size. */
+ * on AArch64 at least, long double is 128 bits in size.
+ *
+ * Fixme: On Windows long double size is 64 bits, same as double is.
```
bryanpkc, line 17: Could you use uppercase for the `FIXME`? Makes it easier to grep for.

kaadam: Sure, updating that.

bryanpkc:
@pawosm-arm @shivaramaarao Please review this PR soon as it fixes a build break on Windows.

pawosm-arm (approved):

PR #1341, opened 2023-01-20
[win][scutil] Add support Windows Complex types for pow folding

kaadam:
On Windows 'complex' or '_Compex' keyword doesn't supported by clang-cl.
Microsoft uses structure types for representing Complex number.
Since complex numbers aren't native type, the standard arithmetic
operators aren't defined on complex types.

Implicit conversions aren't defined between complex types.

### DIFF
--- a/include/fp-folding.h
+++ b/include/fp-folding.h
@@ -57,6 +57,39 @@ typedef float float32_t;
 typedef double float64_t;
 typedef long double float128_t; /* 128 bits in memory, format host-dependent */
 
+#if !defined(_WIN32)
+# define FLOAT_COMPLEX_TYPE float complex
+# define FLOAT_COMPLEX_CREATE(real, imag) (real + imag * I)
+# define DOUBLE_COMPLEX_TYPE double complex
+# define DOUBLE_COMPLEX_CREATE(real, imag) (real + imag * I)
+# define LONG_DOUBLE_COMPLEX_TYPE long double complex
+# define LONG_DOUBLE_COMPLEX_CREATE(real, imag) (real + imag * I)
+
+/* For type conversion */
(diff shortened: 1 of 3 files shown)


lib/scutil/host-fp-folding.c
```diff
@@ -26,6 +26,7 @@
 #include <string.h>
 #if defined(_WIN64) && !defined(_M_ARM64)
 #include <mmintrin.h>
+//#include <intrin.h>
```
bryanpkc, line 29: Remove this if it isn't needed.

bryanpkc (approved):

include/fp-folding.h
```diff
@@ -57,6 +57,39 @@ typedef float float32_t;
 typedef double float64_t;
 typedef long double float128_t; /* 128 bits in memory, format host-dependent */
 
+#if !defined(_WIN64)
```
Neumann-A, line 60:
```suggestion
#if !defined(_MSC_VER)
```
or `_WIN32`. don't know about mingw. I would assume since it is gcc it has a complex type.

bryanpkc: @Neumann-A Does Adam's latest change look good to you?

bryanpkc: @Neumann-A If this looks okay to you, we will merge this PR.

Neumann-A: Looks ok to me but as i said i dont know about mingw.

pawosm-arm (approved):
Considering it only introduces wrapper macros, I tend to approve it.

include/fp-folding.h
bryanpkc: Thanks. We'll merge this.

PR #1346, opened 2023-01-26
win: Fix the build with verbose options

kaadam:
The 'append' method should be used instead of extend, since the 'extend' method treats the arguments as an iterable object, so extends the current list with list of characters.


bryanpkc (approved):

bryanpkc:
A similar CI-related change has been reviewed and merged in classic-flang-llvm-project. Merging with one approval.

PR #1350, opened 2023-02-05
[QuadFP][flang1] Semantic analysis for type conversion

bryanpkc:
Update `cngtyp2` in semutil.c to handle type conversion to and from quad-precision complex numbers. Add a number of tests to check that type conversion involving quad-precision complex numbers are correct.

PR #1351, opened 2023-02-05
[QuadFP][flang1] Semantic check for operand type promotion

bryanpkc:
Update `chkopnds` in semutil.c to promote operand types in various arithmetic operations involving quad-precision complex numbers. Add tests to check that such arithmetic operations work correctly.

This also adds two tests for exponentiation of single- and double-precision complex numbers, for completeness sake.

This PR is part of a series, and follows #1350. Reviewers, please review only the second commit on this branch.


tools/flang1/flang1exe/semutil.c
```diff
@@ -622,6 +634,8 @@ cngtyp2(SST *old, DTYPE newtyp, bool allowPolyExpr)
     case TY_LOG8:
     case TY_INT8:
       break;
+    case TY_QCMPLX:
```
pawosm-arm, line 637: similarly to the other review, shouldn't this part be guarded by an ifdef?

pawosm-arm (commented):
same question as in the other review.

Also, I'm seeing some typo corrected (missing space added), usually I would require a separate NFC commit, but I guess this change is so tightly related, I'll make exceptional approval.

pawosm-arm (approved):

shivaramaarao (approved):

bryanpkc:
The new test cases in this PR use constants that are arbitrarily chosen.

PR #1352, opened 2023-02-11
cmake: stop asking llvm-config for src-root as it is not supported since LLVM16.0

pawosm-arm:
A backward-compatible solution to a problem caused by the changes introduced in the LLVM 16.0.

I've found inability to refer to the LLVM sources particularly painful for classic-flang, and I guess this will many times be a source of heavy confusion for our end-users even with my patch. I tend to agree with an alternative solution which is to revert the patch that drops `--src-root` flag from `llvm-config` in our clone of the LLVM project.

### DIFF
--- a/test/lit.cfg
+++ b/test/lit.cfg
@@ -125,17 +125,9 @@ if config.test_exec_root is None:
     if not llvm_config:
         lit_config.fatal('No site specific configuration available!')
 
-    # Get the source and object roots.
-    llvm_src_root = lit.util.capture(['llvm-config', '--src-root']).strip()
+    # Get the object root.
     llvm_obj_root = lit.util.capture(['llvm-config', '--obj-root']).strip()
-    flang_src_root = os.path.join(llvm_src_root, "tools", "flang")
     flang_obj_root = os.path.join(llvm_obj_root, "tools", "flang")
-
-    # Validate that we got a tree which points to here, using the standard
-    # tools/flang layout.
-    this_src_root = os.path.dirname(config.test_source_root)
-    if os.path.realpath(flang_src_root) != os.path.realpath(this_src_root):
-        lit_config.fatal('No site specific configuration available!')
 
     # Check that the site specific configuration exists.
     site_cfg = os.path.join(flang_obj_root, 'test', 'lit.site.cfg')
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,8 +83,7 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
       "--bindir"
       "--libdir"
       "--includedir"
-      "--prefix"
-      "--src-root")
+      "--prefix")
     execute_process(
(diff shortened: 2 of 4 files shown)


bryanpkc:
I wonder if we should stop depending on `llvm-config`.

`LLVM_INCLUDE_DIR` and `LLVM_BINARY_DIR` can be computed from the real path to the `clang` executable, so we don't need `llvm-config` for those.

`HandleOutOfTreeLLVM.cmake` does not seem to be used anywhere, but it allows the user to override `llvm-config --src-root` with a `LLVM_PATH` variable:

https://github.com/flang-compiler/flang/blob/master/runtime/libpgmath/cmake/modules/HandleOutOfTreeLLVM.cmake#L7-L11

Asking the user to tell us where the LLVM source directory is doesn't seem unreasonable at all. Classic Flang is typically built immediately after building LLVM, so the path should be readily available, e.g. in a wrapper script.

We could further recommend/assume that Classic Flang can be cloned and built in-tree, and make the variable default to `${CMAKE_SOURCE_DIR}/../` (with sanity checks added to confirm that the default is in fact correct when the user does not specify `LLVM_PATH`).

test/lit.cfg
```diff
@@ -125,18 +125,10 @@ if config.test_exec_root is None:
     if not llvm_config:
         lit_config.fatal('No site specific configuration available!')
 
-    # Get the source and object roots.
-    llvm_src_root = lit.util.capture(['llvm-config', '--src-root']).strip()
+    # Get the object root.
     llvm_obj_root = lit.util.capture(['llvm-config', '--obj-root']).strip()
-    flang_src_root = os.path.join(llvm_src_root, "tools", "flang")
     flang_obj_root = os.path.join(llvm_obj_root, "tools", "flang")
```
bryanpkc, line 130: This code assumes an in-tree build, which I suspect is no longer supported or functional. Should we just clean this up? Or can you test the in-tree build process and see if it works?

pawosm-arm: In-tree build is problematic since monorepo has appeared. I'd rather drop this feature before it start to be a serious maintenance nightmare.

pawosm-arm:
> `runtime/libpgmath/cmake/modules/HandleOutOfTreeLLVM.cmake` doesn't seem to be used anywhere; should we just remove the file? Or move the code from `/CMakeLists.txt` into it, to avoid code duplication?

If it's not used anywhere, then it's more convincing to get rid of it

bryanpkc:
@shivaramaarao Do you have any objection to removing the CMake code that supports in-tree builds, which probably doesn't work at the moment?

shivaramaarao:
> @shivaramaarao Do you have any objection to removing the CMake code that supports in-tree builds, which probably doesn't work at the moment?

Hi Bryan. I dont have objection in removing this code.

bryanpkc (approved):

bryanpkc:
We can delete the unsupported code in a separate PR.

shivaramaarao (approved):

PR #1355, opened 2023-03-13
[flang1] Fix a regression about array constructor

PeixinQiao:
When the array constructor element is array of dynamic char, the length can only be checked in runtime, in which case we cannot use `eq_dtype` to check if the two dtypes are equal since the length could be two different asts containing the length info.

This bug should exist long time ago, and it was hidden by 0918a2f2d. It was exposed again in a545a61c9.

### DIFF
--- a/test/f90_correct/src/acl04.f90
+++ b/test/f90_correct/src/acl04.f90
@@ -1,0 +1,27 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Test for array constructor whose element is array of dynamic char.
+
+program m_test_adjust_len
+  call test_adjust_len(4, 4, 6)
+  print *, "PASS"
+
+contains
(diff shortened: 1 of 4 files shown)


pawosm-arm (approved):
Good catch!

test/f90_correct/src/acl04.f90
```diff
@@ -0,0 +1,27 @@
...
+
+contains
+  subroutine test_adjust_len(a, b, c)
+    integer :: a, b, c
+    character(len=a) :: s1(2)
+    character(len=b) :: s2
+    character(len=c) :: s3(1)
+
+    s2 = "aaaa"
+    s3 = "bbbb"
```
bryanpkc, line 20: To make the intent of the test more explicit, I suggest passing 6 to the parameter `c`, and setting `s3` to `"bbbbbb"`.

PeixinQiao: Thanks. Fixed.

PeixinQiao:
gentle ping @shivaramaarao

bryanpkc (approved):

bryanpkc:
@shivaramaarao If you have no objection, I will merge this PR.

PR #1356, opened 2023-03-22
[QuadFP][test] Make comments clear and consistent in style; NFC

bryanpkc:
Also fixed spacing problems in some test cases, and renamed qtriangle.f08 and qatriangle.f08 to qtrig.f08 and qatrig.f08, respectively.


pawosm-arm (approved):
Wow, that's a massive effort, much needed one. Note it will pop up on `git blame` frequently (fortunately, QuadFP is a self-contained subject, so it shouldn't be a problem). The way github presented some of those changes made it harder to make a judgement that it is actually NFC at first sight.

shivaramaarao (approved):

PR #1357, opened 2023-03-22
[QuadFP][runtime] Define C_LONG_DOUBLE_COMPLEX as correct type kind

bryanpkc:
When `TARGET_SUPPORTS_QUADFP` is set, `C_LONG_DOUBLE_COMPLEX` in ISO C binding should refer to quad-precision complex numbers; otherwise, it refers to double-precision complex numbers (which is the current behaviour).


shivaramaarao (approved):

pawosm-arm (approved):
a simple change, LGTM

PR #1358, opened 2023-03-22
[QuadFP][flang1] Support quad complex zero in collapse_assignment

bryanpkc:
If an array is assigned a quad-precision complex value, and the value has a zero real part and a zero imaginary part, the array assignment can be collapsed into a `memset`.


shivaramaarao (approved):

pawosm-arm (commented):
does it need a test case?

bryanpkc:
Added two test cases.

bryanpkc:
@pawosm-arm Can you take a look at this again?

pawosm-arm (approved):

PR #1359, opened 2023-03-22
[QuadFP] Support quad-complex SIN intrinsic

bryanpkc:
- Add relevant libpgmath routines.
- Add relevant ILM opcodes and Fortran runtime symbols.
- Add ILM generation and lowering logic.
- Add test cases.


pawosm-arm (commented):
I guess @bryanpkc has already summarized what's needed.

bryanpkc:
> I guess @bryanpkc has already summarized what's needed.

This patch is complete for `CQSIN`. Support for other trigonometry intrinsics will be added in subsequent PRs. The title of this PR has been updated.

bryanpkc:
> Alternatively, when I have some free time to do so, I can add the other trigonometry intrinsics to this PR, so that the complete set of functionalities are merged all at once.

Unfortunately, right now I won't have time to do this. There are quite a number of patches for other type of QuadFP intrinsics that we were going to upstream; we are discussing internally whether someone else has the time to work on them.

tools/flang1/flang1exe/symacc.h
```diff
@@ -16,7 +16,7 @@ struct SYM;
 #include "sharedefs.h"
 
 /* Keep in sync with INIT_NAMES_SIZE, in syminidf.h. */
-#define STB_NSIZE 5063
+#define STB_NSIZE 5069
```
pawosm-arm, line 19: I'm feeling lost, where is `syminidf.h`? I just wanted to make sure that this shouldn't it be guarded by `#ifdef TARGET_SUPPORTS_QUADFP`

tools/shared/utils/symacc.h
```diff
@@ -17,7 +17,7 @@ struct SYM;
 
 /* Keep in sync with INIT_NAMES_SIZE, in syminidf.h. Note that flang1 and
```
pawosm-arm, line 18: same as above, `syminidf.h`

tools/flang1/flang1exe/symacc.h
bryanpkc: Two versions of `syminidf.h` are generated from `tools/flang1/utils/symtab/symtab.n` and `tools/flang2/utils/symtab/symtab.n`, respectively. `INIT_NAMES_SIZE` refers to the size of the name string table in the respective `STB` structure used in `flang1` and `flang2`. This size changes as more symbols are added, or when a symbol's name changes. The one in `flang2` is smaller than `flang1`, but some attempt to reuse more code between `flang1` and `flang2` had led to a requirement that the tables be kept the same size. The `symtab.n` files do not contain or support any macro, so this size is the same whether or not the feature is enabled.

tools/shared/utils/symacc.h
bryanpkc: See above.

pawosm-arm (approved):

runtime/libpgmath/lib/common/cqsin.c
```diff
@@ -0,0 +1,23 @@
...
+ *
+ */
+
+#include "mthdecls.h"
+
+LZMPLXFUNC_LZ(__mth_i_cqsin)
+{
+  LZMPLXARGS_LZ;
+  float128_t x, y;
+  /*
```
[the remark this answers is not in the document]
bryanpkc: There was no reason in particular to write the code like this, other than to conform to the style of PGI's legacy code, e.g.

https://github.com/flang-compiler/flang/blob/master/runtime/libpgmath/lib/common/csin.c#L14-L21

bryanpkc:
@shivaramaarao Please review this PR when you get a chance.

shivaramaarao (approved):

PR #1360, opened 2023-04-07
[RFC]Add align pragma for derived type and fix shape array/character type

JiaweiHawk:
Add support for aligning the first address of some type with configurable alignment.

The overall idea for this patch set is as follows:

- add flg.x[251] for saving pragma alignment value in `do_sw()`
- add the stg_align field in `STG_MEMBERS(dt)` to save the pragma alignment value corresponding to the dtype
- update the alignment value in `alignment()` and `lower_put_datatype()` in flang1
- parse the pragma alignment value passed by flang1 through stb in flang2 in `read_datatype()`
- update the alignment value in `alignment()` in flang2
- each common block should be aligned to the largest alignment value of the type it contains, flang2 should update the expected alignment of a block in `gen_acon_expr()`
- the offset of each object to relative common block should be aligned to relative dtype alignment in `gen_acon_expr()`

Finally, the following effects can be achieved:
```
!dir$ align 0x100
character(len=10)   :: v1

!dir$ align 0x200
type T4
    integer(kind=2)     :: f1
    integer(kind=4)     :: f2
end type T4

!dir$ align 0x400
integer, dimension (5,5) :: v2
```

I test locally through `make check-flang`, and also pass some custom tests and add them to the test folder in patch.
Please correct me if I am wrong, and I am looking forward to feedback for this patch set.

### DIFF
--- a/tools/flang1/flang1exe/dtypeutl.c
+++ b/tools/flang1/flang1exe/dtypeutl.c
@@ -1288,6 +1288,14 @@ alignment_of_var(int sptr)
       align = ta;
   }
 #endif
+  /*
+   * If alignment of variable set by `!DIR$ ALIGN alignment`
+   * in flang1 is smaller than its original, then this pragma
+   * should have no effect.
+   */
+  if (align < PALIGNG(sptr)) {
+    align = PALIGNG(sptr) - 1;
+  }
   return align;
 } /* alignment_of_var */
 
--- a/tools/flang1/flang1exe/exterf.c
+++ b/tools/flang1/flang1exe/exterf.c
@@ -2337,10 +2337,11 @@ export_symbol(int sptr)
   lzprintf(outlz, "S %d", sptr);
   if (exportmode)
     lzprintf(outlz, " %d", HASHLKG(sptr));
-  lzprintf(outlz, " %d %d %d %d %d %d %d %d", stb.stg_base[sptr].stype,
+  lzprintf(outlz, " %d %d %d %d %d %d %d %d %d", stb.stg_base[sptr].stype,
            stb.stg_base[sptr].sc, stb.stg_base[sptr].b3, stb.stg_base[sptr].b4,
            stb.stg_base[sptr].dtype, stb.stg_base[sptr].symlk,
-           stb.stg_base[sptr].scope, stb.stg_base[sptr].nmptr);
+           stb.stg_base[sptr].scope, stb.stg_base[sptr].nmptr,
+           stb.stg_base[sptr].palign);
 
 #undef PUTFIELD
 #undef PUTISZ_FIELD
--- a/tools/flang1/flang1exe/lowersym.c
+++ b/tools/flang1/flang1exe/lowersym.c
@@ -3705,6 +3705,7 @@ lower_symbol(int sptr)
   } else
 #endif
     putval("dtype", dtype);
+  putval("palign", PALIGNG(sptr));
   /* type specific information */
   switch (stype) {
   case ST_ARRAY:
--- a/tools/flang2/flang2exe/dtypeutl.cpp
+++ b/tools/flang2/flang2exe/dtypeutl.cpp
@@ -557,9 +557,21 @@ align_unconstrained(DTYPE dtype)
 int
 alignment_sym(SPTR sym)
 {
-  if (QALNG(sym))
-    return dtypeinfo[TY_DBLE].align;
-  return alignment(DTYPEG(sym));
+  int align;
+  if (QALNG(sym)) {
+    align = dtypeinfo[TY_DBLE].align;
+  } else {
+    align = alignment(DTYPEG(sym));
+  }
+  /*
+   * If alignment of symbol set by `!DIR$ ALIGN alignment`
+   * in flang1 is smaller than its original, then this pragma
+   * should have no effect.
+   */
+  if (align < PALIGNG(sym)) {
+    align = PALIGNG(sym) - 1;
+  }
+  return align;
 }
 
 int
--- a/tools/flang2/flang2exe/llassem.cpp
+++ b/tools/flang2/flang2exe/llassem.cpp
@@ -858,7 +858,7 @@ void
 assem_init(void)
 {
   INT nmptr;
-  SPTR sptr;
+  SPTR sptr, cmem;
   int align8, mod_or_sub, subprog;
   char *typed;
 
@@ -927,6 +927,22 @@ assem_init(void)
       }
     }
     free(typed);
+
+    /*
+     * Update the alignment for cmn.
+     *
+     * To align the symbol set by `!DIR$ ALIGN alignment` pragma
+     * in flang1, flang should align both its symbol's offset
+     * in AG and AG's alignment in memory.
+     *
+     * Here we update the AG_ALIGN(ag) to ensure cmn is aligned
+     * in memory to the maximum alignment among all symbols in
+     * the cmn.
+     */
+    for (cmem = CMEMFG(sptr); cmem > NOSYM; cmem = SYMLKG(cmem)) {
+      AG_ALIGN(gblsym) = AG_ALIGN(gblsym) > PALIGNG(cmem) ?
+                      AG_ALIGN(gblsym) : PALIGNG(cmem);
+    }
   }
 
   /* ag_local gets allocated and deallocate for every function */
@@ -1171,6 +1187,7 @@ assemble_end(void)
         free(AG_CMBLKINITDATA(gblsym));
         AG_CMBLKINITDATA(gblsym) = NULL;
       } else {
+        int align;
         fprintf(ASMFIL, "%%struct%s = type < { %s } > \n", name, typed);
         if (strstr(cpu_llvm_module->target_triple, "windows-msvc") != NULL) {
           fprintf(ASMFIL, "@%s = %s global %%struct%s ", name,
@@ -1179,8 +1196,17 @@ assemble_end(void)
           fprintf(ASMFIL, "@%s = %s global %%struct%s ", name,
                   AG_ISMOD(gblsym) ? "external" : "common", name);
         }
+
+        /*
+         * cmn should align with its corresponding AG's alignment,
+         * so that all symbols within the cmn align with the alignment set by
+         * `!DIR$ ALIGN alignment` pragma in flang1 as long as the symbol's
+         * offset in AG aligns with the specified alignment.
+         */
+        align =
+            align_value > AG_ALIGN(tdefsym) ? align_value : AG_ALIGN(tdefsym);
         fprintf(ASMFIL, "%s, align %d",
-                AG_ISMOD(gblsym) ? "" : " zeroinitializer", align_value);
+                AG_ISMOD(gblsym) ? "" : " zeroinitializer", align);
       }
       for (llObjtodbgFirst(listp, &i); !llObjtodbgAtEnd(&i);
            llObjtodbgNext(&i)) {
@@ -1721,6 +1747,7 @@ write_comm(void)
 
   for (sptr = gbl.cmblks; sptr > NOSYM; sptr = SYMLKG(sptr)) {
     SPTR cmem;
+    int align;
 
     first_data = 1;
     process_sptr(sptr);
@@ -1766,6 +1793,14 @@ write_comm(void)
     else
       gbl.asmfil = cmn_blk_ir;
 
+    /*
+     * cmn should align with its corresponding AG's alignment,
+     * so that all symbols within the cmn align with the alignment set by
+     * `!DIR$ ALIGN alignment` pragma in flang1 as long as the symbol's
+     * offset in AG aligns with the specified alignment.
+     */
+    align = align_value > AG_ALIGN(gblsym) ? align_value : AG_ALIGN(gblsym);
+
     fprintf(ASMFIL, "%%struct%s = type < { %s } > \n", name, type_only);
     fprintf(ASMFIL, "@%s = global %%struct%s", name, name);
     fprintf(ASMFIL, " < { ");
@@ -1774,7 +1809,7 @@ write_comm(void)
 
     DSRTP(sptr, NULL);
 
-    fprintf(ASMFIL, ", align %d", align_value);
+    fprintf(ASMFIL, ", align %d", align);
 
     for (cmem = CMEMFG(sptr); cmem > NOSYM; cmem = SYMLKG(cmem)) {
       if (MIDNUMG(cmem)) /* some member does not have midnum/no name */
--- a/tools/flang2/flang2exe/llassem.cpp
+++ b/tools/flang2/flang2exe/llassem.cpp
@@ -4368,6 +4368,16 @@ assn_static_off(SPTR sptr, DTYPE dtype, ISZ_T size)
   } else {
     a = align_unconstrained(dtype);
   }
+  /*
+   * To align the symbol set by `!DIR$ ALIGN alignment` pragma in flang1,
+   * flang should align both its symbol's offset in AG and AG's alignment
+   * in memory.
+   *
+   * The following code ensures the alignment of the symbol's offset in AG.
+   */
+  if (a < PALIGNG(sptr)) {
+    a = PALIGNG(sptr) - 1;
+  }
   addr = ALIGN(addr, a);
   ADDRESSP(sptr, addr);
   if (DINITG(sptr)) {
--- a/tools/flang2/flang2exe/llassem.cpp
+++ b/tools/flang2/flang2exe/llassem.cpp
@@ -1592,6 +1592,17 @@ write_statics(void)
   DSRT *dsrtp;
   int count = 0;
   char *static_nm = static_name;
+  int align = 16;
+
+  /*
+   * statics should align with its corresponding AG's alignment, so that
+   * all symbols within the BSS align with the alignment set by
+   * `!DIR$ ALIGN alignment` pragma in flang1 as long as the symbol's
+   * offset in AG aligns with the specified alignment.
+   */
+  for (SPTR sptr = gbl.statics; sptr > NOSYM; sptr = SYMLKG(sptr)) {
+    align = align > PALIGNG(sptr) ? align : PALIGNG(sptr);
+  }
 
   if (lcl_inits) {
     if (DBGBIT(5, 32)) {
@@ -1609,7 +1620,7 @@ write_statics(void)
     fprintf(ASMFIL, "%%struct%s = type <{ %s }>\n", static_nm, type_only);
     fprintf(ASMFIL, "@%s = %s %%struct%s <{ ", static_nm, type_str, static_nm);
     process_dsrt(lcl_inits, gbl.saddr, typed, false, 0);
-    fprintf(ASMFIL, " }>, align 16");
+    fprintf(ASMFIL, " }>, align %d", align);
     ll_write_object_dbg_references(ASMFIL, cpu_llvm_module, static_dbg_list);
     static_dbg_list = NULL;
     fputc('\n', ASMFIL);
@@ -1619,8 +1630,8 @@ write_statics(void)
             (long)gbl.saddr);
     fprintf(ASMFIL,
             "@%s = %s %%struct%s <{ [%ld x i8] zeroinitializer }>"
-            ", align 16",
-            static_name, type_str, static_name, (long)gbl.saddr);
+            ", align %d",
+            static_name, type_str, static_name, (long)gbl.saddr, align);
     ll_write_object_dbg_references(ASMFIL, cpu_llvm_module, static_dbg_list);
     static_dbg_list = NULL;
     fputc('\n', ASMFIL);
--- a/tools/flang2/flang2exe/llassem.cpp
+++ b/tools/flang2/flang2exe/llassem.cpp
@@ -1521,12 +1521,22 @@ write_bss(void)
   char *bss_nm = bss_name;
 
   if (gbl.bss_addr) {
+    /*
+     * BSS should align with its corresponding AG's alignment, so that
+     * all symbols within the BSS align with the alignment set by
+     * `!DIR$ ALIGN alignment` pragma in flang1 as long as the symbol's
+     * offset in AG aligns with the specified alignment.
+     */
+    int align = 32;
+    for (SPTR sptr = gbl.bssvars; sptr > NOSYM; sptr = SYMLKG(sptr)) {
+      align = align > PALIGNG(sptr) ? align : PALIGNG(sptr);
+    }
     fprintf(ASMFIL, "%%struct%s = type <{[%" ISZ_PF "d x i8]}>\n", bss_nm,
             gbl.bss_addr);
     fprintf(ASMFIL,
             "@%s = %s %%struct%s <{[%" ISZ_PF "d x i8] "
-            "zeroinitializer }> , align 32",
-            bss_nm, type_str, bss_nm, gbl.bss_addr);
+            "zeroinitializer }> , align %d",
+            bss_nm, type_str, bss_nm, gbl.bss_addr, align);
     ll_write_object_dbg_references(ASMFIL, cpu_llvm_module, bss_dbg_list);
     bss_dbg_list = NULL;
     fputc('\n', ASMFIL);
--- a/tools/flang2/flang2exe/llassem.cpp
+++ b/tools/flang2/flang2exe/llassem.cpp
@@ -4288,19 +4288,29 @@ unsigned
 align_of_var(SPTR sptr)
 {
   DTYPE dtype = DTYPEG(sptr);
-  if (!PDALN_IS_DEFAULT(sptr))
-    return 1u << PDALNG(sptr);
-  if (QALNG(sptr))
-    return 4 * align_of(DT_INT);
-  if (dtype) {
+  int align = 0;
+  if (!PDALN_IS_DEFAULT(sptr)) {
+    align = 1u << PDALNG(sptr);
+  } else if(QALNG(sptr)) {
+    align = 4 * align_of(DT_INT);
+  } else if (dtype) {
     if (flg.quad && !DESCARRAYG(sptr) && zsize_of(dtype) >= MIN_ALIGN_SIZE) {
-      return DATA_ALIGN + 1;
+      align = DATA_ALIGN + 1;
+    } else {
+      align = align_of(dtype);
     }
-    return align_of(dtype);
+  } else if(STYPEG(sptr) == ST_PROC) {/* No DTYPE */
+    align = align_of(DT_ADDR);
   }
-  if (STYPEG(sptr) == ST_PROC) /* No DTYPE */
-    return align_of(DT_ADDR);
-  return 0;
+  /*
+   * If alignment of variable set by `!DIR$ ALIGN alignment`
+   * in flang1 is smaller than its original, then this pragma
+   * should have no effect.
+   */
+  if (align < PALIGNG(sptr)) {
+    align = PALIGNG(sptr);
+  }
+  return align;
 }
 
 static void
--- a/tools/shared/pragma.c
+++ b/tools/shared/pragma.c
@@ -1053,7 +1053,28 @@ do_sw(void)
   case SW_ESCTYALIAS:
     break;
   case SW_ALIGN:
-    break;
+    if (gtok() != T_INT) {
+      int backup_nowarn = gbl.nowarn;
+      gbl.nowarn = false;
+      error((error_code_t)280, ERR_Warning, lineno,
+            "ALIGN: non-integer alignment", 0);
+      gbl.nowarn = backup_nowarn;
+      return true;
+    }
+
+    /* check whether the alignment is power of 2 */
+    if (itok <= 0 || ((itok & (itok - 1)) != 0)) {
+      int backup_nowarn = gbl.nowarn;
+      gbl.nowarn = false;
+      error((error_code_t)280, ERR_Warning, lineno,
+            "ALIGN: non-power-of-2 alignment", 0);
+      gbl.nowarn = backup_nowarn;
+      return true;
+    }
+
+    TR1("SW_ALIGN alignment[%d]\n", itok);
+    flg.x[251] = itok;
+    return true;
   case SW_BOUNDS:
     if (no_specified) {
       bclr(DIR_OFFSET(currdir, x[70]), 0x02);
--- a/test/directives/align_array_variable.f90
+++ b/test/directives/align_array_variable.f90
@@ -1,0 +1,140 @@
+! RUN: %flang -O0 -S -emit-llvm %s -o - | FileCheck %s
+
+! CHECK: %struct[[BLOCK1:\.BSS[0-9]+]] = type <{ [356 x i8] }>
+! CHECK: %struct[[BLOCK2:\.BSS[0-9]+]] = type <{ [612 x i8] }>
+! CHECK: %struct[[BLOCK3:\.BSS[0-9]+]] = type <{ [1124 x i8] }>
+! CHECK: %struct[[BLOCK4:\.BSS[0-9]+]] = type <{ [2148 x i8] }>
+! CHECK: %struct[[BLOCK5:_module_align_array_[0-9]+_]] = type <{ [228 x i8] }>
+! CHECK: @[[BLOCK1]] = internal global %struct[[BLOCK1]] zeroinitializer, align 256
+! CHECK: @[[BLOCK2]] = internal global %struct[[BLOCK2]] zeroinitializer, align 512
+! CHECK: @[[BLOCK3]] = internal global %struct[[BLOCK3]] zeroinitializer, align 1024
+! CHECK: @[[BLOCK4]] = internal global %struct[[BLOCK4]] zeroinitializer, align 2048
+! CHECK: @[[BLOCK5]] = common global %struct[[BLOCK5]] zeroinitializer, align 128
+
+module module_align_array
+implicit none
+
+    !DIR$ ALIGN 128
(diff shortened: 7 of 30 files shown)


JiaweiHawk:
Gentle Ping...

JiaweiHawk:
Thank @kiranchandramohan for reviewing.

> We would like to support directives in commonly used formats. Is this based on some other compiler's directives? The only one I could find was Intel's which has the syntax `!DIR$ ATTRIBUTES ALIGN : 64 :: R_alloc` https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-0/attributes-align.html

I chose this format only because I found the keyword "SW_ALIGN" in the source code, so I reused that part of the code, which resulted in this format.
It is okay for me to use syntax `!DIR$ ATTRIBUTES ALIGN : alignment :: object`, I will refactor my patch to support this format.

> 
> Also think having a hex number is probably not good for readability.

Here I did not specify a particular format for the numbers, I just use `gtok()` and `itok` to get its value, so both hexadecimal and decimal formats should work. However, I prefer hex number, because the alignment value should be a power of 2, which is clearer when represented in hexadecimal format.

By the way, there seems to be some errors in the test cases added in this patch, I will try to fix them later.

test/directives/align_character_type.f90
```diff
@@ -0,0 +1,111 @@
...
+! CHECK:    %[[V10:v10_[0-9]+]] = alloca [10 x i8], align 2048
+
+    v9 = "801"
+! CHECK:      store volatile i64 %{{[0-9]+}}, ptr %[[V9]], align
+
+    v10 = "802"
+! CHECK:      store volatile i64 %{{[0-9]+}}, ptr %[[V10]], align
+
+    return
+end function function_align
```
pawosm-arm, line 111: A small remark regarding all of the new files you have added. This red sign which github has visibly added is a warning which tells you that you did not hit 'Enter' after the very last line of a file (e.g. here you need to press 'Enter' after `end function function_align`). This applies to every new file being introduced.

JiaweiHawk: Thanks @pawosm-arm for the review, I will refactor the patches according to your comments.

test/directives/align_array_type.f90
```diff
@@ -0,0 +1,157 @@
...
+! CHECK: @[[BLOCK1]] = internal global %struct[[BLOCK1]] zeroinitializer, align 256
+! CHECK: @[[BLOCK2]] = internal global %struct[[BLOCK2]] zeroinitializer, align 512
+! CHECK: @[[BLOCK3]] = internal global %struct[[BLOCK3]] zeroinitializer, align 1024
+! CHECK: @[[BLOCK4]] = internal global %struct[[BLOCK4]] zeroinitializer, align 2048
+! CHECK: @[[BLOCK5]] = common global %struct[[BLOCK5]] zeroinitializer, align 128
+
+module module_align
+implicit none
+
+    !dir$ align 0x80
```
pawosm-arm, line 17: I'm finding this use of hex slightly awkward. The Fortran world is populated by mathematicians and physicists, they don't usually reason in hex numbers. Apart from `!pgi$` pragmas which control the xflags directly, our test cases don't seem to use it like that. Can decimals be used here (as well or instead)?

pawosm-arm:
> Thanks @JiaweiHawk for this RFC and patch.
> 
> We would like to support directives in commonly used formats. Is this based on some other compiler's directives? The only one I could find was Intel's which has the syntax `!DIR$ ATTRIBUTES ALIGN : 64 :: R_alloc` https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-0/attributes-align.html
> 
> Also think having a hex number is probably not good for readability.

IMO if the cost of sticking to Intel's syntax is too high and requires major refactor, then we'd rather use the simpler syntax proposed by the author (as it is simpler and straightforward). I do agree though about use of hex, it's uncommon in the Fortran world.

test/directives/align_array_type.f90
JiaweiHawk: Hi @pawosm-arm, 

> I'm finding this use of hex slightly awkward. The Fortran world is populated by mathematicians and physicists, they don't usually reason in hex numbers. Apart from `!pgi$` pragmas which control the xflags directly, our test cases don't seem to use it like that. Can decimals be used here (as well or instead)?

Thanks for your explanation.

This align pragma supports both hexadecimal and decimal formats. In the patch I use gtok() and `itok` to get alignment value, and gtok() can parse both hexadecimal and decimal formats.

I will refactor the test cases into decimal format as you and @kiranchandramohan  suggested.

JiaweiHawk:
> > Thanks @JiaweiHawk for this RFC and patch.
> > We would like to support directives in commonly used formats. Is this based on some other compiler's directives? The only one I could find was Intel's which has the syntax `!DIR$ ATTRIBUTES ALIGN : 64 :: R_alloc` https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-0/attributes-align.html
> > Also think having a hex number is probably not good for readability.
> 
> IMO if the cost of sticking to Intel's syntax is too high and requires major refactor, then we'd rather use the simpler syntax proposed by the author (as it is simpler and straightforward). I do agree though about use of hex, it's uncommon in the Fortran world.

Yes, changing to the Intel's syntax does require some refactor on the `tools/shared/pragma.c`. We need to add the `attributes` entry in pragmas & directives `table`, and in its handler we need to find the corresponding `dtype` according to pragma's argument and set its alignment value.

In fact, there is already `align` entry in pragmas & directives `table`, but just implementing nothing in its handler. So I reuse the
syntax `!DIR$ ALIGN` and achieve its function with a small amount of code.

If possible, I'd like to keep my original syntax, which allows me to do less work.

As for the use of hex, this patch supports both hexadecimal and decimal formats. And I have changed the test case to  decimal formats and it works fine.

pawosm-arm:
> If possible, I'd like to keep my original syntax, which allows me to do less work.

I'ts ok for me.

JiaweiHawk:
Hi @kiranchandramohan,
> Thank @kiranchandramohan for reviewing.
> 
> > We would like to support directives in commonly used formats. Is this based on some other compiler's directives? The only one I could find was Intel's which has the syntax `!DIR$ ATTRIBUTES ALIGN : 64 :: R_alloc` https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-0/attributes-align.html
> 
> I chose this format only because I found the keyword "SW_ALIGN" in the source code, so I reused that part of the code, which resulted in this format. It is okay for me to use syntax `!DIR$ ATTRIBUTES ALIGN : alignment :: object`, I will refactor my patch to support this format.

If it is possible, can I still keep my original syntax?

It appears that Flang originally intended to implement `!DIR$ align` synatx, as there are `align` entry in pragmas & directives `table` in `tools/shared/pragma.c`, but no code in its handler.

So I think we can reuse this interface, and this can allow me to do less work.

kiranchandramohan:
> Hi @kiranchandramohan,
> 
> > Thank @kiranchandramohan for reviewing.
> > > We would like to support directives in commonly used formats. Is this based on some other compiler's directives? The only one I could find was Intel's which has the syntax `!DIR$ ATTRIBUTES ALIGN : 64 :: R_alloc` https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-0/attributes-align.html
> > 
> > 
> > I chose this format only because I found the keyword "SW_ALIGN" in the source code, so I reused that part of the code, which resulted in this format. It is okay for me to use syntax `!DIR$ ATTRIBUTES ALIGN : alignment :: object`, I will refactor my patch to support this format.
> 
> If it is possible, can I still keep my original syntax?
> 
> It appears that Flang originally intended to implement `!DIR$ align` synatx, as there are `align` entry in pragmas & directives `table` in `tools/shared/pragma.c`, but no code in its handler.
> 
> So I think we can reuse this interface, and this can allow me to do less work.

If we are using a non-standard format then we will have to define the syntax, meaning, restrictions (which values can be provided, which types  apply). Also, we will be setting a precedent and if the format is not available in `llvm/flang` then users might face issues in porting.

Having said that, we do not want to spend too much time adding the feature to Classic Flang. So I am OK if @bryanpkc and @shivaramaarao are OK. @pawosm-arm has already given his go ahead.

tools/flang1/flang1exe/dtypeutl.c
```diff
@@ -1417,6 +1424,25 @@ get_type(int n, TY_KIND v1, int v2)
...
+     * flang1 reaches here only when creating
+     * new dtype. So flang1 should handle the
+     * possible align pragma for new type here.
+     *
+     * flang1 should store the alignment from flg.x[251]
+     * to DTA, later it will updates the dtype alignment
+     * when in alignment() or lower_put_datatype().
+     *
+     * Note that this pragma should only be used to
+     * this strucutre, so flang1 need to clear the flg.x[251]
```
bryanpkc, line 1439: NIT: Please correct capitalization, punctuation, and spelling in the comment. Consider using clang-format to wrap the text to 80 columns for a consistent look.

- `it will updates` &rarr; `it will update`
- `this pragma should only be used to this strucutre` $rarr; `this pragma should apply only to this structure`

tools/flang1/flang1exe/exterf.c
```diff
@@ -2097,18 +2097,40 @@ export_dt(int dtype)
       }
     } else /* 'null' descriptor */
       lzprintf(outlz, " %d", 0);
+
+    /* If align pragma value is smaller than its original alignment,
+     * then align pragma should have no effect
+     */
+    if (DTA(dtype) == alignment(dtype))
```
bryanpkc, line 2104: The comment seems inconsistent with the code. Should it say "not equal to" instead of "smaller than"?

tools/flang1/flang1exe/exterf.c
```diff
@@ -2097,18 +2097,40 @@ export_dt(int dtype)
...
     break;
 
   case TY_CHAR:
   case TY_NCHAR:
     lzprintf(outlz, " %d", (int)DTY(dtype + 1));
+
+    /* If align pragma value is smaller than its original alignment,
+     * then align pragma should have no effect
+     */
+    if (DTA(dtype) == alignment(dtype))
```
bryanpkc, line 2129: Same here.

tools/flang1/flang1exe/lowersym.c
```diff
@@ -2715,6 +2715,15 @@ lower_put_datatype(int dtype, int usage)
         }
       }
     }
+
+    /* If align pragma value is smaller than its original alignment,
+     * then align pragma should have no effect
+     */
+    if (DTA(dtype) == alignment(dtype))
```
bryanpkc, line 2722: `==` &rarr; `>` ?

tools/flang1/flang1exe/lowersym.c
```diff
@@ -3071,6 +3084,15 @@ lower_put_datatype(int dtype, int usage)
       }
     }
     putsym("numelm", numelm);
+
+    /* If align pragma value is smaller than its original alignment,
+     * then align pragma should have no effect
+     */
+    if (DTA(dtype) == alignment(dtype))
```
bryanpkc, line 3091: `==` &rarr; `>` ?

tools/flang1/flang1exe/lowersym.c
```diff
@@ -2753,7 +2762,11 @@ lower_put_datatype(int dtype, int usage)
     putsym("member", DTY(dtype + 1));
     putval("size", DTY(dtype + 2));
     putsym("tag", DTY(dtype + 3));
-    putval("align", DTY(dtype + 4));
+
+    if (DTA(dtype) > DTY(dtype + 4))
```
bryanpkc, line 2766: Comment?

tools/flang1/flang1exe/exterf.c
```diff
@@ -2097,18 +2097,40 @@ export_dt(int dtype)
...
     break;
   case TY_UNION:
   case TY_STRUCT:
   case TY_DERIVED:
     /*  print dtype and  descriptor entry */
-    lzprintf(outlz, " %d %d %d %d", (int)DTY(dtype + 1), (int)DTY(dtype + 2),
-             (int)DTY(dtype + 3), (int)DTY(dtype + 4));
+    lzprintf(outlz, " %d %d %d", (int)DTY(dtype + 1), (int)DTY(dtype + 2),
+             (int)DTY(dtype + 3));
+    if (DTA(dtype) > DTY(dtype + 4))
```
bryanpkc, line 2116: Comment?

tools/flang1/utils/symtab/sharedefs.h
```diff
@@ -30,21 +30,21 @@
 
 /* declare:
  *  struct{
- *     dt* stg_base;
+ *     dt* stg_base, *stg_align;
```
bryanpkc, line 33: `STG_MEMBERS` is instantiated in multiple data structures (twice in `STB` and three times in `ASTB`, among others), but it seems that only `stb.dt.stg_align` is ever used. Did you consider another way to declare this array that could have less impact on unrelated types?

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -12628,11 +12629,30 @@ gen_acon_expr(int ilix, LL_Type *expected_type)
   sym_is_refd(sptr);
   idx = (STYPEG(sptr) == ST_STRUCT || STYPEG(sptr) == ST_ARRAY
          || ACONOFFG(opnd) < 0) ? 0 : ACONOFFG(opnd);
-  process_sptr_offset(sptr, variable_offset_in_aggregate(sptr, idx));
+  if (DTA(dtype)) {
```
bryanpkc, line 12632: NIT: Please format the whole `if`-`else` statement with clang-format. Capitalize and punctuate the comment.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -12628,11 +12629,30 @@ gen_acon_expr(int ilix, LL_Type *expected_type)
...
+                                      variable_offset_in_aggregate(sptr, 0),
+                                      DTA(dtype)));
+  }else {
+    process_sptr_offset(sptr, variable_offset_in_aggregate(sptr, idx));
+  }
   idx = ACONOFFG(opnd); /* byte offset */
 
   ty1 = make_lltype_from_dtype(DT_ADDR);
-  idx = variable_offset_in_aggregate(sptr, idx);
+  if (DTA(dtype)) {
```
bryanpkc, line 12646: Ditto.

tools/flang2/flang2exe/cgmain.cpp
bryanpkc: It may be clearer to rewrite like this (following the example of the code below):
```c
  /* Handle !dir$ align pragma. Adjust sptr's offset in ag to be aligned with
   * sptr's alignment.
   */
  if (DTA(dtype))
    idx += ALIGN(variable_offset_in_aggregate(sptr, 0), DTA(dtype));
  else
    idx = variable_offset_in_aggregate(sptr, idx);
  process_sptr_offset(sptr, idx);
```

tools/flang2/flang2exe/dtypeutl.cpp
```diff
@@ -512,7 +512,10 @@ alignment(DTYPE dtype)
   case TY_LOG128:
   case TY_FLOAT128:
   case TY_CMPLX128:
-    return dtypeinfo[ty].align;
+    align_bits = dtypeinfo[ty].align;
```
bryanpkc, line 515: `align_bits` is a misleading variable name. Can we rename it to `align_bytes`?

tools/flang2/flang2exe/llassem.cpp
```diff
@@ -1506,12 +1513,24 @@ write_bss(void)
   char *bss_nm = bss_name;
 
   if (gbl.bss_addr) {
+    char gname[MXIDLN + 50];
+    int align;
+    SPTR ag;
+
+    /* handle the flang1 pragma !dir$ align
+     * Here we refer bss's alignment from
```
bryanpkc, line 1521: `refer` &rarr; `infer` ?

tools/shared/pragma.c
```diff
@@ -1053,7 +1053,29 @@ do_sw(void)
   case SW_ESCTYALIAS:
     break;
   case SW_ALIGN:
-    break;
+    /* declaration align XX */
+    if (gtok() != T_INT) {
+      int backup_nowarn = gbl.nowarn;
+      gbl.nowarn = false;
+      error((error_code_t)280, ERR_Warning, lineno, \
+      "ALIGN: allow int alignment only", 0);
```
bryanpkc, line 1061: Suggestion for error message: `ALIGN: non-integer alignment`

tools/shared/pragma.c
```diff
@@ -1053,7 +1053,29 @@ do_sw(void)
...
+      gbl.nowarn = backup_nowarn;
+      return true;
+    }
+
+    /* check whether the alignment is power of 2 */
+    if (itok <= 0 || ((itok & (itok - 1)) != 0)) {
+      int backup_nowarn = gbl.nowarn;
+      gbl.nowarn = false;
+      error((error_code_t)280, ERR_Warning, lineno, \
+      "ALIGN: allow power of 2 alignment only", 0);
```
bryanpkc, line 1071: Suggestion for error message: `ALIGN: non-power-of-2 alignment`

JiaweiHawk:
Thank @bryanpkc for reviewing.

I will find time to refactor these patches based on your feedback and resolve the conflicts.

bryanpkc:
@JiaweiHawk gentle ping

JiaweiHawk:
> @JiaweiHawk gentle ping

Hi, @bryanpkc , I might be occupied until September this year. However, I'll try my best to refactor this patch series as soon as I find some free time.

Thanks for your understanding!

tools/flang1/flang1exe/dtypeutl.c
JiaweiHawk: > NIT: Please correct capitalization, punctuation, and spelling in the comment. Consider using clang-format to wrap the text to 80 columns for a consistent look.

Thank you for your suggestion. I hope I have addressed all of these issues in the latest patches.

> 
>     * `it will updates` → `it will update`
> 
>     * `this pragma should only be used to this strucutre` → `this pragma should apply only to this structure`

tools/flang1/flang1exe/lowersym.c
JiaweiHawk: > `==` → `>` ?

Yes, you are correct. I have corrected the code using >.

tools/flang1/flang1exe/exterf.c
JiaweiHawk: > Same here.

Yes, you are correct. I have corrected the code using >.

tools/flang1/flang1exe/lowersym.c
JiaweiHawk: > Comment?

This code is used to export alignment values to the .stb file. I have refactored this section of the code and added the necessary comment to the import code part.

tools/flang1/flang1exe/exterf.c
JiaweiHawk: > Comment?

This code is used to export alignment values to the .mod file. I have refactored this section of the code and added the necessary comment to the import code part.

tools/flang1/utils/symtab/sharedefs.h
JiaweiHawk: > `STG_MEMBERS` is instantiated in multiple data structures (twice in `STB` and three times in `ASTB`, among others), but it seems that only `stb.dt.stg_align` is ever used. Did you consider another way to declare this array that could have less impact on unrelated types?

Yes, you are right. I declare this array in `STB` structure as `  struct{
    STG_MEMBERS(ISZ_T);
  }dt_align;` field in the latest patches.

tools/flang2/flang2exe/cgmain.cpp
JiaweiHawk: > It may be clearer to rewrite like this (following the example of the code below):
> 
> ```c
>   /* Handle !dir$ align pragma. Adjust sptr's offset in ag to be aligned with
>    * sptr's alignment.
>    */
>   if (DTA(dtype))
>     idx += ALIGN(variable_offset_in_aggregate(sptr, 0), DTA(dtype));
>   else
>     idx = variable_offset_in_aggregate(sptr, idx);
>   process_sptr_offset(sptr, idx);
> ```

Thanks for you suggestion.
This part is no longer needed in the latest patches.

tools/flang2/flang2exe/cgmain.cpp
JiaweiHawk: > Ditto.

I have formatted the statements using clang-format.

tools/flang2/flang2exe/dtypeutl.cpp
JiaweiHawk: > `align_bits` is a misleading variable name. Can we rename it to `align_bytes`?

Yes, you are right. I have refactored this part of the code.

tools/flang2/flang2exe/llassem.cpp
JiaweiHawk: > `refer` → `infer` ? `belonging ag` → `corresponding ag`?

Thank you, I have revised this comment.

tools/shared/pragma.c
JiaweiHawk: > Suggestion for error message: `ALIGN: non-integer alignment`

Thank you, I have revised this error message.

tools/shared/pragma.c
JiaweiHawk: > Suggestion for error message: `ALIGN: non-power-of-2 alignment`


Thank you, I have revised this error message.

erickq:
```
#ifndef X
#define X 216
#endif
#ifndef Y
#define Y 37
#endif
#ifndef Z
#define Z 197
#endif

program main
  INTEGER,parameter :: ims=1,ime=X,jms=1,jme=Y,kms=1,kme=Z
  real :: c1,c2,c3,cn,cs
  real :: time_start, time_end

  !DIR$ ALIGN 64
  REAL,DIMENSION(ims:ime,kms:kme,jms:jme) :: cpi1,cpi2,cpi3,cu0,cv0,cw0
  !DIR$ ALIGN 64
  REAL,DIMENSION(jms:jme) :: cosv
  !DIR$ ALIGN 64
  REAL,DIMENSION(ims:ime,kms:kme,jms:jme) :: b_helm

  print *, loc(cpi1)
  print *, loc(cpi2)
  print *, loc(cpi3)
  print *, loc(cu0)
  print *, loc(cv0)
  print *, loc(cw0)
  print *, loc(cosv)
  print *, loc(b_helm)

end program
```
Hi, I tried cherry-picking your patch build compiler and found that the arrays are not 64-byte aligned. Is there any plan to fix it in this patch?

JiaweiHawk:
> ```
> #ifndef X
> #define X 216
> #endif
> #ifndef Y
> #define Y 37
> #endif
> #ifndef Z
> #define Z 197
> #endif
> 
> program main
>   INTEGER,parameter :: ims=1,ime=X,jms=1,jme=Y,kms=1,kme=Z
>   real :: c1,c2,c3,cn,cs
>   real :: time_start, time_end
> 
>   !DIR$ ALIGN 64
>   REAL,DIMENSION(ims:ime,kms:kme,jms:jme) :: cpi1,cpi2,cpi3,cu0,cv0,cw0
>   !DIR$ ALIGN 64
>   REAL,DIMENSION(jms:jme) :: cosv
>   !DIR$ ALIGN 64
>   REAL,DIMENSION(ims:ime,kms:kme,jms:jme) :: b_helm
> 
>   print *, loc(cpi1)
>   print *, loc(cpi2)
>   print *, loc(cpi3)
>   print *, loc(cu0)
>   print *, loc(cv0)
>   print *, loc(cw0)
>   print *, loc(cosv)
>   print *, loc(b_helm)
> 
> end program
> ```
> 
> Hi, I tried cherry-picking your patch build compiler and found that the arrays are not 64-byte aligned. Is there any plan to fix it in this patch?

Hi @erickq,

Thanks for testing this patchset.
I will look into your test case as soon as possible.

JiaweiHawk:
> > ```
> > #ifndef X
> > #define X 216
> > #endif
> > #ifndef Y
> > #define Y 37
> > #endif
> > #ifndef Z
> > #define Z 197
> > #endif
> > 
> > program main
> >   INTEGER,parameter :: ims=1,ime=X,jms=1,jme=Y,kms=1,kme=Z
> >   real :: c1,c2,c3,cn,cs
> >   real :: time_start, time_end
> > 
> >   !DIR$ ALIGN 64
> >   REAL,DIMENSION(ims:ime,kms:kme,jms:jme) :: cpi1,cpi2,cpi3,cu0,cv0,cw0

In the current implementation, each `!DIR$ ALIGN` pragma only applies to a single dtype, while in this case, each array creates a new dtype.

But I think this definition approach should also be supported, I will try to refactor the patchset to support it.

> >   !DIR$ ALIGN 64
> >   REAL,DIMENSION(jms:jme) :: cosv
> >   !DIR$ ALIGN 64
> >   REAL,DIMENSION(ims:ime,kms:kme,jms:jme) :: b_helm
> > 
> >   print *, loc(cpi1)
> >   print *, loc(cpi2)
> >   print *, loc(cpi3)
> >   print *, loc(cu0)
> >   print *, loc(cv0)
> >   print *, loc(cw0)
> >   print *, loc(cosv)
> >   print *, loc(b_helm)
> > 
> > end program
> > ```
> > 
> > 
> > Hi, I tried cherry-picking your patch build compiler and found that the arrays are not 64-byte aligned. Is there any plan to fix it in this patch?
> 
> Hi @erickq,
> 
> Thanks for testing this patchset. I will look into your test case as soon as possible.

erickq:
> > > ```
> > > #ifndef X
> > > #define X 216
> > > #endif
> > > #ifndef Y
> > > #define Y 37
> > > #endif
> > > #ifndef Z
> > > #define Z 197
> > > #endif
> > > 
> > > program main
> > >   INTEGER,parameter :: ims=1,ime=X,jms=1,jme=Y,kms=1,kme=Z
> > >   real :: c1,c2,c3,cn,cs
> > >   real :: time_start, time_end
> > > 
> > >   !DIR$ ALIGN 64
> > >   REAL,DIMENSION(ims:ime,kms:kme,jms:jme) :: cpi1,cpi2,cpi3,cu0,cv0,cw0
> > > ```
> 
> In the current implementation, each `!DIR$ ALIGN` pragma only applies to a single dtype, while in this case, each array creates a new dtype.
> 
> But I think this definition approach should also be supported, I will try to refactor the patchset to support it.
> 
> > > !DIR$ ALIGN 64
> > > REAL,DIMENSION(jms:jme) :: cosv
> > > !DIR$ ALIGN 64
> > > REAL,DIMENSION(ims:ime,kms:kme,jms:jme) :: b_helm
> > > print *, loc(cpi1)
> > > print *, loc(cpi2)
> > > print *, loc(cpi3)
> > > print *, loc(cu0)
> > > print *, loc(cv0)
> > > print *, loc(cw0)
> > > print *, loc(cosv)
> > > print *, loc(b_helm)
> > > end program
> > > ```
> > > 
> > > 
> > > Hi, I tried cherry-picking your patch build compiler and found that the arrays are not 64-byte aligned. Is there any plan to fix it in this patch?
> > > ```
> > 
> > 
> > Hi @erickq,
> > Thanks for testing this patchset. I will look into your test case as soon as possible.
Okay, come on.

tools/shared/symfun.h
```diff
@@ -234,6 +234,10 @@ inline void DTySet(DTYPE dtype, ISZ_T val) {
   stb.dt.stg_base[static_cast<int>(dtype)] = val;
 }
 
+inline void DTaSet(DTYPE dtype, ISZ_T val) {
+  stb.dt_align.stg_base[static_cast<int>(dtype)] = val;
```
bryanpkc, line 238: Instead of using a separate table, did you consider storing the alignment information in `dt` itself at an offset, e.g. `stb.dt.stg_base[static_cast<int>(dtype) + 6]`? Could you estimate the amount of effort that would require?

JiaweiHawk:
> LGTM in general. I would prefer that the alignment information be stored in the `dt` table itself but I understand that it would complicate things quite a bit.

Thanks @bryanpkc for the review.

Presently, there are some incompatibilities between my patch and @GARRYHU's patch at #1427, so we intend to establish a foundational infrastructure that caters to both. If the new infrastructure still requires the use of a separate table to record the alignment information, we will address that issue at that point.

JiaweiHawk:
> LGTM

Thanks @shivaramaarao for the review.

In order to ensure compatibility with @GARRYHU's patchset at #1427 and address the issues pointed out by @bryanpkc and @erickq, I and @GARRYHU collaboratively developed the latest solution. Here is the changelog from the previous version.

ChangeLog
=======
1. Binding the alignment of variable to a newly added field in its `sptr` during reduction.
2. Iterating through variables in `ag` in `write_bss()`,`write_statics()` and `assem_init()` to determine the alignment for `ag` as the least common multiple of the alignment for these variables.
3. Small fixes

Approach
======
Now the overall approach remains focused on ensuring the alignment of both the `ag` containing the symbol and the offset of the symbol within `ag`. The specific detail are below:
1. parse the pragma within `p_pragma()` and clear it within the reduction rule `<stmt> ::= <stbeg> <statement> <stend>` to ensure that pragma scope is limited to the statement level.
2. bind the variable alignment to its `sptr` in the `ENTITY_IDx` reduction reule.
3. export the `palign` field of symbol to .stb/.mod file and import it from .stb/.mod file.
4. iterate through all variable symbols in ag, setting the alignment of `ag` to the least common multiple of these symbol alignment values.
5. Fix alignment values in `alignment_of_var()`,`alignment_var()` and `align_of_var()` to ensure that the offset of the symbol within `ag` is aligned.

Advantages
=======
Compared to the previous version, the latest version has advantages in the following aspects:
1. use the newly added field `palign` in the `struct SYM` instead of a seperate table, addresses the issue pointed out by @bryanpkc.
2. bind all variables within the statement to the alignment set by pragma, addresses the issue pointed out by @erickq.
3. ensure compatibility with @GARRYHU's patchset at #1427, and can easily extends to nearly all types of variables.

Notes
====
For the consistency, pragmas only affect variables and will not have any impact on derived types as previous version. However, achieving alignment for derived types as previous implementations can be easily accomplished by binding the alignment value to the derived type symbol in reduction.

pawosm-arm (approved):

bryanpkc (changes requested):
I believe that adding a `palign` field to symbols in the ILM output necessitates bumping up the ILM version number in `tools/flang2/flang2exe/upper.h`.

JiaweiHawk:
> I believe that adding a `palign` field to symbols in the ILM output necessitates bumping up the ILM version number in `tools/flang2/flang2exe/upper.h`.

Thanks @bryanpkc for the review.

I've identified the similar issue in `tools/flang1/flang1exe/lower.h`, and I'll address that as well. 

Additionally, I would like to inquire how to generate the numeric values on the left side of the comments, it seems that there isn't much information on how these numbers should be generated.

```c
/*
 * 20.1         -- 1.55
 *              All of 1.54 +
 *              pass elemental field for subprogram when emitting ST_ENTRY.
 *
 *              For ST_PROC, receive IS_PROC_PTR_IFACE flag.
 *
 * ${how to generate this number}         -- 1.56
 *              All of 1.55 + PALIGN
 */

#include "gbldefs.h"
#include "semant.h"

#define VersionMajor 1
#define VersionMinor 56

```

bryanpkc:
> Additionally, I would like to inquire how to generate the numeric values on the left side of the comments, it seems that there isn't much information on how these numbers should be generated.

@JiaweiHawk I think those numbers are nvfortran version numbers, and are no longer relevant in open-source Classic Flang. You could use something like `23.12` to represent December 2023.

JiaweiHawk:
> > Additionally, I would like to inquire how to generate the numeric values on the left side of the comments, it seems that there isn't much information on how these numbers should be generated.
> 
> @JiaweiHawk I think those numbers are nvfortran version numbers, and are no longer relevant in open-source Classic Flang. You could use something like `23.12` to represent December 2023.

Thank you for explanation. I will update this patchset according to your suggestions.

JiaweiHawk:
Hi all,

It seems like CI does not find the clang binary.

build_flang (AArch64, release_15x) CI encountered an error during the "Check tools" step, the complete log is provided below:

> Run git --version
> git version 2.25.1
> cmake version 3.20.0
> 
> CMake suite maintained and supported by Kitware (kitware.com/cmake).
> GNU Make 4.2.1
> Built for aarch64-unknown-linux-gnu
> Copyright (C) 1988-2016 Free Software Foundation, Inc.
> License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
> This is free software: you are free to change and redistribute it.
> There is NO WARRANTY, to the extent permitted by law.
> /__w/_temp/600f1194-c66d-4760-ac8f-f663a73d066f.sh: 4: /home/github/usr/local/bin/clang: not found
> Error: Process completed with exit code 127.

bryanpkc:
> It seems like CI does not find the clang binary.

Yep, I became aware of the CI failures yesterday and have been trying to fix them. The cause is that the `release_15x` container image had expired and been deleted by GitHub. I will regenerate the image and rerun the workflow for this PR.

bryanpkc (approved):
LGTM. (Sorry I had re-requested a review from you by mistake, @pawosm-arm.)

pawosm-arm (approved):
suggested squashing the commits on merge

bryanpkc:
Thank you very much for your contribution, @JiaweiHawk!

JiaweiHawk:
> Thank you very much for your contribution, @JiaweiHawk!

Thanks @bryanpkc and all the other reviewers for their help in getting this patchset finally merged!

Issue #1336, opened 2022-12-06
Unexpected result when the function with BIND(C) attribute has ENTRY statements.


xinliu-hnc:
Consider following Fortran code:
```
module m
contains
function f1() bind(C)
  complex :: f1
  integer :: e1
  f1 = (2.0, 0.0)
  return
entry e1() bind(C)
  e1 = 2
end function
end module

program test
  use m
  print *, f1(),e1()
end program
```
the result of flang and gfortran:
```
$ flang test.f90
$ ./a.out
(2.000000,0.000000) -993867848

$ gfortran test.f90
$ ./a.out
(2.00000000,0.00000000) 2
```
As shown in the LLVM IR below, `__master_entry_rslt355` stores the address of the return value of the entry instead of the expected return value of the entry, which results in an incorrect result.
```
define void @m__master___f1_(i32 %__master_entry_choice354.arg, i64* %__master_entry_rslt355) {
...
%10 = bitcast i64* %__master_entry_rslt355 to i32**
store i32* %e1_333, i32** %10, align 4
ret void, !dbg !5
...
}

define i32 @m_e1() {
...
%0 = bitcast i32* %e1_333 to i64*
call void @m__master___f1_(i32 1, i64* %0)
%1 = load i32, i32* %e1_333, align 4
ret i32 %1
}
```

shivaramaarao:
Hi,

I am seeing errors while compiling this program. Are you using any specific flags to compile the program?

flang entry.F90
/tmp/entry-cedfb3.ll:81:6: error: value doesn't match function result type 'double'
        ret <{float, float}> %0
            ^
1 error generated.

If not, let me know the revision of flang that you are using.

xinliu-hnc:
> Hi,
> 
> I am seeing errors while compiling this program. Are you using any specific flags to compile the program?
> 
> flang entry.F90 /tmp/entry-cedfb3.ll:81:6: error: value doesn't match function result type 'double' ret <{float, float}> %0 ^ 1 error generated.
> 
> If not, let me know the revision of flang that you are using.

I compiled the above program on **AArch64** before and didn't use any specific flags. I just tested the latest version(commit: f78c4497) of the `master` branch on both AArch64 and X86, and there is no error on AArch64 but there is the error you mentioned on X86.

PR #1362, opened 2023-04-23
Fix issue 1336. incorrect results with entry statements

shivaramaarao:
Fixing the issue 1336.
currently the address of entry return value is stored when the entry return type is pointer. this gives incorrect result.
The fix addresses the issue by storing the scalar return value to the entry return value.

### DIFF
--- a/test/f90_correct/inc/en06.mk
+++ b/test/f90_correct/inc/en06.mk
@@ -1,0 +1,23 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+$(TEST): run
+
+
+build:  $(SRC)/$(TEST).f90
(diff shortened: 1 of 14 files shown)


pawosm-arm (changes requested):
The patch itself looks OK, but it raises at least one question: if it required a change in the x86 ABI, should similar change be done to other architectures too? Could it be notified in a short statement in the commit's message? Also, somewhat more descriptive commit message would be helpful, e.g. what was failing before this patch and why.

pawosm-arm (approved):

pawosm-arm:
Hi @shivaramaarao 
I've tested your change on AArch64 machine and it helped with the problem signalled in the issue #1336.

A small typographic comment: While applying the patch, I've noticed that you're mixing spaces and tabs. Would it be possible to use only the spaces instead?

shivaramaarao:
> Hi @shivaramaarao I've tested your change on AArch64 machine and it helped with the problem signalled in the issue #1336.
> 
> A small typographic comment: While applying the patch, I've noticed that you're mixing spaces and tabs. Would it be possible to use only the spaces instead?

Incorporated the changes.

pawosm-arm:
What this PR needs is squashing the commits into one and force-push to get rid of the old history with all of those merge commits.

bryanpkc (approved):

PR #1368, opened 2023-06-01
[Flang1] Fix the string operation of file name

PeixinQiao:
This bug came from 0bcbadbf3f7f056b298a7724b9689ebacbdae4e9, which uses strdup to allocate non-const versions of const string, but does not change the use of the original string. The use of original string (file name) terminates the string by removing the last directory. So, we need to replace the use of original string with the allocated string.

### DIFF
--- a/test/f90_correct/makefile
+++ b/test/f90_correct/makefile
@@ -24,6 +24,8 @@ COMP_CHECK=python $(HOMEQA)/../tools/check_compilation.py
 
 RM=rm -f
 CP=cp -f
+CD=cd
+MKDIR=mkdir -p
 UNAME := $(shell uname -a)
 
 INCLUDES=$(BASE_DIR)/inc
--- a/test/f90_correct/src/import_mod_from_user.f90
+++ b/test/f90_correct/src/import_mod_from_user.f90
@@ -1,0 +1,17 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Test for the import module from the correct path.
+
+module import_mod_from_user
+contains
+  subroutine sub()
+     use import_source2
+  end
+end
+
+program import_mod_from_user_test
+  print *, "PASS"
+end
--- a/test/f90_correct/src/import_mod_from_user_dir1_m2.f90
+++ b/test/f90_correct/src/import_mod_from_user_dir1_m2.f90
@@ -1,0 +1,20 @@
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Test for the import module from the correct path. This file is in the wrong
+! path for testing.
+
+module import_source
+contains
+  subroutine sub1()
+  end
+
+  subroutine sub3()
+  end
+
+  function func() result(funit)
+    integer :: funit
+  end
+end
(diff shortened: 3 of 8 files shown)


PeixinQiao:
Simple reproduce:
```
$ cat shared/m2.f90 
module import_source
contains
  subroutine sub1()
  end

  subroutine sub3()
  end

  function func() result(funit)
    integer :: funit
  end
end
$ cat dir1/m1.f90 
module import_source2
    use import_source, only : func, sub1
end
$ cat dir1/m2.f90
module import_source
contains
  subroutine sub1()
  end

  subroutine sub2(m)
    integer, intent(out) :: m
  end

  function func() result(funit)
    integer :: funit
  end
end
$ cat test.f90
module import_mod_from_user
contains
  subroutine sub()
     use import_source2
  end
end

  print *, "PASS"
end
$ cat track.sh
#!/bin/bash

cd shared
flang m2.f90 -c
cd ../dir1
flang m2.f90 -c
flang m1.f90 -c
cd ..
flang test.f90 -I./shared -I./dir1 
$ bash track.sh 
$ 

```

pawosm-arm (approved):
Are you aware of any other places like this one? I'm surprised this didn't hurt us thus far.

test/f90_correct/makefile
```diff
@@ -24,6 +24,8 @@ COMP_CHECK=python $(HOMEQA)/../tools/check_compilation.py
 
 RM=rm -f
 CP=cp -f
+CD=cd
+MKDIR=mkdir
```
bryanpkc, line 28: I suggest `mkdir -p`.

test/f90_correct/src/import_mod_from_user.f90
```diff
@@ -0,0 +1,16 @@
...
+
+module import_mod_from_user
+contains
+  subroutine sub()
+     use import_source2
+  end
+end
+
+  print *, "PASS"
+end
```
bryanpkc, line 16: This looks like a dangling `end`.

test/f90_correct/src/import_mod_from_user_dir1_m2.f90
```diff
@@ -0,0 +1,12 @@
+module import_source
```
bryanpkc, line 1: Missing standard copyright template, and a comment about what the file is part of.

bryanpkc:
> Actually I am confused about the previous changes using `strdup`. If there is the compile error when changing the argument from `char *` to `const char *`, then it should not be changed, right? @bryanpkc

The propagation of `const`-ness everywhere was necessitated by the use of the two non-`static` functions in this file (`import_host` and `import_host_subprogram`) by `semutil2.c`, which passes constant string literals to these calls. We did not try to understand why literals were used in those calls, but even if we were able to refactor the calls to not use string literals, I believe it was a good idea to preserve `const`-ness where possible.

test/f90_correct/makefile
PeixinQiao: Good suggestion. Fixed.

test/f90_correct/src/import_mod_from_user.f90
PeixinQiao: Right. It is for testing for execution. Is there any other way in classic-flang test tools?

test/f90_correct/src/import_mod_from_user_dir1_m2.f90
PeixinQiao: Oh, I missed that. Fixed. Thanks.

PeixinQiao:
> > Actually I am confused about the previous changes using `strdup`. If there is the compile error when changing the argument from `char *` to `const char *`, then it should not be changed, right? @bryanpkc
> 
> The propagation of `const`-ness everywhere was necessitated by the use of the two non-`static` functions in this file (`import_host` and `import_host_subprogram`) by `semutil2.c`, which passes constant string literals to these functions. We did not try to understand why literals were used in those calls, but even if we were able to refactor the calls to not use string literals, I believe it was a good idea to preserve `const`-ness where possible.

OK. Got it. It's reasonable.

PeixinQiao:
gentle ping @bryanpkc @shivaramaarao

bryanpkc (approved):
LGTM.

test/f90_correct/src/import_mod_from_user.f90
PeixinQiao: Fixed. Adding the program name truly looks more beautiful.

shivaramaarao (approved):

PR #1369, opened 2023-06-05
Enable NCAR kernel tests

bryanpkc:
This PR fixes a couple of failures in the NCAR tests on X86, adds a new Makefile goal named `check-flang-long`, and updates the GitHub CI workflow to run that instead of `check-all`.  `check-all` remains synonymous to `check-flang`, which skips the NCAR kernel tests as `UNSUPPORTED`.

### DIFF
--- a/.github/workflows/build_flang.yml
+++ b/.github/workflows/build_flang.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Test flang
         run: |
           cd build/flang
-          make check-all
+          make check-flang-long
 
       # Archive documentation just once, for the fastest job.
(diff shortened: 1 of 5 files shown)


pawosm-arm (approved):
LGTM,

N.B. I assume increasing _TOL wasn't an option here?

bryanpkc:
> N.B. I assume increasing _TOL wasn't an option here?

No. On X86, the difference between the actual result and the reference result is literally 0.0, and `ABS(0.0) .LT. 0.0` evaluates to `.FALSE.`. Conceptually, `_TOL` is an unsigned value and the code would be more confusing if we made it negative.

shivaramaarao (approved):

PR #1370, opened 2023-06-06
Add Windows support to openmp_examples tests

tdusnoki:


bryanpkc (approved):

test/openmp_examples/lit/Example_SIMD.1f.sh
```diff
@@ -5,5 +5,5 @@
 
 # Shared lit script for each tests. Run bash commands that run tests with make.
 
-# RUN: KEEP_FILES=%keep FLAGS=%flags TEST_SRC=%s MAKE_FILE_DIR=%S/.. bash %S/runmake | tee %t 
+# RUN: env KEEP_FILES=%keep FLAGS=%flags TEST_SRC=%/s MAKE_FILE_DIR=%/S/.. bash %/S/runmake | tee %/t
```
pawosm-arm, line 8: what are those `%/`?

bryanpkc: From https://llvm.org/docs/CommandGuide/lit.html#substitutions :

pawosm-arm: I presume this couldn't be done any nicer way...

pawosm-arm (approved):

test/openmp_examples/lit/Example_SIMD.1f.sh
kaadam: Yeah,  mingw64 bash is a bit picky about path. Unfortunately, We haven't found other way how we can force Posix slashes on Windows. that's why we decided to use predefined lit substitution. Despite it results really huge changes in tests.

PR #1379, opened 2023-06-23
Add docker container for building flang

dmikushin:
This PR provides a clean Docker environment to build Flang from source. No assumptions, no complicated scripting, no extra parameters to provide. This gives everyone an easy way to build Flang, without asking questions and searching for proper instructions.

First, it builds the patched LLVM branch of choice, and then builds the Flang itself.


bryanpkc:
@dmikushin Now that you are building things correctly, can we close this PR?

dmikushin:
Hi @bryanpkc , @pawosm-arm Actually this is the most important on-topic PR, because it builds the whole Flang without depending on CI. With this Docker, any user could build an own version of Flang independently.

I will pull few more changes here to make it fully aligned with the correct Flang building method.

Then I will ask you to merge it and advertise as a preferred "build from source" method. Then I will ask you to start offering Flang releases based on it.

dmikushin:
# Build Flang from source in Docker container

[Flang](https://github.com/flang-compiler/flang) (also known as "Classic Flang") is an out-of-tree Fortran compiler targeting LLVM.

This builds the Flang compiler in Ubuntu 22.04 Docker container, according to the instructions given on GitHub, plus OpenMP offload features for AMD and NVIDIA GPUs.

Classic Flang build must follow the following sequence.

First, the [Classic Flang LLVM Project](https://github.com/flang-compiler/classic-flang-llvm-project) must be build with `-DLLVM_ENABLE_CLASSIC_FLANG=ON` and `-DLLVM_ENABLE_PROJECTS="clang"` options. These options and patches to the mainline LLVM applied in [Classic Flang LLVM Project](https://github.com/flang-compiler/classic-flang-llvm-project) will prepare a toolchain that could incorporate Classic Flang, when it is built. Specifically, options and patches add support for legacy PGI options such as `-Mreentrant` and call the `flang1` and `flang2` stages, when Fortran compiler is invoked. Note: this scenario completely replaces ("predates") the official [LLVM Flang](https://flang.llvm.org/docs/): it can't be enabled and used together with Classic Flang.

Second, the [Flang](https://github.com/flang-compiler/flang) project must be built with the LLVM toolchain prepared above. CMake configuration must include `flang` as a Fortran compiler, which refers to the Classic Flang frontend enabled by the patches. This Fortran compiler is not yet usable, so we also provide `-DCMAKE_Fortran_COMPILER_ID=Flang`, which instructs CMake to assume Fortran compiler works without checking it. During the build, `flang1` and `flang2` executables will be created to get a working Flang frontend, and then this frontend will be deployed to compile runtime libraries in Fortran language.

Please find more details on the build process in [classic-flang-llvm-project.sh](classic-flang-llvm-project/classic-flang-llvm-project.sh) and [classic-flang.sh](classic-flang/classic-flang.sh).

## Building

Build Flang with the following command:

```
docker-compose up --build
```

Note that the command above will store source files, build intermediates and installation folders as Docker mounted volumes. This method greatly speeds up partial rebuilds, but may require you to manually clean the build folders, if you want to try something completely different (e.g. a different LLVM branch).

The status of build is not printed right into the current terminal window. In order to watch the build log, run the following command in another terminal:

```
docker-compose logs -f
```

## Using distcc

This build supports distributed parallel build with distcc as an extra option.

The distcc hosts shall be provided on ports of a centralized gateway server with an SSH access to it. Gateway configuration is not covered by this instruction and should be provided by the user, e.g. by setting up SSH port forwarding.

1. Create SSH keys in the docker folder:

```
ssh-keygen -t rsa -b 4096 -C "autossh" -f id_rsa
```

2. Create a gateway server account, and use `id_rsa.pub` as an authorized key for it

3. Define distcc host in .env, for example:

```
DISTCC_HOSTS="localhost/8 tunnels:3632/8,lzo,cpp"
```

4. Build Flang as usual and enjoy the distributed compilation being deployed!


Timing of distributed compilation on 4 machines with 8, 8, 8 and 4 cores:

```
time docker-compose up --build
________________________________________________________
Executed in   22.79 mins    fish           external
   usr time   51.68 secs  836.00 micros   51.68 secs
   sys time    2.17 secs  203.00 micros    2.17 secs
```

The same compilation on a local machine with 8 cores:

```
time docker-compose up --build
________________________________________________________
Executed in   53.55 mins    fish           external
   usr time  127.28 secs   26.23 millis  127.25 secs
   sys time    5.02 secs    0.35 millis    5.02 secs
```

## Debugging

By default, LLVM and Flang are built in Release mode. In order to build in debug mode, run `docker-compose up --build` after adding `CMAKE_BUILD_TYPE=Debug` into the `.env` file.

bryanpkc:
I can imagine how a Dockerfile like this could be useful for newcomers to Classic Flang, but I am wary of the lack of maintenance going foward. In particular, if we want to make it release-worthy, we would have to be careful about what base image to use, and what build- and run-time dependencies we choose to include.

On the other hand, if we improve our build instructions to recommend using `build-llvm-project.py` and `build-flang.py` across all platforms, would that make things sufficiently easy that we don't need to add and maintain a Dockerfile?

dmikushin:
@bryanpkc "Let's not have any build system and not have any downloads at all, just because we cannot support them enough" is a very suspicious argument. I'm under impression that you want to intentionally impose difficulties for people to build or download and use Flang. Your reason could be that your clients should be using Huawei's compilers. So you are in a conflict of interest. This matter should be justified by community members who could neutrally recognize the need of builds and releases independent of any company.

bryanpkc:
@dmikushin

> "Let's not have any build system...

I don't know how you could have inferred that I don't want a build system. In fact, we have two build systems (one using bash scripts, one using Python scripts), and they both work. We have GitHub CI workflows that test the code. We discussed binary releases but concluded that there wasn't enough interest to make them worthwhile. We can revisit these discussions if the situation has changed...

> This matter should be justified by community members...

Would you like to attend the community call tomorrow to discuss this in person?

bryanpkc:
> Would you like to attend the community call tomorrow to discuss this in person?

Join the Flang Compiler Slack; there is a `#classic-flang` channel where the meeting link is pinned. Currently, the call is scheduled at 10:30am EST every other week.

dmikushin:
My point is:

1) There are no releases of Flang to download
2) Nobody inexperienced could follow the existing instructions

These two  points make good and very robust Flang compiler sufficiently unusable for almost everyone.

This project exists for many years without a single release. However, we know that multiple companies make regular releases based on it. It's very easy to make 2+2 of these two facts and find out why there is no release here.

bryanpkc:
Earlier in the project's history (before my time), volunteers had helped put together binary releases in downstream distributions such as Spack and [Debian](https://packages.qa.debian.org/f/flang.html). They eventually stopped supporting Classic Flang (e.g. see spack/spack#19350) as interest in the project waned and momentum grew behind the new LLVM Flang development. The reality is that Classic Flang is provided on an as-is basis, and everybody has to make one's own decision on how much time one  can volunteer towards improving it.

dmikushin:
Thank you for this interesting information. I have not heard of Spack before. From what I see Spack is run by national labs affiliates, not by "volunteers" per se. This is an honorable, but a different thing: they do it for the internal HPC park, which is huge, but not a public community.

dmikushin:
@bryanpkc , it is not possible to sign into [https://flang-compiler.slack.com](https://flang-compiler.slack.com/messages), it refuses to log me in. Perhaps I need an invite to join the channel?

bryanpkc:
Since we already provide bash and Python scripts that can build Classic Flang fine on a typical Ubuntu system, we would prefer to avoid the burden of maintaining an unnecessary Docker image. Closing the PR.

bryanpkc:
I added a paragraph in [the wiki page](https://github.com/flang-compiler/flang/wiki/Building-Flang) to describe how to install the build tools on a typical Ubuntu system. Hope that helps.

Issue #1374, opened 2023-06-21
implicit truncation from 'int' to a one-bit wide bit-field changes value from 1 to -1


dmikushin:
This is a compiler error during building with clang-16:

```
classic-flang                 | /classic-flang/src/classic-flang/tools/flang1/flang1exe/commopt.c:389:23: error: implicit truncation from 'int' to a one-bit wide bit-field changes value from 1 to -1 [-Werror,-Wsingle-bit-bitfield-constant-conversion]
classic-flang                 |         FT_FUSED(nd1) = 1;
classic-flang                 |                       ^ ~
classic-flang                 | /classic-flang/src/classic-flang/tools/flang1/flang1exe/commopt.c:391:22: error: implicit truncation from 'int' to a one-bit wide bit-field changes value from 1 to -1 [-Werror,-Wsingle-bit-bitfield-constant-conversion]
classic-flang                 |         FT_FUSED(nd) = 1;
classic-flang                 |                      ^ ~
```

bryanpkc:
Classic Flang does not support LLVM 16 yet. Please continue to use the `release_15x` branch of classic-flang-llvm-project. A working `release_16x` branch will be created very soon.

PR #1381, opened 2023-06-25
[LLVM16][flang1] Suppress -Wsingle-bit-bitfield-constant-conversion warnings

bryanpkc:
Some single-bit bitfields are declared as `int`, and Clang 16 now emits warnings for such uses, e.g.

    .../flang1exe/commopt.c:389:23: error: implicit truncation from 'int' to a one-bit wide bit-field changes value from 1 to -1

Such bitfields should be declared `unsigned` instead.


dmikushin:
This fixes https://github.com/flang-compiler/flang/issues/1374

PR #1383, opened 2023-06-26
Use preprocessor in traditional mode

dmikushin:
Fortran compiler is supposed to use traditional C preprocessor only. Therefore, we can't use modern C preprocessor features, such as catenation or stringizing


bryanpkc:
gfortran's preprocessor does not support stringization. There may be a historical reason for that, but we should be able to provide improved preprocessing functionality. LLVM Flang's preprocessor supports stringization (see https://github.com/llvm/llvm-project/blob/main/flang/test/Preprocessing/pp045.F, https://github.com/llvm/llvm-project/blob/main/flang/test/Preprocessing/pp131.F), so I don't think we need to change anything in Classic Flang.

bryanpkc:
This PR will not be accepted as-is, so closing it for now.

PR #1389, opened 2023-06-26
Re-arranging the code such that ili is always initialized and used

dmikushin:


dmikushin:
@pawosm-arm , I don't have the exact error message with me, but it's a usual "use without initialization" error, which fails the build in -Werror mode with clang-16.

pawosm-arm (approved):

PR #1391, opened 2023-06-26
The array dimensions must be defined *before* being used

dmikushin:
It's a pity that for someone the order of arguments takes precedence over this obvious requirement


bryanpkc:
Thanks for fixing the problem. The original code happens to work correctly thanks to implicit typing, but the corrected form looks better and will silence warnings from compilers that enforce newer language standards.

Could you please follow LLVM conventions and add a topic prefix (e.g. "[flang1]", "[runtime]", etc.) to the commit title?

Also, would you mind removing your commentary from commit messages? This code probably predated the Fortran 2008 standard, when the rules for declaration order in specification expressions were made explicit. There is no need to insult the original author of the code.

dmikushin:
This and my other comments express how deeply saddened I am by the poor state of this project. We deal now with the result of years of destructive management and attempts to raise as much money as possible with the least effort. Perhaps you are not aware that a single license for this "product" was once sold at a price of $900.

And even now that we have the opportunity to bring this legacy nightmare into good quality state, it is being run by companies that use it as a spoiler in their hardware sales, develop their closed-source branches, and are not interested in developing anything other than their own successful deals.

I would like to see the releases of this project as a vendor-independent product. Releases should be easily accessible and users should be well informed of their availability. By supporting this effort you will get a lot of respect from the community, and there will be a lot less criticism in all aspects.

bryanpkc:
LLVM Flang is much better designed from the get-go, and most developers have already moved on to it, so unfortunately this "legacy nightmare" isn't going to get any more corporate attention than it is getting today.

We are doing our best in upstreaming useful features and important fixes, and we are well aware of the huge amount of technical debt we inherited, but our first priority is stability. A lot of the known issues require massive refactoring to fix, and we cannot justify the time investment or the potential risk to stability.

We discussed community releases a long time ago, but we could not find any interest or volunteers, so we decided not to provide any.

bryanpkc (approved):

pawosm-arm (approved):

PR #1392, opened 2023-06-26
Adding intent(in)

dmikushin:
Planting intent(in) all over the code, per strict modern Fortran compilers requirements. Do not duplicate eq and == , ne and /= operators definitions, as they seem to be already aliases of each other, and compilers are not happy if we try to overdefine this fact.

### DIFF
--- a/runtime/flang/ieee_arithmetic.F95
+++ b/runtime/flang/ieee_arithmetic.F95
@@ -88,19 +88,7 @@ module IEEE_ARITHMETIC
     module procedure ieee_arithmetic_eqtct
   end interface
 
-  interface operator (==)
-    module procedure ieee_arithmetic_eqti
-    module procedure ieee_arithmetic_eqtr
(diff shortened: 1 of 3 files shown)


bryanpkc:
@dmikushin For reference, could you identify the code in `flang1` that aliases operators `.ne.` etc. to `/=` etc.? Also, could you attach the binary output of compliing these files with and without the `intent(in)`, to show that your change has no adverse impact on performance?

bryanpkc:
@dmikushin gentle ping

dmikushin:
`intent(in)` is an element of Fortran interfaces design rules, such as iso_c_bindings. AFAIK, it does not change the compilation result, and focuses entirely on mandatory understanding by the developer the code being developed. The requirements used by PGI/Intel are in fact more liberal than the actual Fortran standard. So the main point of this PR is that its makes the code more standard-compliant, NFC.

Issue #1363, opened 2023-04-26
Running Flang tests on Windows


kaadam:
The goal is to add a basic Windows support to Flang tests. My approach is to adopting the current test environment to Windows, try to preserve the current test process [lit->sh->make->FileCheck] how they are launched as much as possible by using gnuwin32 tools. As far as I see this is the simplest way to find a general solution in a short time.

I managed to determine some issues which need to be fixed in order to run Flang tests on Windows:

1) Defining temporary "environment" variable for forked process:
The following logic how these variables are defined and pass them to the forked bash process doesn't work in PowerShell or Command Promt.  
` RUN: KEEP_FILES=%keep FLAGS=%flags TEST_SRC=%s MAKE_FILE_DIR=%S/.. bash %S/runmake | tee %t  `
Instead we should use this format:
`RUN: bash -c "KEEP_FILES=%keep FLAGS=%flags TEST_SRC=%:s MAKE_FILE_DIR=%:S/.. %:S/runmake | tee %:t" `
It would be good, if we could use a common sh file for Linux and Windows. But maybe it is more clear if we create a separate bat/ps1 file for each tests.

2) Using proper path separator for gnuwin32 bash.
We should use Linux style path '/c/path/to//test/dir' instead of Windows path. Maybe we should override %:S behaviour which removes : from the path (this is good for us) , but also removes / single leading by default.

3) Detecting Clang-cl and Flang compiler.
On Linux by default classic flang llvm tools are installed into /usr/local and the make easily finds them.
On Windows it needs to tell explicitly to the make where the Flang compiler is.
It could be done if we pass the path to llvm tools dir through the script chain, or generating a 'config.mk' file with cmake which will be included in main Makefile. Or add the 'llvm_tools_dir' to the PATH before the tests execution.

4) Linking issue
It seems it needs to fix the how Flang pass the flang runtime libraries to MS Linker.

```
InstalledDir: C:\flang_work_dir\classic-flang-llvm-project\build_15\.\bin "C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.34.31933\\bin\\Hostx64\\x64\\link.exe" -out:ab10.exe "-libpath:C:\\flang_work_dir\\classic-flang-llvm-project\\build_15\\lib" -lflang -lflangrti -lpgmath -lpthread -lrt /subsystem:console -nologo ab10.o check.o
Microsoft (R) Incremental Linker Version 14.34.31937.0
[...]

PR #1396, opened 2023-06-27
Fix llvm_ir_correct tests on Windows

tdusnoki:
On Windows some tests are unable to open iso_fortran_env.mod unless we include it explicitly.

related issue: #1363

### DIFF
--- a/test/llvm_ir_correct/iso_fortran_env_functions.f90
+++ b/test/llvm_ir_correct/iso_fortran_env_functions.f90
@@ -5,7 +5,7 @@
 !
 ! Simply check the file compiles with no error.
 !
-! RUN: %flang -S -emit-llvm %s -o - | FileCheck %s
+! RUN: %flang -I%flags -S -emit-llvm %s -o - | FileCheck %s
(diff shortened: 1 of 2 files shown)


bryanpkc (changes requested):
`%flags` is for specifying custom options, and typically undefined. You need to find another way to solve this problem. Why can't the compiler locate `iso_fortran_env.mod`? It should be in the `include/` directory of the Classic Flang installation, which should be one of the standard include paths.

kaadam:
> `%flags` is for specifying custom options, and typically undefined. You need to find another way to solve this problem. Why can't the compiler locate `iso_fortran_env.mod`? It should be in the `include/` directory of the Classic Flang installation, which should be one of the standard include paths.

Yes, you're right. The 'AddFlangSystemIncludeArgs' implementation is missed from MSVC toolchain, please see: flang-compiler/classic-flang-llvm-project#161

tdusnoki:
After testing with flang-compiler/classic-flang-llvm-project/pull/161 this PR is no longer necessary, thanks for the feedback.

PR #1400, opened 2023-06-28
Do not dump debug location if debug info is not available

dmikushin:
This works around the crash due to the NULL value of `module->debug_info`.

### DIFF
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -3519,12 +3519,14 @@ write_instructions(LL_Module *module)
     }
     /*
      *  Do not dump debug location here if
+     *  - debug info is not available
      *  - it is NULL
      *  - it is already written (dbg_line_op_written) or
      *  - it is a known internal (f90 runtime) call in prolog (fort_init &
      * f90_*)
      */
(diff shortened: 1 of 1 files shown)


bryanpkc:
Can you provide a test case or a simple reproducer? I suspect that we haven't seen this problem previously, thanks to the short-circuit evaluation of `LL_MDREF_IS_NULL(instrs->dbg_line_op)`. I'm curious how `instrs->dbg_line_op` is non-null when no `-g` flag is specified on the `flang` command line.

dmikushin:
I'll try to remember the details. For now I can tell that the crash occurred when testing openmp target offloading. Maybe code generation for offloading goes with some different options?

pawosm-arm (approved):
Doesn't hurt, so I'm fine with it.

If a test case can't be worked out, a commit message could be extended at least, with the description how this problem was encountered.

bryanpkc:
@dmikushin gentle ping

dmikushin:
I will try to reproduce it again. This happens when I try to perform openmp offloading. It has some code extraction mechanics that the loop goes into a separate compilation unit, but the debug info remains attached to the main object, so the offloaded object does not inherit it, hence is the unexpectedly missing pointer value.

So for the current state of Flang, this scenario is purely teoretical, as long as openmp offload support is not officially endorsed.

PR #1401, opened 2023-06-28
Link argument parsing functions from libraries, instead of duplicating their source files

dmikushin:
This PR basically removes

```diff
-set(SHARED_SOURCES
-  ${SHARED_SOURCES}
-  ${LIB_SOURCE_DIR}/ADT/hash.c
-  ${LIB_SOURCE_DIR}/ArgParser/arg_parser.c
-  ${LIB_SOURCE_DIR}/ArgParser/debug_action.c
-  ${LIB_SOURCE_DIR}/ArgParser/xflag.c
-  )
```

and changes a few more things accordingly. 

The idea of `SHARED_SOURCES` is to have the same source files compiled with different macro definitions. For this reason, the source files in SHARED_SOURCES cannot be shipped normally, as a static library.

But the purpose above does not apply to the argument parsing functions - they are universal. Therefore, this PR is taking them out of the SHARED_SOURCES.


dmikushin:
Need something extra on Windows:

```
lld-link: error: undefined symbol: flg
lld-link: error: undefined symbol: gbl
lld-link: error: undefined symbol: version
```

tools/flang1/flang1exe/gbldefs.h
```diff
@@ -143,13 +143,8 @@ typedef int LOGICAL;
...
 
-/* mall.c */
-char *sccalloc(BIGUINT64);
-void sccfree(char *);
-char *sccrelal(char *, BIGUINT64);
-#ifdef DEBUG
-void bjunk(void *p, BIGUINT64 n);
-#endif
+#include "mall.h"
+#include "listing.h"
```
bryanpkc, line 147: Can we avoid including these headers within `gbldefs.h`, and instead include them in the `.c` or `.cpp` files that actually need the function declarations?

tools/flang2/flang2exe/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ set(SOURCES
   dtypeutl.cpp
   fenddf.cpp
   ilmutil.cpp
-  listing.cpp
+  listing.c
```
bryanpkc, line 22: This rename doesn't seem necessary at this point. In a future PR, perhaps we could move `listing.c` and `listing.cpp` from under `tools/flang1/flang1exe/` and `tools/flang2/flang2exe/`, and merge them into one file under `tools/shared/`.

bryanpkc:
Will you fix the Windows build failure as well?

tools/flang2/flang2exe/CMakeLists.txt
dmikushin: I think I did this to come up with a proper name mangling and resolve link errors.

dmikushin:
I don't have a Windows builder yet, but I'm looking forward to getting it!

Currently, we are still discussing with you enablement of Linux builds through Docker. How could I encourage you to have them, or accept mine? This is related, because I'm not able to use the CI and develop in it.

pawosm-arm (commented):
I wouldn't do that. The original authors wanted to decouple these two as so they wouldn't be stopped with improving flang2 by the anchors deeply rooted in flang1 (and e.g. making flang2 compiled with a C++ compiler (versus C compiler) was one of the means for achieving such improvement). Some duplication was inevitable. This commit goes against that tide and restores the coupling.

bryanpkc:
Closing the PR as it will not be accepted as-is.

PR #1403, opened 2023-07-03
[LLVM16][build] Specify LLVM_MAIN_SRC_DIR when running build-flang.sh

bryanpkc:
Otherwise Classic Flang's `CMakeLists.txt` will be unable to find `llvm-lit` and cannot generate test-related goals, e.g. `check-all`.

Also fix a typo in `build-flang.sh` that prevented the option from being passed to CMake correctly.

Also add support for the same variable to libpgmath's CMakeLists.txt, as well as `scripts/build_flang.py`.

Also modify helpers in `runtime/libpgmath/cmake/modules/LibmathUtils.cmake`, e.g. `libpgmath_say`, `libpgmath_warning_say`, to print all given arguments instead of only the first named argument. This allows callers to break long strings into multiple lines, resulting in better-formatted code.


.github/workflows/build_flang_arm64.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       - name: Build and install flang & libpgmath
         run: |
           cd ${{ env.build_path }}/flang
-          ./build-flang.sh -t ${{ matrix.target }} -p ${{ env.install_prefix }} -n `nproc --ignore=1` -v
+          ./build-flang.sh -t ${{ matrix.target }} -p ${{ env.install_prefix }} -n `nproc --ignore=1` -v -l /home/root/classic-flang-llvm-project/llvm
```
pawosm-arm, line 57: `/home/root`?

pawosm-arm (changes requested):

.github/workflows/build_flang_arm64.yml
bryanpkc: Yes, `/home/root` is the location of the latest `classic-flang-llvm-project` source tree for the AArch64 build job. [It is hardcoded in the Docker image.](https://github.com/flang-compiler/classic-flang-llvm-project/blob/release_15x/.github/workflows/Ubuntu20Dockerfile#L57)

pawosm-arm (approved):

bryanpkc:
@pawosm-arm I added a couple more commits. Do they look okay to you?

runtime/libpgmath/cmake/modules/LibmathUtils.cmake
```diff
@@ -1,17 +1,17 @@
-macro(libpgmath_say message_to_user)
-  message(STATUS "LIBPGMATH: ${message_to_user}")
+macro(libpgmath_say)
```
pawosm-arm, line 1: Nice change there, but I'd rather see it mentioned in the commit message or moved to another commit.

bryanpkc: I have updated the commit message as you suggested. Thanks.

pawosm-arm:
> @pawosm-arm I added a couple more commits. Do they look okay to you?

Yes, they do

bryanpkc:
Merging CI-only PR with one approval.

PR #1404, opened 2023-07-11
Fix for handling absolute include paths

pawosm-arm:
A silly bug that prevented building OpenMPI OOT.

### DIFF
--- a/tools/flang1/flang1exe/accpp.c
+++ b/tools/flang1/flang1exe/accpp.c
@@ -2103,6 +2103,7 @@ add_to_incllist(char *fullname)
 static void
 _doincl(char *name, int type, LOGICAL include_next)
 {
+  FILE *tmpfp;
   char fullname[MAX_PATHNAME_LEN];
   int i;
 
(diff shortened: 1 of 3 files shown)


tools/flang1/flang1exe/accpp.c
```diff
@@ -2130,6 +2131,16 @@ _doincl(char *name, int type, LOGICAL include_next)
       goto found;
     }
 
+  if (type == 0) { /* could be absolute path, check where it leads to */
+    tmpfp = fopen(name, "r");
+    if (tmpfp) {
+      snprintf(fullname, MAX_PATHNAME_LEN, "%s", name);
+      idir.last = 0;
+      if (fclose(tmpfp) == 0)
```
bryanpkc, line 2139: Using `fopen` and `fclose` here makes the preprocessor behave different for this specific case. In the other cases, `fndpath` calls `access` to determine if an included file exists. Shall we do the same here?

bryanpkc (approved):

PR #1405, opened 2023-07-13
Add Windows support to lit.cfg

kaadam:
* Enable running Flang tests on Windows

### DIFF
--- a/test/lit.cfg
+++ b/test/lit.cfg
@@ -27,10 +27,6 @@ import lit.llvm
 lit.llvm.initialize(lit_config, config)
 lit.llvm.llvm_config.with_system_environment("COMPILER_PATH")
 
-if platform.system() == 'Windows':
-    lit_config.note('We do not support Windows, but hey, congratulations on porting to Windows!')
-    raise SystemExit
-
 # Configuration file for the 'lit' test runner.
 
 # name: The name of this test suite.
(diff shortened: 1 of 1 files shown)


bryanpkc (approved):
LGTM

shivaramaarao (approved):

test/lit.cfg
```diff
@@ -27,10 +27,6 @@ import lit.llvm
 lit.llvm.initialize(lit_config, config)
 lit.llvm.llvm_config.with_system_environment("COMPILER_PATH")
 
-if platform.system() == 'Windows':
```
pawosm-arm, line 30: The comments below refers to Msys (MinGW companion), what about Cygwin or VisualStudio, are we compatible with them so we can drop this condition?

kaadam: I've run tests from Windows environment (PowerShell, Dev Command Prompt) so basically the configuration itself is compatible with Windows /MS Visual Studio/. I didn't try it under Cygwin or MSYS envs. It should work. However we use several gnu tools to support Flang test running on Windows such 'bash, sh'  (from Git for Windows /mingw64/) or make and others from 'gnuwin32' package. 
Shall I add a reminder comment for that, or change the condition to exclude 'cygwin'?

pawosm-arm: A comment seems sufficient

kaadam: @pawosm-arm What do you think about the comment?

pawosm-arm: I'd rather put MS Visual Studio (full name), so it won't make confusion with MSVC or MSYS.

kaadam: okay, updated.

pawosm-arm (approved):

PR #1409, opened 2023-07-22
[runtime] Fix infinite loop in `__fort_bcopysl`

### LINKED ISSUE #1406: Regression caused by #PR1076
This test case failed to run, after PR1076 merged into flang.
```
program test
  type t
    integer :: a(8)
  end type
  type(t) :: stdm1(4, 5), stdm2(5, 4)
  integer, dimension (1:2) :: order1 = (/ 2, 1 /)
  type(t) :: array1(20)
  stdm1 = reshape(array1, (/ 4, 5 /), order = order1)
end
```
```
./a.out
Segmentation fault (core dumped)
```
PR1076 changes type of 'k' to size_t in runtime/flang/bcopys.c, which causes the following loop to become a infinite loop.
```
for (k = len - 1; k >= 0; k--) {
  to[i + k] = fr[j + k];
}
```
Thank you for your attention.

bryanpkc:
Fix a downward-counting loop with an unsigned loop counter to exit when the counter reaches 0 (instead of -1, which never happens).

Closes #1406.

### DIFF
--- a/runtime/flang/bcopys.c
+++ b/runtime/flang/bcopys.c
@@ -168,9 +168,12 @@ __fort_bcopysl(char *to, char *fr, size_t cnt, size_t tostr, size_t frstr,
     i *= len;
     j *= len;
     for (; cnt > 0; cnt--, i -= tostr, j -= frstr) {
-      for (k = len - 1; k >= 0; k--) {
+      // Since k is unsigned, stop the loop when k == 0, then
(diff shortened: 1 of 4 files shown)


pawosm-arm (approved):

bryanpkc:
gentle ping @shivaramaarao

shivaramaarao (approved):

PR #1414, opened 2023-08-08
Fix failures for imports with LLVM-17

alokkrsharma:
Symptom: Compilation of any Fortran program with imports (USE statement) fails with assert in compiler backend.

Diagnosis: LLVM-17 expects imports list in retainedNode field of DISubprogram wihch while Flang is still generating it as imports field of DICompilationUnit.

Fix: Flang is now updated to new IR as expected by LLVM-17.

### DIFF
--- a/tools/flang2/flang2exe/lldebug.cpp
+++ b/tools/flang2/flang2exe/lldebug.cpp
@@ -4115,7 +4115,7 @@ lldbg_function_end(LL_DebugInfo *db, int func)
 static LL_MDRef
 lldbg_create_imported_entity(LL_DebugInfo *db, SPTR entity_sptr, SPTR func_sptr,
                              IMPORT_TYPE entity_type, LL_MDRef elements_mdnode,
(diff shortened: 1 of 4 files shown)


tools/flang2/flang2exe/lldebug.cpp
```diff
@@ -4107,7 +4114,8 @@ lldbg_function_end(LL_DebugInfo *db, int func)
 
 static LL_MDRef
 lldbg_create_imported_entity(LL_DebugInfo *db, SPTR entity_sptr, SPTR func_sptr,
-                             IMPORT_TYPE entity_type, LL_MDRef elements_mdnode)
+                             IMPORT_TYPE entity_type, LL_MDRef elements_mdnode,
+                            LL_MDRef imported_list)
```
bryanpkc, line 4118: Nit: The parameter declaration is not left-aligned with the previous line.

bryanpkc (approved):
LGTM other than a nit.

PR #1415, opened 2023-08-18
Supplement to fix failures for imports with LLVM-17

Liuyunlong0336:
LLVM-17 expects imports list in retainedNode field of DISubprogram, and flang has completed this step but it's still generating imports list as imports field of DICompilationUnit.

Flang is now updated to new IR as expected by LLVM-17.

### DIFF
--- a/tools/flang2/flang2exe/ll_write.cpp
+++ b/tools/flang2/flang2exe/ll_write.cpp
@@ -935,7 +935,6 @@ static const MDTemplate Tmpl_DICompileUnit_ver39[] = {
   { "retainedTypes",            NodeField,               0 },
   { "globals",                  NodeField,               0 },
   { "emissionKind",             DWEmissionField,         0 },
-  { "imports",                  NodeField,               0 },
   { "splitDebugFilename",       StringField,             0 },
   { "nameTableKind",            DebugNameTableKindField, 0 }
 };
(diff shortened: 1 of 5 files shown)


bryanpkc:
@Liuyunlong0336 In your commit message you should probably mention that this is an update of the fix initially introduced in 08b46d71fc6301435f6e478208f8098df1714997, and explain why that commit couldn't catch the case you are fixing.

bryanpkc:
@alokkrsharma Does this look okay to you?

tools/flang2/flang2exe/ll_write.cpp
```diff
@@ -935,7 +935,6 @@ static const MDTemplate Tmpl_DICompileUnit_ver39[] = {
   { "retainedTypes",            NodeField,               0 },
   { "globals",                  NodeField,               0 },
   { "emissionKind",             DWEmissionField,         0 },
```
pawosm-arm, line 937: Shouldn't a new array `Tmpl_DICompileUnit_ver170` and `Tmpl_DICompileUnit_ver39` kept intact for backward compatibility? Although our inner policy is kind of 'no going back', others may not be so happy about it.

bryanpkc: I'm pretty sure that LLVM 17 still accepts `imports` as an optional field. There should be no need to delete this. It should be feasible for Classic Flang to generate this field as `imports: null` or `imports: !2` where `!2` is empty.

bryanpkc:
Upon further review, I don't think this PR is necessary. #1414 had fixed the problem properly.

PR #1416, opened 2023-09-05
Fix a sret attribute bug

Liuyunlong0336:
When the return value of the function is a structure, sometimes the compiler will use the structure as an input argument and add the sret attribute in IR generation stage. At this case, the IR generated by flang will lack the actual structure type.

### DIFF
--- a/test/llvm_ir_correct/test-sret-01.f90
+++ b/test/llvm_ir_correct/test-sret-01.f90
@@ -1,0 +1,39 @@
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM
+! Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Test for the scenario that when Fortran case calls the function of
+! C with a struct as the return value.
+
+! RUN: %flang -S -emit-flang-llvm %s -o %t
+! cat %t | FileCheck %s -check-prefix=CHECK-SRET
+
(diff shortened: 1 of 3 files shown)


bryanpkc:
Please reword your commit to provide a commit title on a separate line, with a topic prefix, e.g.
```
[flang2] Fix sret attribute in LLVM IR

When the return value of the function is a structure, sometimes the compiler
will use the structure as an input argument and add the sret attribute in IR
generation stage. In this case, the attribute generated by flang will lack
the actual structure type, which is required when opaque pointers are enabled.
```

test/llvm_ir_correct/test-sret-01.f90
```diff
@@ -0,0 +1,39 @@
...
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Test for the scenario that when Fortran case calls the function of
+! C with a struct as the return value.
+
+! RUN: %flang -S -emit-flang-llvm %s -o %t
+! cat %t | FileCheck %s -check-prefix=CHECK-SRET
+
+! CHECK-SRET: call void @c_function (ptr sret([[TYPENAME:%.*]]) {{%.*}}, ptr byval([[TYPENAME]]) {{%.*}})
```
bryanpkc, line 12: This won't work unless you also handle the `byval` attribute.

bryanpkc (changes requested):
This patch should also handle `byval`.

bryanpkc:
Replaced by #1420.

PR #1417, opened 2023-09-05
Fix an entry statement debuginfo bug

Liuyunlong0336:
When the compiler generates debuginfo, it does not consider that when there is an entry, the number of arguments will increase by one.
We can fix this problem by getting the returned information by process_ll_abi_func_ftn_mod function, and using the new number of arguments.

### DIFF
--- a/test/debug_info/entry_argument_1.f90
+++ b/test/debug_info/entry_argument_1.f90
@@ -1,0 +1,20 @@
+!RUN: %flang -g -S -emit-llvm %s -o - | FileCheck %s
+
+!CHECK: [[N1:![0-9]+]] = distinct !DISubprogram
+!CHECK: !DILocalVariable(arg: 1, scope: [[N1]]
+!CHECK: !DILocalVariable(arg: 2, scope: [[N1]]
+!CHECK: !DILocalVariable(name: "a", arg: 3, scope: [[N1]]
+
+module test
+contains
+  subroutine sub(a)
+    implicit none
+    integer(kind = 4) :: m
+    real(kind = 8), intent(inout) :: a(:,:)
+    m = size(a, 1)
+    entry subsub(a)
+    m = size(a, 1) + 1
+    entry subsub1(a)
+    m = size(a, 1) + 2
+  end subroutine sub
+end module
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -275,6 +275,7 @@ typedef struct ComplexResultList_t {
   unsigned entries;
 } ComplexResultList_t;
 static ComplexResultList_t complexResultList;
+LL_ABI_Info *entry_abi;
 
 /* ---  static prototypes (exported prototypes belong in cgllvm.h) --- */
 
@@ -1613,11 +1614,12 @@ restartConcur:
 
   /* Build up the additional items/dummys needed for the master sptr if there
      are entries, and call process_formal_arguments on that information. */
-  if (has_multiple_entries(gbl.currsub) && get_entries_argnum())
-    process_formal_arguments(
-        process_ll_abi_func_ftn_mod(current_module, get_master_sptr(), 1));
-  else
+  if (has_multiple_entries(gbl.currsub) && get_entries_argnum()) {
+    entry_abi = process_ll_abi_func_ftn_mod(current_module, get_master_sptr(), 1);
+    process_formal_arguments(entry_abi);
+  } else {
     process_formal_arguments(llvm_info.abi_info);
+  }
   made_return = false;
 
   get_local_overlap_size();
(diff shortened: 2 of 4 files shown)


test/debug_info/entry_argument_1.f90
```diff
@@ -0,0 +1,18 @@
...
+!CHECK: !DILocalVariable(name: "a", arg: 3, scope: [[N1]]
+
+module test
+contains
+  subroutine sub(a)
+    implicit none
+    integer(kind = 4) :: m
+    real(kind = 8),intent(inout) :: a(:,:)
+    m = size(a, 1)
+    entry subsub(a)
```
bryanpkc, line 15: What happens if there are more entries in the subroutine? Can you add another `entry` declaration and the corresponding `CHECK` pattern to this test case?

bryanpkc:
Please also reword the commit message. It should start with a commit title, with a topic prefix, and separated from the commit message body with a blank line. For example:
```
[DebugInfo] Fix bug in multi-entry procedures

When the compiler generates debuginfo for a procedure, it does not consider
that if is an ENTRY statement, the number of arguments will increase
by one. We can fix this problem by using the new number of arguments as
returned by the process_ll_abi_func_ftn_mod function.
```

test/debug_info/entry_argument_1.f90
Liuyunlong0336: For the entry feature, flang's processing logic is to turn an ordinary function into a function with parameter selection, and jump to which label (entry) according to the selection.
Therefore, adding one more entry will not affect the argument's metadata.

Liuyunlong0336: ``For entry declaration,the complier will turn an ordinary function into a function with branch, and jumping to which label (entry) is according to the choice argument.
Therefore, adding one more entry will not affect the argument's metadata, and that only to add branch label.
Like this:
`define void @test__master___test_sub__(i32 %__master_entry_choice430.arg, ptr %a, ptr %a$sd)`
......
`br i1 %3, label %Ltest_subsub_, label %L.LB2_432`
......
`br i1 %5, label %Ltest_subsub1_, label %L.LB2_433`

Liuyunlong0336: I added another entry declaration in test case

test/debug_info/entry_argument_1.f90
```diff
@@ -0,0 +1,20 @@
+!RUN: %flang -g -S -emit-llvm %s -o - | FileCheck %s
+
+!CHECK: [[N1:![0-9]+]] = distinct !DISubprogram
+!CHECK: !DILocalVariable(arg: 1, scope: [[N1]]
+!CHECK: !DILocalVariable(arg: 2, scope: [[N1]]
```
bryanpkc, line 5: I suggest changing the patterns to ensure that these two `DILocalVariable` nodes have the `DIFlagArtificial` flag.

test/debug_info/entry_argument_1.f90
bryanpkc: OK thanks.

bryanpkc (approved):
LGTM.

Liuyunlong0336 (commented):
I may not particularly understand.This test case generate ir as follows：
```
!22 = !DILocalVariable(arg: 1, scope: !6, file: !3, line: 10, type: !23)
!23 = !DIBasicType(name: "integer", size: 32, align: 32, encoding: DW_ATE_signed)
!24 = !DILocalVariable(arg: 2, scope: !6, file: !3, line: 10, type: !15)
!25 = !DILocalVariable(name: "a", arg: 3, scope: !6, file: !3, line: 10, type: !26)
```
There is not have DIFlagArtificial. Of course，`arg1:1`  correspond  `%__master_entry_choice430.addr` , it should have DIFlagArtificial. I think maybe this is another debuginfo bug?

test/debug_info/entry_argument_1.f90
bryanpkc: I was mistaking some other variables for the artificial arguments. What you have is fine.

bryanpkc:
@pawosm-arm @shivaramaarao gentle ping

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -1613,11 +1614,12 @@ schedule(void)
 
   /* Build up the additional items/dummys needed for the master sptr if there
      are entries, and call process_formal_arguments on that information. */
-  if (has_multiple_entries(gbl.currsub) && get_entries_argnum())
-    process_formal_arguments(
-        process_ll_abi_func_ftn_mod(current_module, get_master_sptr(), 1));
-  else
+  if (has_multiple_entries(gbl.currsub) && get_entries_argnum()) {
+    entry_abi = process_ll_abi_func_ftn_mod(current_module, get_master_sptr(), 1);
```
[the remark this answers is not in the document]
Liuyunlong0336: > any good reason why this had to be stored in a local variable?

Thank you very much， `entry_abi` is a global variables，The reason for storing it in such a variable is that when there is no entry, we follow this logic：
```
  } else {
    process_formal_arguments(llvm_info.abi_info);
  }
```
In `lldbg_emit_type` we will by `get_parnum(SPTR sptr)`, that is, `llvm_info.abi_info` to know which argument it corresponds.

When there is an entry attribute, this information is rewritten by `process_ll_abi_func_ftn_mod,` but is not stored in this variable. And it should be noted that the function `process_ll_abi_func_ftn_mod` does not seem to generate a complete `llvm_info.abi_info`. I dare not take this variable directly.

We need to get this information in `lldbg_emit_type`. According to the original logic of flang, this information is taken directly from `get_parnum`, that is, `llvm_info.abi_info`,  then we need a global variable to store it.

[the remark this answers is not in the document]
pawosm-arm: Ah yes, actually, you're right, I didn't look at the wider context.

pawosm-arm (approved):

shivaramaarao (approved):

PR #1418, opened 2023-09-06
Fix a min-max bug

Liuyunlong0336:
This patch solves two problems:

1.According to the fortran 2008 standard, add syntax check for MAX/MIN intrinsic, as follows:
The arguments shall all be of the same type which shall be integer, real, or character and they shall all have the same kind type parameter.

2.The final type of argument with MIN/MAX intrinsic is determined by the type of highest precision atgument, not the first argument.


tools/flang1/flang1exe/semfunc.c
```diff
@@ -4264,11 +4318,25 @@ ref_intrin(SST *stktop, ITEM *list)
...
       if (DTY(argdtype) == TY_ARRAY)
         break;
     } else {
+      if (intrin == I_MAX || intrin == I_MIN) {
+        int dtypecompare = DTYG(argdtype);
+        if (dtypecompare == TY_WORD) dtypecompare = TY_INT;
+        if (dtypecompare == TY_DWORD) dtypecompare = TY_INT8;
+        if (dtypecompare > maxtype)
+          maxtype = dtypecompare;
+	  maxdtyppe = argdtype;
```
bryanpkc, line 4337: Typo: `maxdtyppe` &rarr; `maxdtype`

tools/flang1/flang1exe/semfunc.c
```diff
@@ -4264,11 +4318,25 @@ ref_intrin(SST *stktop, ITEM *list)
...
+
     if (!dtype1) {
       f_dt = dtype1 = argdtype; /* Save 1st arg's data type */
       if (DTY(argdtype) == TY_ARRAY)
         break;
     } else {
+      if (intrin == I_MAX || intrin == I_MIN) {
+        int dtypecompare = DTYG(argdtype);
+        if (dtypecompare == TY_WORD) dtypecompare = TY_INT;
+        if (dtypecompare == TY_DWORD) dtypecompare = TY_INT8;
```
bryanpkc, line 4334: Please `clang-format` the diff and indent the code correctly.

bryanpkc (changes requested):
The code doesn't compile. Did you push an incomplete patch?

tools/flang1/flang1exe/semfunc.c
bryanpkc: Also, this line is not indented correctly. It's causing a `-Wmisleading-indentation` error in the CI.

Also, I don't understand why you need two variables. It seems that you only need one.

tools/flang1/flang1exe/semfunc.c
bryanpkc: The indentation on lines 4332 through 4336 is still confusing and arguably incorrect.

tools/flang1/flang1exe/semfunc.c
```diff
@@ -4178,6 +4228,9 @@ ref_intrin(SST *stktop, ITEM *list)
   int tmp, tmp_ast;
   FtnRtlEnum rtlRtn;
   int intrin; /* one of the I_* constants */
+  int dtype_new = 0;
```
bryanpkc, line 4231: `dtype_new` is not used after it is set by `check_max_min_argument`. Can we delete it?

tools/flang1/flang1exe/semfunc.c
```diff
@@ -4264,11 +4317,26 @@ ref_intrin(SST *stktop, ITEM *list)
...
       if (DTY(argdtype) == TY_ARRAY)
         break;
     } else {
+      if (intrin == I_MAX || intrin == I_MIN) {
+        int dtypecompare = DTYG(argdtype);
+        if (dtypecompare == TY_WORD) dtypecompare = TY_INT;
+        if (dtypecompare == TY_DWORD) dtypecompare = TY_INT8;
+        if (dtypecompare > maxtype) {
+          maxtype = dtypecompare;
+          dtype1 = argdtype;
```
bryanpkc, line 4336: Can you clarify why the argument type checks and promotion cannot be done for `I_MAX` and `I_MIN` in the `intrinsic_error` block? It semes that that block of code was designed to handle these two intrinsics.

Liuyunlong0336: The logic in this function is relatively complex.

For `intrinsic_error` ,these situations will occur:

The first scenario is as follows：
```
program test
  implicit none
  integer :: a, b, c
  a = 1
  b = 2
  c = max(a)
end program test
```
The second scenario is as follows：
```
program test
  implicit none
  integer :: a, b, c
  a = 1
  b = 2
  c = max()
end program test
```
I understand that when there is an obvious error in the arguments, there are no arguments or the arguments number is wrong, in this case `intrinsic_error` will be called.

And For logic in the `intrinsic_error` block, it is prepared for generic situations：
As follows:
```
/* Need to add a check for min and max first */
if (STYPEG(sptre) == ST_GENERIC && (intrin == I_MAX || intrin == I_MIN)) {
```
So I don’t dare to modify this logic.

And For our case, for example, argument type program will not trigger `intrinsic_error`.
For example:
```
program test
  implicit none
  integer :: a, c
  real :: b
  a = 1
  b = 2
  c = max(a, b)
end program test
```

In this case, the compilation can still pass, but the native version of the compiler will occur a warning.
```
F90-W-0093-Type conversion of expression performed
```
This is what a general check of this logic reports，`chktyp(sp, argtyp, TRUE);` in this logic will be checked, and this check is for many kinds of intrinsics.
```
             if (((intrin == I_BGE) || (intrin == I_BGT) ||
                  (intrin == I_BLE) || (intrin == I_BLT)) &&
                 (SST_IDG(sp) == S_CONST || SST_IDG(sp) == S_IDENT)) {
               mkexpr1(sp);
             } else
               chktyp(sp, argtyp, TRUE);
```

So I haven’t found a perfect method yet. For type judgment and type selection, I chose to add judgment and type update logic in front like this PR.

tools/flang1/flang1exe/semfunc.c
Liuyunlong0336: `dtype_new` can be deleted, but `dtype_last` needs to record the last value（The type of the previous argument） and compare it with the new value, So I haven't thought of a good way to delete `dtype_last` yet.

pawosm-arm (changes requested):
a small typo in the commit message: `precision atgument` vs. `precision argument`

tools/flang1/flang1exe/semfunc.c
bryanpkc: OK

pawosm-arm:
Anything happens here?

PR #1420, opened 2023-09-20
[flang2] byval and sret attributes should have associated types

bryanpkc:
When the return value of the function is a structure, the ABI might require that the caller pass the structure into the function call as an implicit input argument; in such cases the generated IR for the argument will have the sret attribute. Similarly, pass-by-value arguments are given the byval attribute in the IR. Since the adoption of opaque pointers, such attributes are required to be accompanied with the pointee type. This patch fixes flang2 to generate correct IR for such argument/return types.


bryanpkc:
@pawosm-arm @shivaramaarao This is ready for review.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -7279,6 +7279,8 @@ gen_arg_operand(LL_ABI_Info *abi, unsigned abi_arg, int arg_ili)
     operand = gen_llvm_expr(value_ili, arg_type);
   if (arg->kind == LL_ARG_BYVAL && !missing)
     operand->flags |= OPF_SRARG_TYPE;
+  if (arg->kind == LL_ARG_INDIRECT && !missing && (abi_arg == 0))
```
pawosm-arm, line 7282: IMO it would be easier to read if it could have more parentheses...

bryanpkc: Heh, I feel the opposite. `abi_arg == 0` in a compound `&&` expression doesn't really need the parentheses. I can change it if you insist.

pawosm-arm (approved):

bryanpkc:
@shivaramaarao gentle ping

shivaramaarao (approved):

PR #1422, opened 2023-09-25
Fix semantic analysis failures of LBOUND/UBOUND in type specification

wanbinchen-hnc:
An issue occurs when LBOUND/UBOUND appears in the type specification
of a function result and the DIM argument is missing.

The root cause of this issue is that LBOUND/UBOUND is converted to
runtime call in semantic phase, which is too early for type-spec of
a function result. On the calling side of the function, we have no
way of knowing whether LBOUND/UBOUND exists or its reference to the
function's dummy arguments.

We fix the issue by delaying the conversion to the tranform phase.
In particular, we perform the conversion in dummy-actual argument
replacement if the array argument of LBOUND/UBOUND is an assumed-shape
array dummy.

In addition, we fix an ICE that occurs when LBOUND/UBOUND appears in
subscript expressions in type-spec, fix the link error that occurs
when both DIM and KIND arguments are present, fix an issue with LBOUND
on the last dimension of an assumed-size array, and add support for
LBOUND/UBOUND on an assumed-rank array.

### DIFF
--- a/test/f90_correct/src/submod32.f90
+++ b/test/f90_correct/src/submod32.f90
@@ -32,10 +32,11 @@ program prog
     print *, "kind: ", kind(x)
     print *, "maxval: ", maxval(x)
     call check_arr(x)
-    if ( a .EQ. lbound(x, DIM=1) .AND. b .EQ. size(x) .AND. maxval(x) .EQ. m) then 
+    if ( a .EQ. 1 .AND. lbound(x, DIM=1) .EQ. 7 .AND. b .EQ. size(x) .AND. maxval(x) .EQ. m) then
         print *, " PASS "
     else 
-        print *, "FAILED: lbound of arr in submod is ", a, " and lbound of x is ", lbound(x, DIM=1)
+        print *, "FAILED: lbound of arr in submod is ", a, " and the expection is 1"
+        print *, "FAILED: lbound of x is is ", lbound(x, DIM=1), "and the expection is 7"
         print *, "FAILED: size of arr in submod is ", b, " and size of x is ", size(x)
         print *, "FAILED: maxval of arr in submod is", m, " and maxval of x is", maxval(x)
     end if
--- a/tools/flang1/flang1exe/semfunc.c
+++ b/tools/flang1/flang1exe/semfunc.c
@@ -53,6 +53,7 @@ static void replace_arguments(int, int);
 static void rewrite_triples(int, int, int);
 static void rewrite_subscr(int, int, int);
 static void replace_formal_triples(int, int, int);
+static void replace_formal_bounds(int, int, int, int);
 static int getMergeSym(int, int);
 static void ref_pd_subr(SST *, ITEM *);
 static void ref_intrin_subr(SST *, ITEM *);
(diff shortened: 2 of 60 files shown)


bryanpkc:
There are typos and format errors in your commit message; please reword it. It should start with a commit title, with a topic prefix, and separated from the message body with a blank line. For example:
```
[flang1] Fix handling of LBOUND/UBOUND in type specification

An issue occurs when LBOUND/UBOUND appears in the type specification
of a function result and the DIM argument is missing.

The root cause of this issue is that LBOUND/UBOUND is converted to
runtime call in semantic phase, which is too early for type-spec of
a function result. On the calling side of the function, we have no
way of knowing whether LBOUND/UBOUND exists or its reference to the
function's dummy arguments.

We fix the issue by delaying the conversion to the tranform phase.
In particular, we perform the conversion in dummy-actual argument
replacement if the array argument of LBOUND/UBOUND is an assumed-shape
array dummy.

In addition, we fix an ICE that occurs when LBOUND/UBOUND appears in
subscript expressions in type-spec, fix the link error that occurs
when both DIM and KIND arguments are present, fix an issue with LBOUND
on the last dimension of an assumed-size array, and add support for
LBOUND/UBOUND on an assumed-rank array.
```

wanbinchen-hnc:
@bryanpkc  Thanks for your comments, I have updated the commit message.

test/f90_correct/src/submod32.f90
```diff
@@ -32,7 +32,7 @@ program prog
     print *, "kind: ", kind(x)
     print *, "maxval: ", maxval(x)
     call check_arr(x)
-    if ( a .EQ. lbound(x, DIM=1) .AND. b .EQ. size(x) .AND. maxval(x) .EQ. m) then 
+    if ( a .EQ. 1 .AND. b .EQ. size(x) .AND. maxval(x) .EQ. m) then
```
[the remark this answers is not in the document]
wanbinchen-hnc: Thanks for your comments, I have updated the test case.

pawosm-arm (approved):
I can give you a tentative approval. This is an extensive change where it would be difficult to prove validity of the consequences of all the possible interactions. Therefore I've hauled this commit through various flows of our CI, and it didn't break any of the workloads, also NAG tests seemed to be happy with it.

tools/flang1/flang1exe/semfunc.c
```diff
@@ -2727,6 +2731,67 @@ replace_formal_triples(int ast, int dscptr, int nactuals)
   }
 }
 
+/** \brief Replace the intrinsic call of LBOUND/UBOUND on assumed-shape formal.
+ *
```
bryanpkc, line 2735: Please add a line here to explain why we only need to handle `A_BINOP`, `A_CONV`, etc. in the `switch` statement in this function. Our internal version has this:

```c
/** \brief Replace the intrinsic call of LBOUND/UBOUND on assumed-shape formal.
 * Model after replace_formal_triples which replaces triples in bounds.
 *
```

bryanpkc (approved):
This has passed our internal testing.

tools/flang1/flang1exe/semfunc.c
wanbinchen-hnc: The Minor nit has been fixed.

shivaramaarao (approved):

PR #1431, opened 2023-11-14
[flang1] Fix a parallel do bug

zhushengtao1:
When the stride of parallel do is negative, flang will always exit loop.
It is because check_loop_bound function always check the incr of a loop
to avoid out of loop bound. It is not necessary since __kmp_for_static_init
in openmp have do the same thing, and will change the stride to match
the actual number of iterations. The function is only needed when SCHEDULE
is enabled.

### DIFF
--- a/test/mp_correct/src/do20.f90
+++ b/test/mp_correct/src/do20.f90
@@ -1,0 +1,21 @@
+!* Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+!* See https://llvm.org/LICENSE.txt for license information.
+!* SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+!       simple parallel do, negative stride
+
+program main
+  call mmm(10, 1, -2)
+contains
+
+subroutine mmm (m1, m2, m3)
+  integer :: m, m1, m2, m3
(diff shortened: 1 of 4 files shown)


test/mp_correct/src/do20.f90
```diff
@@ -0,0 +1,27 @@
+!* Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+!* See https://llvm.org/LICENSE.txt for license information.
+!* SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+!	parallel do with negative stride
```
bryanpkc, line 5: I suggest mentioning the schedule type, like in the other `pd*.f` and `do*.f` tests:
```fortran
!   simple parallel do, negative stride
```

test/mp_correct/src/do20.f90
```diff
@@ -0,0 +1,27 @@
...
+
+  call mmm(10, 1, -2)
+
+contains
+
+subroutine mmm (m1, m2, m3)
+
+    integer :: m, m1, m2, m3
+
+
```
bryanpkc, line 17: Can you delete unnecessary blank lines and make your indentation consistent (2 spaces per level)?

PR #1433, opened 2023-12-13
[workflows] Update GitHub workflows

bryanpkc:
Use the Ubuntu 22.04 runner and the `release_16x` and `release_17x` branches of classic-flang-llvm-project to build and test Classic Flang. We use the pre-built classic-flang-llvm-project for building, so we do not actually need to install a specific version of GCC. Also simplify the logic for finding and downloading a suitable prebuilt compiler.

### DIFF
--- a/.github/workflows/build_flang.yml
+++ b/.github/workflows/build_flang.yml
@@ -14,15 +14,24 @@ on:
 
 jobs:
   build_flang:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     env:
       install_prefix: /usr/local
     strategy:
       matrix:
         target: [X86]
-        cc: [clang, gcc]
-        version: [10, 11]
-        llvm_branch: [release_15x]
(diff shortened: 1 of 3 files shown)


.github/workflows/build_flang.yml
```diff
@@ -14,15 +14,15 @@ on:
 
 jobs:
   build_flang:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
pawosm-arm, line 17: Hopefully it won't break anything. As we shifted to ubuntu-22.04, we're observing sudden segfault in `malloc()` in nag-tests...

pawosm-arm (approved):
Approval with some general observation about Ubuntu.

pawosm-arm (approved):

kaadam (approved):
Bryan, thanks for updating the Windows part of the actions. LGTM.

PR #1434, opened 2023-12-14
flang2: fix the expected number of parameters for __pd_asin_1

pawosm-arm:
This bug was hidden unnoticed for years, until the most recent changes in LLVM (LLVM-18 candidate) have introduced additional assertions in the loop vectorizer, and one of them is failing on wrongly declared functions like this one.


tools/flang2/flang2exe/iliutil.cpp
```diff
@@ -6134,7 +6134,7 @@ addarth(ILI *ilip)
 
   case IL_DASIN:
     if (XBIT_NEW_MATH_NAMES) {
-      fname = make_math(MTH_asin, &funcsptr, 1, false, DT_DBLE, 2, DT_DBLE,
+      fname = make_math(MTH_asin, &funcsptr, 1, false, DT_DBLE, 1, DT_DBLE,
```
bryanpkc, line 6137: You should only specify one argument data type instead of two:
```c
      fname = make_math(MTH_asin, &funcsptr, 1, false, DT_DBLE, 1, DT_DBLE);
```

pawosm-arm: Thanks for noticing this.

shivaramaarao (approved):

Issue #1442, opened 2024-04-15
[regression] An ICE while building libmbd after PR #1422


pawosm-arm:
After introduction of #1422 the libmbd library (see https://github.com/libmbd/libmbd.git ) cannot be built. An Internal Compiler Error (ICE) can be observed when trying to compile the mbd_methods.F90 file (https://github.com/libmbd/libmbd/blob/master/src/mbd_methods.F90 ).
```
[ 60%] Building Fortran object src/CMakeFiles/mbd.dir/mbd_methods.F90.o
cd src && flang -Dmbd_EXPORTS -Isrc -g -fPIC -c src/mbd_methods.F90 -o CMakeFiles/mbd.dir/mbd_methods.F90.o
Lowering Error: symbol freq$sd is an inconsistent array descriptor
Lowering Error: symbol freq$sd is an inconsistent array descriptor
F90-F-0000-Internal compiler error. Errors in Lowering       2  (src/mbd_methods.F90: 154)
F90/aarch64 Linux F90  - 1.5 2017-05-01: compilation aborted
```

bryanpkc:
@wanbinchen-hnc Could you help take a look at this?

wanbinchen-hnc:
@pawosm-arm @bryanpkc A fixing PR has been submitted.

pawosm-arm:
The fix has worked for me.

PR #1443, opened 2024-04-18
Fix regression introduced by PR #1422

wanbinchen-hnc:
This regression referred in issue #1442 occurs when we pass a struct component for an assumed-shape formal argument, e.g. `geom%freq` in
```fortran
alpha = alpha_dyn_qho(alpha_0, omega, geom%freq, dalpha, grad_request_t())
```
where the third formal argument is declared as
```fortran
type(quad_pt_t), intent(in) :: freq(0:)
```
The root cause is that we have lost the parent `geom` in the expression for the descriptor of `geom%freq` when using the function `get_extent`.

We fix this by manually constructing the subscript expression for the extent of `geom%freq`.

### DIFF
--- a/test/f90_correct/inc/bound17.mk
+++ b/test/f90_correct/inc/bound17.mk
@@ -1,0 +1,19 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+$(TEST): run
+
+build:  $(SRC)/$(TEST).f90
+	-$(RM) $(TEST).$(EXESUFFIX) core *.d *.mod FOR*.DAT FTN* ftn* fort.*
+	@echo ------------------------------------ building test $@
+	-$(FC) -c $(FFLAGS) $(LDFLAGS) $(SRC)/$(TEST).f90 -o $(TEST).$(OBJX)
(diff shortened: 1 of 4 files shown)


pawosm-arm (approved):

bryanpkc (approved):
LGTM. This solves the problem in #1442 but the libmbd build still fails, this time with a different error; I'll create a new issue. I'll give @shivaramaarao a chance to review before merging this.

shivaramaarao (approved):
looks good to me.

pawosm-arm:
Hi @bryanpkc this commit fixes the building issue for me, yet bear in mind that I'm building libmbd from the `0.10.3` tag, not the top of the trunk. Also, I'm doing plain regular build, an attempted MPI build has failed like this:
```
[ 10%] Building Fortran object src/CMakeFiles/mbd.dir/mbd_mpi.F90.o
cd src && mpifort -DWITH_MPI -DWITH_SCALAPACK -Dmbd_EXPORTS -Isrc  -DWITH_MPI -g -fPIC   -c src/mbd_mpi.F90 -o CMakeFiles/mbd.dir/mbd_mpi.F90.o
F90-S-0000-Internal compiler error. incomplete interface file, missing AST    1038  (src/mbd_mpi.F90: 6)
F90-F-0004-incomplete IPA file, missing AST   (src/mbd_mpi.F90: 6)
F90/aarch64 Linux FlangArm F90  - 1.5 2017-05-01: compilation aborted
```
I don't know how often MPI build of this library is needed, for me the plain build is sufficient.

bryanpkc:
>  Also, I'm doing plain regular build, an attempted MPI build has failed like this

Thanks for the heads-up. Let me try to look at this as well.

PR #1448, opened 2024-06-22
[NFC] modifying the testcase by adding Mbackslash and Mnobackslash options together

shivaramaarao:
flang should use the last specified option when both these options are used. e.g Mbaskslash should be used when "-Mnobackslash -Mbackslash <filename>" arguments are specified.

### DIFF
--- a/test/lex/backslash.f90
+++ b/test/lex/backslash.f90
@@ -8,6 +8,12 @@
 ! RUN: %flang -c -fno-backslash %s
 ! RUN: not %flang -c -Mnobackslash %s 2>&1 | FileCheck %s
 ! RUN: not %flang -c -fbackslash %s 2>&1 | FileCheck %s
+! RUN: %flang -c -Mnobackslash -Mbackslash %s
+! RUN: %flang -c -fbackslash -fno-backslash %s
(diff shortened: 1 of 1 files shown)


test/lex/backslash.f90
```diff
@@ -8,6 +8,10 @@
 ! RUN: %flang -c -fno-backslash %s
 ! RUN: not %flang -c -Mnobackslash %s 2>&1 | FileCheck %s
 ! RUN: not %flang -c -fbackslash %s 2>&1 | FileCheck %s
+! RUN: %flang -c -Mnobackslash -Mbackslash %s
+! RUN: %flang -c -fbackslash -fno-backslash %s
+! RUN: not %flang -c -Mnobackslash -Mbackslash -Mnobackslash %s 2>&1 | FileCheck %s
```
bryanpkc, line 13: Is the first instance of `-Mnobackslash` intentional? It doesn't match the pattern on other lines.

shivaramaarao: yes. I intentionally added the combination of Mnobackslash -Mbackslash to check if compiler resolves to Mbackslash correctly.

bryanpkc: In that case, do you want to apply the same treatment to line 14?

pawosm-arm (approved):

PR #1449, opened 2024-07-23
[flang2] Generate loads of complex variables with correct alignment

bryanpkc:
flang2 calls the function `make_load()` to create various types of load instructions in the LLVM IR output. The function accepts a `flags` argument of type `LL_InstrListFlags` which should encode the alignment of the load, among other things. Apparently, for the `IL_LDSCMPLX`, `IL_LDDCMPLX`, and `IL_LDQCMPLX` opcodes, the flags had never been computed correctly. This patch makes such loads consistent with loads of other types, and also adds a test case.

### DIFF
--- a/test/llvm_ir_correct/load-alignment.f90
+++ b/test/llvm_ir_correct/load-alignment.f90
@@ -1,0 +1,50 @@
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+! Check that load instructions are emitted with correct alignment.
+
+! RUN: %flang -S -emit-flang-llvm %s -o %t.ll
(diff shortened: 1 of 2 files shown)


pawosm-arm (commented):
NB, checking for `TARGET_SUPPORTS_QUADFP` doesn't seem to be needed in the new execution path, is it intentional?

bryanpkc:
> NB, checking for `TARGET_SUPPORTS_QUADFP` doesn't seem to be needed in the new execution path, is it intentional?

@pawosm-arm Yes, it is intentional. `ldst_instr_flags_from_dtype_nme` and `ldst_instr_flags_from_dtype` behave the same way regardless of the data type, and the underlying `alignment` function (in `dtypeutl.cpp`) handles `TY_QCMPLX` only when `TARGET_SUPPORTS_QUADFP` is defined.

pawosm-arm (approved):

shivaramaarao (approved):

PR #1455, opened 2024-10-02
[DebugInfo][test] Fix tests for LLVM 19 compatibility

bryanpkc:
Upstream LLVM has eliminated debug intrinsics like `@llvm.dbg.value()` and `@llvm.dbg.declare()`, and replaced them with `DPValue` objects that are printed as `#dbg_value()` and `#dbg_declare()`. This patch modifies the FileCheck patterns in the tests to match, and make them require the `llvm-19` feature.


alokkrsharma (approved):
LGTM

shivaramaarao (approved):

Issue #1453, opened 2024-09-20
OpenMP test failure in GitHub workflow


bryanpkc:
[This CI job](https://github.com/flang-compiler/flang/actions/runs/10942659878/job/30380212588?pr=1452) had failed, even though there had been no code change in either LLVM or Classic Flang. The commands and errors were as follows:
```
clang -c -c -I/home/runner/work/flang/flang/test/mp_correct/lit/../src  -fopenmp  /home/runner/work/flang/flang/test/mp_correct/lit/../src/check.c -o check.o
------------ building test pv01.o
flang -c -I/home/runner/work/flang/flang/test/mp_correct/lit/../src  -fopenmp    /home/runner/work/flang/flang/test/mp_correct/lit/../src/pv01.f90
flang   -fopenmp pv01.o check.o  -o a.exe
------------ executing test pv01
OMP_NUM_THREADS=4 ./a.exe 
Warning: ieee_inexact is signaling
FORTRAN STOP
 Error occured while allocating when parallel
```

bryanpkc:
This problem has happened [again](https://github.com/flang-compiler/flang/actions/runs/11028626608/job/30629239465), also on X86 with `release_18x` branch of classic-flang-llvm-project.

bryanpkc:
[Another occurence.](https://github.com/flang-compiler/flang/actions/runs/11212010211/attempts/1)

PR #1457, opened 2024-10-07
[workflows] Disable intermittently failing build/test job

bryanpkc:
The test case mp_correct/lit/pv01.sh has been failing intermittently in the GitHub workflows (see issue #1453). It seems that it only happens when classic-flang-llvm-project is built with Clang 15. This patch disables the job temporarily, until we can determine the root cause of the problem.

### DIFF
--- a/.github/workflows/build_flang.yml
+++ b/.github/workflows/build_flang.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         target: [X86]
         cc: [clang]
-        version: [14, 15]
+        version: [14]
         llvm_branch: [release_17x, release_18x]
         include:
           - target: X86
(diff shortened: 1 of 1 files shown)


bryanpkc:
I'm going to run the workflows on this PR multiple times, to confirm that the error no longer occurs (i.e. it only occurs when classic-flang-llvm-project is built with Clang 15).

bryanpkc:
The failure has not occurred in 10 CI runs. It seems that the problem is indeed triggered by building with Clang 15.

pawosm-arm (approved):
I'm bit confused by this versioning.

bryanpkc:
> I'm bit confused by this versioning.

We have always built Clang and LLVM with multiple compilers (two different versions of Clang and a recent GCC) before using that Clang to build Classic Flang. This practice was adopted before I joined the project. Apparently, building LLVM with different compilers can be useful for catching problems like this one. 😄

shivaramaarao (approved):

PR #1460, opened 2024-11-06
[Flang2] Fix duplicate TBAA type systems across modules

### LINKED ISSUE #1458: Incorrect no alias by TBAA due to a problem with metadata tags
In Flang, variables are assigned distinct Type-Based Alias Analysis (TBAA) tags, even when they share identical types. This method, however, is insufficient for determining whether two variables from different functions are aliased. To address this, a separate metadata structure—referred to as the "type system"—is created for each function. Within this system, variables defined in a function are derived from its specific type system. For two variables to be confirmed as non-aliasing, they must originate from the same type system but possess distinct TBAA tags.

The type system metadata utilizes a simple counter assigned to functions based on their order of appearance. However, because this metadata is generated at the frontend, functions in separate modules can be assigned identical type systems. For example, the first and second functions in File A have type systems "Flang FAA 1" and "Flang FAA 2," respectively, while the first function in File B also be assigned "Flang FAA 1." This overlap can cause incorrect non-aliasing determinations between variables in the first functions of Files A and B.

In the following example, two accesses to the same element of an array are mistakenly identified as non-aliasing, which can lead to issues when the modify1 function is inlined.

```fortran
      program main
      implicit none
      integer, parameter :: n = 5
      real :: arr1(n)
      integer :: i
      i = 0
      arr1 = 3.2
      arr1(i) = 4
      call modify1(arr1)
      call modify2(arr1)

      ! t1.4 from type system Flang FAA 1
      arr1(i) = arr1(i) + 2.5

      call printout(arr1)
      end program main
```

```fortran
      subroutine modify1(arr1)
      implicit none
      real, intent(inout) :: arr1(:)

      ! t1.9 from type system Flang FAA 1
      arr1(0) = arr1(0) + 0.5
      end subroutine modify1

      subroutine modify2(arr1)
      implicit none
      real, intent(inout) :: arr1(:)
      arr1(2) = arr1(2) + 1.5
      end subroutine modify2

      subroutine printout(arr1)
      implicit none
      real, intent(in) :: arr1(:)
      integer :: i
      do i = 1, size(arr1)
        print arr1(i), " "
      enddo
      end subroutine printout
```

A simple solution is to append the hash of the module name to the type system metadata.

1997alireza:
Currently, functions within the same module are assigned unique TBAA type system IDs based on their order, but functions in different modules can unintentionally receive the same ID, leading to incorrect non-aliasing results. This patch resolves the issue by appending a hash of the module name to each function type system metadata, ensuring unique type systems across modules and preventing aliasing errors. This resolves #1458.

### DIFF
--- a/tools/flang2/flang2exe/cgmain.cpp
+++ b/tools/flang2/flang2exe/cgmain.cpp
@@ -10,29 +10,30 @@
    \brief Main source module to translate into LLVM
  */
 
-#include "cgmain.h"
-#include "dtypeutl.h"
-#include "ll_ftn.h"
-#include "exp_rte.h"
-#include "error.h"
-#include "machreg.h"
-#include "dinit.h"
-#include "cg.h"
-#include "mach.h"
-#include "fih.h"
-#include "pd.h"
-#include "llutil.h"
-#include "lldebug.h"
-#include "go.h"
-#include "sharedefs.h"
-#include <stdlib.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
-#include "llassem.h"
-#include "ll_write.h"
+#include <string>
+#include "cgmain.h"
+#include "cg.h"
+#include "dinit.h"
+#include "dtypeutl.h"
+#include "error.h"
+#include "exp_rte.h"
 #include "expand.h"
-#include "outliner.h"
+#include "fih.h"
+#include "go.h"
+#include "ll_ftn.h"
+#include "ll_write.h"
+#include "llassem.h"
+#include "lldebug.h"
+#include "llutil.h"
+#include "mach.h"
+#include "machreg.h"
 #include "mth.h"
+#include "outliner.h"
+#include "pd.h"
+#include "sharedefs.h"
 #if defined(SOCPTRG)
 #include "soc.h"
 #endif
@@ -2639,11 +2640,12 @@ get_omnipotent_pointer(LL_Module *module)
     LL_MDRef s0;
     LL_MDRef r0;
     LL_MDRef a[3];
-    char baseBuff[32];
+    char baseBuff[64];
     const char *baseName = "Flang FAA";
     const char *const omniName = "unlimited ptr";
     const char *const unObjName = "unref ptr";
-    snprintf(baseBuff, 32, "%s %x", baseName, funcId);
+    snprintf(baseBuff, 64, "%s %zx %x", baseName,
+             std::hash<std::string>{}(current_module->module_name), funcId);
     s0 = ll_get_md_string(module, baseBuff);
     r0 = ll_get_md_node(module, LL_PlainMDNode, &s0, 1);
     a[0] = ll_get_md_string(module, unObjName);
--- a/test/llvm_ir_correct/Inputs/tbaa_multimod_01_input.f90
+++ b/test/llvm_ir_correct/Inputs/tbaa_multimod_01_input.f90
@@ -1,0 +1,22 @@
+! A part of the test provided in tbaa_multimod_01.f90
+
+subroutine modify1(arr)
+  implicit none
+  real, intent(inout) :: arr(:)
+  arr(0) = arr(0) + 0.5
+end subroutine modify1
+
+subroutine modify2(arr)
+  implicit none
+  real, intent(inout) :: arr(:)
+  arr(2) = arr(2) + 1.5
+end subroutine modify2
+
+subroutine printout(arr)
+  implicit none
+  real, intent(in) :: arr(:)
+  integer :: i
+  do i = 1, size(arr)
+    print arr(i), " "
+  enddo
+end subroutine printout
--- a/test/llvm_ir_correct/Inputs/tbaa_multimod_02_input.f90
+++ b/test/llvm_ir_correct/Inputs/tbaa_multimod_02_input.f90
@@ -1,0 +1,10 @@
+! A part of the test provided in tbaa_multimod_02.f90
+
+subroutine to_load(arr)
+  implicit none
+  real, intent(inout) :: arr(:)
+  real :: var
+  var = arr(0)
+  var = var * 2
+  print var
+end subroutine to_load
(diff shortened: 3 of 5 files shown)


tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -43,6 +43,8 @@
 #include "symfun.h"
 #include "ilidir.h"
 #include "fdirect.h"
+#include <string>
+#include <hash_map>
```
bryanpkc, line 48: Use of this deprecated header is causing a build failure. Please do not depend on it.

1997alireza: Replaced it with unordered_map.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -43,6 +43,8 @@
 #include "symfun.h"
 #include "ilidir.h"
 #include "fdirect.h"
+#include <string>
+#include <unordered_map>
```
shivaramaarao, line 47: we dont need to include unordered_map for this change.

test/llvm_ir_correct/tbaa_multimod_01_callee.f90
```diff
@@ -0,0 +1,24 @@
...
+end subroutine modify2
+
+subroutine printout(arr)
+  implicit none
+  real, intent(in) :: arr(:)
+  integer :: i
+  do i = 1, size(arr)
+    print arr(i), " "
+  enddo
+end subroutine printout
```
bryanpkc, line 24: Make sure that your files end with a newline.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -2643,7 +2644,7 @@ get_omnipotent_pointer(LL_Module *module)
     const char *baseName = "Flang FAA";
     const char *const omniName = "unlimited ptr";
     const char *const unObjName = "unref ptr";
-    snprintf(baseBuff, 32, "%s %x", baseName, funcId);
+    snprintf(baseBuff, 32, "%s %zx %x", baseName, std::hash<std::string>{}(current_module->module_name), funcId);
```
bryanpkc, line 2647: Can you format this line according to LLVM Coding Standards?

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -43,6 +43,7 @@
 #include "symfun.h"
 #include "ilidir.h"
 #include "fdirect.h"
+#include <string>
```
bryanpkc, line 46: Based on [this build error](https://github.com/flang-compiler/flang/actions/runs/11842152761/job/33009497541?pr=1460), this `#include` statement should be placed above line 18 (`#include "machreg.h"`). In fact, all standard headers should be included before custom headers, including `stdlib.h`, `stdio.h`, etc.

test/llvm_ir_correct/tbaa_multimod_01_caller.f90
```diff
@@ -0,0 +1,18 @@
+! This test contians two files, tbaa_multimod_01_caller.f90 and tbaa_multimod_01_callee.f90
+! RUN: %flang -Wl,-mllvm,-aa-trace -fuse-ld=lld -flto=full -O3 %s %S/tbaa_multimod_01_callee.f90 -o - 2>&1 1>- | FileCheck %s
```
bryanpkc, line 2: These tests are failing in the CI jobs. I understand that `-aa-trace` only works whne LLVM has assertions enabled, but I believe that the GitHub workflow for classic-flang-llvm-project already configures LLVM with `-DLLVM_ENABLE_ASSERTIONS=ON`:

https://github.com/flang-compiler/classic-flang-llvm-project/blob/release_17x/.github/workflows/pre-compile_llvm.yml#L65

https://github.com/flang-compiler/classic-flang-llvm-project/blob/release_17x/.github/workflows/UbuntuDockerfile#L54

@1997alireza  Can you build classic-flang-llvm-project using the same commands as used by the CI jobs, and confirm whether you can reproduce the test failures in your environment?

bryanpkc: I think I know what the problem is. Your tests specify `-fuse-ld=lld` but `lld` is not available in the CI. And for some reason, the errors from stderr are eaten by `FileCheck`, and not shown in the `llvm-lit` output, causing confusion for us.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -15,6 +15,10 @@
 #include "ll_ftn.h"
 #include "exp_rte.h"
 #include "error.h"
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <string>
```
bryanpkc, line 21: All standard headers should be included before custom headers (I apologize if I wasn't clear). And please sort these alphabetically.

test/llvm_ir_correct/tbaa_multimod_01_input.f90
```diff
@@ -0,0 +1,25 @@
+! A part of the test provided in tbaa_multimod_01.f90
+! Empty RUN and CHECK to prevent error
+! RUN: echo "NoCheck" | FileCheck %s
+! CHECK: NoCheck
```
bryanpkc, line 4: Is it possible to move these files into `test/llvm_ir_correct/Inputs/`, and remove the unnecessary `RUN` and `CHECK` directives?

bryanpkc (approved):
LGTM.

tools/flang2/flang2exe/cgmain.cpp
```diff
@@ -2643,7 +2644,8 @@ get_omnipotent_pointer(LL_Module *module)
     const char *baseName = "Flang FAA";
     const char *const omniName = "unlimited ptr";
     const char *const unObjName = "unref ptr";
-    snprintf(baseBuff, 32, "%s %x", baseName, funcId);
+    snprintf(baseBuff, 32, "%s %zx %x", baseName,
```
pawosm-arm, line 2647: Are you sure it still fits into 32 characters?

pawosm-arm (approved):
LGTM, just asked one control question.

tools/flang2/flang2exe/cgmain.cpp
bryanpkc: @pawosm-arm Check again?

pawosm-arm (approved):

PR #1467, opened 2026-05-22
[flang1][openmp] Add missing check for statement function var

zhushengtao1:
When statement function used in an openmp block, some variables used in stfunc does not added in openmp shared list which will create undefined value in flang2 and cause error in llvm. This commit add a traverse for statement function variables and add rvalue to openmp shared list by sem_check_scope like what ref_object do.

### DIFF
--- a/test/mp_correct/inc/stfunc_derived.mk
+++ b/test/mp_correct/inc/stfunc_derived.mk
@@ -1,0 +1,19 @@
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+$(TEST): run
+
+build:  $(SRC)/$(TEST).f90
+	-$(RM) $(TEST).$(EXESUFFIX) core *.d *.mod FOR*.DAT FTN* ftn* fort.*
+	@echo ------------------------------------ building test $@
+	-$(FC) $(FFLAGS) $(SRC)/$(TEST).f90 -o $(TEST).$(OBJX)
+	-$(FC) $(LDFLAGS) $(TEST).$(OBJX) $(LIBS) -o $(TEST).$(EXESUFFIX)
+
+run:
+	@echo ------------------------------------ executing test $(TEST)
+	-$(RUN4) $(TEST).$(EXESUFFIX)
+
+verify: ;
--- a/tools/flang1/flang1exe/semfunc2.c
+++ b/tools/flang1/flang1exe/semfunc2.c
@@ -84,6 +84,11 @@ static ITEM *stfunc_argl;
 static int pass_position = 0;
 static int pass_object_dummy = 0;
 
+/** \brief current statement function argument information used in AST traversal
+ *  to exclude dummy argument in statement function.
+ */
+static ARGINFO *cur_stfunc_arginfo;
+
 /** \brief Set position of type-bound procedure passed object location. */
 void
 set_pass_objects(int pos, int pod)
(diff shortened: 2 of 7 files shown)


tools/flang1/flang1exe/semfunc2.c
```diff
@@ -270,6 +271,21 @@ dump_stfunc(int sptr)
   fprintf(gbl.dbgfil, "\n");
 }
 
+static void
+chk_stfunc_shared_var(int ast, int *unused)
+{
+  ARGINFO* arginfo = cur_arginfo;
+  if (ast && A_TYPEG(ast) == A_ID) {
+    for (arginfol arginfo != NULL; arginfo = arginfo -> next)
```
bryanpkc: This seems incorrect. Did you intend to declare and initialize `arginfo` in this statement? If you do that, the variable `arginfo` can be abbreviated. For example:
```c
    for (ARGINFO *arg = cur_arginfo; arg != NULL; arg = arg->next)
```
Please run `clang-format` on your patch.

test/mp_correct/inc/stfunc_derived.mk
```diff
@@ -0,0 +1,19 @@
...
+	-$(RM) $(TEST).$(EXESUFFIX) core *.d *.mod FOR*.DAT FTN* ftn* fort.*
+	@echo ------------------------------------ building test $@
+	-$(FC) $(FFLAGS) $(SRC)/$(TEST).f90 -o $(TEST).$(OBJX)
+	-$(FC) $(LDFLAGS) $(TEST).$(OBJX) $(LIBS) -o $(TEST).$(EXESUFFIX)
+
+run:
+	@echo ------------------------------------ executing test $(TEST)
+	-$(RUN4) $(TEST).$(EXESUFFIX)
+
+verify: ;
```
bryanpkc: Missing newline at the end of the last line of the file. Please also make sure that your patch does not add trailing spaces on any line. Ditto for other files in this PR.

tools/flang1/flang1exe/semfunc2.c
```diff
@@ -83,6 +83,7 @@ static ITEM *stfunc_argl;
 
 static int pass_position = 0;
 static int pass_object_dummy = 0;
+static ARGINFO *cur_arginfo;
```
bryanpkc: It would probably be a good idea to rename this variable with a more specific name that indicates the intended usage clearly. Or at least you could add a comment.

tools/flang1/flang1exe/semfunc2.c
```diff
@@ -424,6 +440,12 @@ ref_stfunc(SST *stktop, ITEM *args)
    */
   asn_sfuse(sfdsc->l_use);
 
+  if (sem.parallel || sem.task || sem.target || sem.teams || sem.orph) {
+    cur_arginfo = sfdsc->args;
```
bryanpkc: Is there any possibility of multiple OpenMP scopes being nested within each other, causing this use of `cur_arginfo` to break?

bryanpkc (changes requested):

tools/flang1/flang1exe/semfunc2.c
zhushengtao1: add a comment

test/mp_correct/inc/stfunc_derived.mk
zhushengtao1: fixed

tools/flang1/flang1exe/semfunc2.c
zhushengtao1: fixed

tools/flang1/flang1exe/semfunc2.c
zhushengtao1: We just need to judge wheather the line which contains statement function in an OpenMP scope and add its rvalue to openmp shared list. I think it is irrelevant to whether there are multiple OpenMP nestings.
This condition is the same as  what in ref_object

tools/flang1/flang1exe/semfunc2.c
bryanpkc: There is a typo in "information".

Also, I still think the code will be more readable if you rename the variable to `cur_stfunc_arginfo`.

tools/flang1/flang1exe/semfunc2.c
bryanpkc: The condition is the same as in `ref_object`, but `ref_object` doesn't use a static variable to remember the context. My worry was whether it's possible for `cur_arginfo` to be overwritten inside the `ast_traverse` call below. On second thought, it seems extremely rare, perhaps impossible, for a statement function to nest another statement function. So this may be okay.

tools/flang1/flang1exe/semfunc2.c
zhushengtao1: fixed, thanks

tools/flang1/flang1exe/semfunc2.c
bryanpkc: Please make sure to format the code correctly. Lines should be under 80 characters in width (that's why I suggested using a shorter name for the iterator variable), and there should be no space around the `->` sign for pointer dereferences.

bryanpkc (approved):
LGTM now.

PR #1470, opened 2026-06-01
Fix RESHAPE for assumed length character arrays

zhushengtao1:
Fix a bug in the lowering of the RESHAPE intrinsic, when the element
type of the source parameter is assumed length character.

The problem is that flang1 calls the function _reshape() to optimize
RESHAPE by representing the result of RESHAPE as a (Cray) pointer of
the source argument, but the above-mentioned argument element type is
not supported according to the description of _reshape(). This patch
adds a check for the character length in _reshape(), so that it can
bail out when it sees an assumed length character array.

*(Edited to remove a second commit.)*

### DIFF
--- a/test/f90_correct/src/reshape_asschar.f90
+++ b/test/f90_correct/src/reshape_asschar.f90
@@ -1,0 +1,63 @@
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Test for the reshape function calls when the array element of the source is
+! an assumed length character varivable.
+
+program test
+  call check_alloc('1', '2', '3', '4')
+  call check_auto('12', '34', '56', '78', 4)
+  call check_cp('123', '345', '567', '789')
+  call check_fixed('abcd', 'efgh', 'ijkl', 'mnop')
+  call check_p('12', '34', '56', '78')
+  print *, "PASS"
+contains
+  subroutine check_alloc(w, x, y, z)
+    character(*), intent(in) :: w, x, y, z
+    character(1), allocatable :: a(:), b(:)
+
+    a = [w, x, y, z]
+    b = reshape([w, x, y, z], [4])
+    if (any(a /= b)) stop 1
+  end subroutine
+
+  subroutine check_auto(w, x, y, z, n)
+    character(*), intent(in) :: w, x, y, z
+    integer, intent(in) :: n
+    character(2) :: a(n), b(n)
+
+    a = [w, x, y, z]
+    b = reshape([w, x, y, z], [4])
+    if (any(a /= b)) stop 2
+  end subroutine
+
+  subroutine check_cp(w, x, y, z)
+    character(*), intent(in) :: w, x, y, z
+    character(3), contiguous, pointer :: a(:), b(:)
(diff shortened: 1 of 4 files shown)


tools/flang1/flang1exe/func.c
```diff
@@ -7359,8 +7359,11 @@ _reshape(int func_args, DTYPE dtype, int lhs)
     /* pad and order must not be present */
     return 0;
   if (DTYG(dtype) == TY_CHAR) {
+    /* character must has constant length */
```
bryanpkc: English: "must have".

test/f90_correct/src/reshape_asschar.f90
```diff
@@ -0,0 +1,63 @@
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+! Test for the reshape function calls when the array element of the resource is
+! an assumed length character varivable
```
bryanpkc: Instead of "the resource", I think you intended to say "the source" (argument of the `reshape` function).

There is also a typo at the end: "variable".

test/f90_correct/src/reshape_asschar.f90
```diff
@@ -0,0 +1,63 @@
...
+  call check_p('12', '34', '56', '78')
+  print *, "PASS"
+contains
+  subroutine check_alloc(w, x, y, z)
+    character(*), intent(in) :: w, x, y, z
+    character(1), allocatable :: a(:), b(:)
+
+    a = [w, x, y, z]
+    b = reshape([w, x, y, z], [4])
+    if (any(a /= b)) stop 1
```
bryanpkc, line 22: Is this check sufficient? Did you consider also comparing the return values of `shape(a)` and `shape(b)`?

bryanpkc (changes requested):
Fir the first commit, can you clarify the problem description in the commit message? What is the symptom, and what do you mean by "above case is not supported according to the description"? Do you mean that the `_reshape` optimization is not legal for this case, and the `RESHAPE` call must execute the slow path instead?

Similarly, the commit message of the second commit should explain the problem better.

test/f90_correct/src/associate_subs.f90
```diff
@@ -0,0 +1,30 @@
...
+    a(2)%p => tgt(2)
+    c(1)%ptr => tgt1(1,:)
+    c(2)%ptr => tgt1(2,:)
+
+    if (.not. associated(a(1)%p, tgt(1))) stop 1
+    if (.not. associated(a(2)%p, tgt(2))) stop 2
+    if (.not. associated(c(1)%ptr, tgt1(1,:))) stop 3
+    if (.not. associated(c(2)%ptr, tgt1(2,:))) stop 4
+    print *, "PASS"
+end program
```
bryanpkc: Missing newline at the end of the file.

test/f90_correct/src/reshape_asschar.f90
zhushengtao1: This is test for shape(a) and shape(b). I think this change will not affect the shape

Image

bryanpkc: OK, thanks.

bryanpkc (approved):
LGTM.

PR #1471, opened 2026-07-08
Fix compile-time warnings; NFC

bryanpkc:
Clang 21 produces new warnings when compiling Classic Flang.
This NFC PR suppress the warnings.


psoni2628 (approved):
LGTM

